PR #19, opened 2016-05-13
Generate a www.gitignore.io .gitignore file

yschimke:
Had lots of files flagged by git after a build, so generating something sensible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,48 @@
+# Created by https://www.gitignore.io/api/cmake,c++
+
+### CMake ###
+CMakeCache.txt
+CMakeFiles
+CMakeScripts
+Makefile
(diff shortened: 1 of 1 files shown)


stupaq-fb:
They are very much catch-all. I believe that using an off-source build would solve your problems entirely.

yschimke:
That makes sense.  I believe I was following the README.md, is that something I should add there?

stupaq-fb:
I'll update it. tl;dr CMake is flexible about location of the build dir (I've added /build to .gitignore initially, as that's what I've seen in other ptojects).

stupaq-fb:
Sorry, did I just reject it? I wanted to merge it in, sorry.

PR #31, opened 2016-05-17
fixing lifetime of the ConnectionAutomaton

lehecka:
It is possible that ReactiveSocket class instance is destroyed by a user client call back. Eq. onComplete will tare down everything. In that case ReactiveSocket destructor was also taking down the ConnectionAutomaton instance which was executing the callback and had some more business to do. In order to stop this premature destruction of ConnectionAutomaton, it has to have independent lifetime from ReactiveSocket class. Also we need to make sure that all streams will be properly cleaned up when the ReactiveSocket instance is going away.

### DIFF
--- a/reactivesocket-cpp/src/ConnectionAutomaton.cpp
+++ b/reactivesocket-cpp/src/ConnectionAutomaton.cpp
@@ -1,7 +1,7 @@
 // Copyright 2004-present Facebook. All Rights Reserved.
 
 
-#include "lithium/reactivesocket-cpp/src/ConnectionAutomaton.h"
+#include "reactivesocket-cpp/src/ConnectionAutomaton.h"
 
 #include <limits>
 
--- a/reactivesocket-cpp/src/ConnectionAutomaton.cpp
+++ b/reactivesocket-cpp/src/ConnectionAutomaton.cpp
@@ -1,7 +1,7 @@
 // Copyright 2004-present Facebook. All Rights Reserved.
 
 
-#include "reactivesocket-cpp/src/ConnectionAutomaton.h"
+#include "lithium/reactivesocket-cpp/src/ConnectionAutomaton.h"
 
 #include <limits>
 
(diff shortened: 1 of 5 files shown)


reactivesocket-cpp/src/ConnectionAutomaton.cpp
```diff
@@ -103,6 +113,7 @@ void ConnectionAutomaton::onSubscribe(Subscription& subscription) {
   connectionInputSub_.reset(&subscription);
   // This may result in signals being issued by the connection in-line, see
   // ::connect.
+  // TODO: implement lease here. For now we will disable flow control
```
stupaq-fb: Lease cannot be implemented here, we must always accept all frames as soon as we can, since some of them do not translate into lease-controlled events (e.g. closing a stream). Please remove.

reactivesocket-cpp/src/ConnectionAutomaton.h
```diff
@@ -36,7 +37,10 @@ class ConnectionAutomaton :
     /// Registered as an input in the DuplexConnection.
     public Subscriber<Payload>,
     /// Receives signals about connection writability.
-    public Subscription {
+    public Subscription,
+    /// The ConnectionAutomaton instance is shared between streams and also
+    /// ReactiveSocket class, using refcounting for lifetime management
+    public IntrusiveDeleter {
```
stupaq-fb: Ugh, you really don't want everyone to call increment/decrementRefCount, please make it protected inheritance.

reactivesocket-cpp/src/ConnectionAutomaton.h
```diff
@@ -91,7 +93,11 @@ class ConnectionAutomaton :
   void endStream(StreamId streamId, StreamCompletionSignal signal);
   /// @}
 
+  void close();
```
stupaq-fb: What does that do?

stupaq-fb:
We need changes to automata as well, despite documentation saying so, there is no smart pointer to the Connection and automata currently assume that ConnectionAutomaton::endStream is idempotent... but it no longer is after your change.

@yschimke, @lehecka I have second thoughts. What is the use case? Destroying ReactiveSocket from, for example Channel's Subscriber::onComplete callback is equivalent(ish) to destroying ThriftServer from an RPC method, which guarantees a crash, last time I checked. This is more complicated change than in this PR, maybe we don't need it? I don't know :/

yschimke:
From the java and spec side there doesn't appear to be any limits on who or when you can shutdown a reactivesocket. e.g. shutdown method in https://github.com/ReactiveSocket/reactivesocket-java/blob/master/src/main/java/io/reactivesocket/ReactiveSocket.java

stupaq-fb:
I absolutely agree that it would be ideal to make any code allowed by compiler not crash, but it's hard in C++, hence the question. :)

Point taken though, since you've written a code that triggers unsupported shutdown sequence, we have a use case for this...

stupaq-fb:
@lehecka Do you mind if I experiment with plain old shared_ptr here?

I remember saying yesterday (in an offline conversation) that there is something wrong with using it, but I haven't taken into account that there is no flow control between stream automata and ConnectionAutomaton. That actually simplifies things a lot, as spontaneous communication is unidirectional.

stupaq-fb:
Actually, the docs even say it's idempotent https://github.com/ReactiveSocket/reactivesocket-cpp/blob/master/reactivesocket-cpp/src/ConnectionAutomaton.h#L89. I will make sure this holds true in my experiment. I don't want to overcomplicate stream automata.

Issue #34, opened 2016-05-17
Allow destruction of ReactiveSocket instance from anywhere


stupaq-fb:
...including any callbacks that are associated with the instance.

@yschimke @lehecka What should be the semantics over here?
1) Destroying ReactiveSocket kills all existing streams.
2) Destroying ReactiveSocket doesn't kill existing streams, but prevents any new streams from being accepted or requested.

yschimke:
I don't have a strong opinion yet.  For me, its a synchronous or at least observable (Future?) operation that cleanly shuts down all streams and releases resources.  But I don't know that is agreed and may not make sense when invoked from a callback.

stupaq-fb:
How about that. After destructor of ReactiveSocket finishes we guarantee:
* no new streams can be accepted,
* no new streams can be requested (trivial: ReactiveSocket instance is gone),
* all existing streams are closed and corresponding Subscriptions/Subscribers will _eventually_ receive terminal signals.

So it's not synchronous, but each Subscriber/Subscription on an open stream will learn about the failure as soon as possible. If we were to make d'tor synchronous, we'd likely deadlock if it's called from Subscriber/Subscription signal handler.

We don't need to cast the semantics in stone, but we should at least agree the v0 semantics of this operation "makes sense".

The reason things crash currently is simple: destroying ReactiveSocket instance from signal handler is not something I considered to be a valid behaviour. By an analogy to Thrift, destroying ThriftServer from RPC callback guarantees a crash/deadlock.

yschimke:
That sounds like a reasonable plan to me.  I think we are just learning what makes sense as we go now, and that will keep evolving as we continue building and testing.

stupaq-fb:
> I think we are just learning what makes sense as we go now, and that will keep evolving as we continue building and testing.

Absolutely agree. I just want us to know what to expect for now. We'll be able to expect a reasonable behaviour after a few iterations.

PR #37, opened 2016-05-17
ReactiveSocket d'tor can be called from signal handler of a stream.

stupaq-fb:
DO NOT MERGE!

This is still a WIP, to do:
* write a test to verify it actually works
* provide stronger guarantees to DuplexConnection, so that it's easier to get right.

### DIFF
--- a/reactivesocket-cpp/src/ConnectionAutomaton.cpp
+++ b/reactivesocket-cpp/src/ConnectionAutomaton.cpp
@@ -33,6 +33,15 @@ void ConnectionAutomaton::connect() {
   connection_->setInput(*this);
 }
 
+void ConnectionAutomaton::disconnect() {
+  // Send terminal signals to the DuplexConnection's input and output before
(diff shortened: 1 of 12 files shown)


stupaq-fb:
This fixes #34.

stupaq-fb:
So far, I know that:
* it's safe, since we hold a strong reference to ConnectionAutomaton, i.e. we cannot possibly dereference dangling pointer/reference,
* ConnectionAutomaton doesn't hold strong references to AbstractStreamAutomata, so there is no leak there -- I actually verified that d'tor of ConnectionAutomaton is invoked,
* as a bonus, ConnectionAutomaton::endStream is still idempotent, which simplifies AbstractStreamAutomata.

stupaq-fb:
GOOD TO MERGE (after review)

@yschimke, take a look at the test, please.

yschimke:
The test makes sense to me, but and other changes all looked appropriate given the pointer changes.

Issue #12, opened 2016-05-06
Separate compilation for each automaton


stupaq-fb:
They are all instantiated in ReactiveSocket.cpp right now, I believe.

lehecka:
what is your motivating factor? compile time? Or do we plan on using them separately?

stupaq-fb:
More like a general refactoring how we instantiate automata. That'd inline all of the mixins in one function and save a bunch of kB of code, I suppose.

While doing it, I can also take care of not instantiating all templates in the same CU, so `-j` we pass to make actually does something.

yschimke:
seems low enough priority we will never do this

PR #40, opened 2016-05-18
Travis Build

benjchristensen:
Continuous builds on Travis CI

- this pulls a pre-compiled version of folly from Bitbucket: https://bitbucket.org/reactivesocket/reactivesocket-cpp-dependencies/src
- it updates libs via apt-get
- it builds reactivesocket-cpp
- it runs ./ReactiveSocketTest

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,17 @@
 sudo: required
 dist: trusty
 
-env:
-  global:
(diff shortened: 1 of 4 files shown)


benjchristensen:
Note that I have not yet tested whether it correctly shows as "failed" if a unit test fails. I'm assuming ./ReactiveSocketTest will return a value other than 0 and thus fail the build.

stupaq-fb:
Yep, I believe it to be the case.

Could you update `make` command to `make -j8` or something? It seems that build is slow because of that.

benjchristensen:
This pull request successfully shows how Travis does the build and shows as "green". The log can be seen here: https://travis-ci.org/ReactiveSocket/reactivesocket-cpp/builds/131003661

benjchristensen:
men4a

Yak shaved.

stupaq-fb:
Great!

benjchristensen:
Build time with `make`: 1 minutes 43 seconds
Build time with `make -j8`: 1 minute 42 seconds
Build time without precompiled folly: > 10 minutes

stupaq-fb:
#12 will speed things up, assuming we actually have multiple cores to build (I don't know how big the container is).

PR #41, opened 2016-05-18
Fixed use-after-free in ReactiveSocketTest.

stupaq-fb:


stupaq-fb:
@benjchristensen I "tested" whether contbuild fails when a test fails :)

benjchristensen:
All tests passed, were you expecting one to fail?

stupaq-fb:
After my last commit the test was failing, this PR fixed it.

benjchristensen:
Ah nice, so it's working as we want it to.

PR #43, opened 2016-05-18
Matrix Build: gcc and clang

benjchristensen:
Build on both gcc and clang


benjchristensen:
This works and is ready to merge.

Issue #38, opened 2016-05-17
remove top level namespace


yschimke:
I'm happy to make the change, but want to agree the change first.

Are we ok with just dropping lithium and using reactivestreams and reactivesocket?

stupaq-fb:
Strongly approve.

lehecka:
what is left for this issue to be able to close it? @yschimke you pushed your change to remove it already..

PR #45, opened 2016-05-20
Remove lithium namespace

yschimke:
After this, the only lithium references are in TARGET files

https://github.com/ReactiveSocket/reactivesocket-cpp/issues/38


stupaq-fb:
Cool, did you run `./devtools/format_all.sh` successfully?

yschimke:
Of course, what sort of 2 bit huckster do you take me for?

PR #54, opened 2016-06-21
tcp impl

yschimke:
Working cpp -> cpp and java -> cpp.  Would like to land this as is, and ass more tests in a second PR.

later
- mocked tests
- nice constructor e.g. TcpDuplexConnection::connect(host, port), TcpDuplexConnection::list(port, conectionCallback).  This is probably creating a promise, registering a callback that completes the promise and returning the future that can be chained.  Will do in a later change.


yschimke:
Current status

```
$ ./tcpserver
server listening on [::]:9898
connectionAccepted[::1]:61901
ServerRequestHandler.handleRequestSubscription from client
```

```
$ ./tcpclient
attempting connection to [::]:9898
PrintSubscriber.onNext from server
```

reactivesocket-cpp/test/simple/CancelSubscription.h
```diff
@@ -0,0 +1,22 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#pragma once
+
+
+#include <reactivesocket-cpp/src/ReactiveStreamsCompat.h>
+#include <reactivesocket-cpp/src/mixins/IntrusiveDeleter.h>
+
+namespace reactivesocket {
+    class CancelSubscription
```
lehecka: The name CancelSubscription suggests that there will be something cancelled. But looking at the implementation, it is an empty implementation. Can you consider renaming this to EmptySubscription or NullSubscription?

reactivesocket-cpp/src/tcp/TcpDuplexConnection.h
```diff
@@ -0,0 +1,93 @@
...
+
+        void readErr(const folly::AsyncSocketException &ex) noexcept override;
+
+        bool isBufferMovable() noexcept override;
+
+        void readBufferAvailable(std::unique_ptr<folly::IOBuf> readBuf) noexcept override;
+    private:
+        folly::IOBufQueue readBuffer_{folly::IOBufQueue::cacheChainLength()};
+        std::unique_ptr<TcpOutputSubscriber> outputSubscriber_;
+        reactivesocket::Subscriber<Payload> *inputSubscriber_{nullptr};
```
lehecka: the SubscriberPtr helper class, instead of the raw pointer, would make sure that you will always call onComplete no matter how you will destroy this class.

reactivesocket-cpp/src/tcp/TcpDuplexConnection.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+    void TcpDuplexConnection::writeSuccess() noexcept {
+    }
+
+    void TcpDuplexConnection::writeErr(size_t bytesWritten, const AsyncSocketException& ex) noexcept {
+      std::cout << "TODO writeErr" << bytesWritten << ex.what() << "\n";
+    }
+
+    void TcpDuplexConnection::getReadBuffer(void **bufReturn, size_t *lenReturn) noexcept {
+      std::tie(*bufReturn, *lenReturn) = readBuffer_.preallocate(4096, 4096);
```
lehecka: are you sure that this expression is going to set the **bufReturn and lenReturn output params?
From what I am reading about std::tie, you are only going to create a local variable of type std::tuple which components are going to be set by the assignment operator.
I thing the *bufReturn and *lenReturn are going to be values and not references to values, which you probably intended.
I would verify this in the debugger or at least with some logging.

reactivesocket-cpp/src/streams/AllowanceSemaphore.h
```diff
@@ -59,6 +59,10 @@ class AllowanceSemaphore {
     return limit;
   }
 
+  ValueType current() {
```
lehecka: ValueType current() const {

reactivesocket-cpp/src/tcp/TcpDuplexConnection.cpp
lehecka: it is not clear for me whether you are allocating a buffer and still managing the lifetime via IOBufQueue or if you are giving this up. Is readBufferAvailable method giving you the same memory back?

reactivesocket-cpp/src/tcp/TcpDuplexConnection.h
```diff
@@ -0,0 +1,93 @@
...
+
+namespace reactivesocket {
+    class TcpSubscriptionBase: public virtual ::reactivesocket::IntrusiveDeleter,
+                               public ::reactivesocket::Subscription {
+    public:
+        ~TcpSubscriptionBase() = default;
+
+        // Subscription methods
+        void request(size_t n) override {
+            LOG(INFO) << "TcpSubscriptionBase.request " << n;
```
lehecka: in the TODO comments, I would add
+ implement proper request(n) logic.

it looks like you are ignoring it for now.

reactivesocket-cpp/src/tcp/TcpDuplexConnection.h
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        void cancel() override {
+            LOG(INFO) << "TcpSubscriptionBase.cancel";
```
lehecka: should you close the tcp connection or send EOM over when cancel() is called?

reactivesocket-cpp/test/tcp/TcpClient.cpp
```diff
@@ -0,0 +1,112 @@
...
+  });
+
+  std::unique_ptr<ReactiveSocket> reactiveSocket;
+  Callback callback;
+
+  eventBase.runInEventBaseThreadAndWait([&callback, &reactiveSocket, &eventBase]() {
+      folly::AsyncSocket::UniquePtr socket(new folly::AsyncSocket(&eventBase));
+
+      folly::SocketAddress addr;
+      addr.setFromLocalIpPort(FLAGS_address);
```
lehecka: if the address value will be "host:port", I think setFromLocalIpPort wont work.
(you can use the constructor of SocketAddress with a boolean flag to do the name resolution (localhost))

reactivesocket-cpp/test/tcp/TcpServer.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include <reactivesocket-cpp/src/framed/FramedDuplexConnection.h>
+
+using namespace ::testing;
+using namespace ::reactivesocket;
+using namespace ::folly;
+
+DEFINE_string(address, "9898", "host:port to listen to");
+
+namespace {
+    class ServerSubscription
```
lehecka: formatting seems to be pretty inconsistent in this diff. I dont' think we are using class indentation in this project.

reactivesocket-cpp/test/tcp/TcpServer.cpp
```diff
@@ -0,0 +1,141 @@
...
+        virtual void connectionAccepted(int fd, const SocketAddress &clientAddr) noexcept override {
+          std::cout << "connectionAccepted" << clientAddr.describe() << "\n";
+
+          auto socket = folly::AsyncSocket::UniquePtr(new AsyncSocket(&eventBase_, fd));
+
+          std::unique_ptr<DuplexConnection> connection = folly::make_unique<TcpDuplexConnection>(std::move(socket));
+          std::unique_ptr<DuplexConnection> framedConnection = folly::make_unique<FramedDuplexConnection>(std::move(connection));
+          std::unique_ptr<RequestHandler> requestHandler = folly::make_unique<ServerRequestHandler>();
+
+          reactiveSocket_ = ReactiveSocket::fromClientConnection(std::move(framedConnection), std::move(requestHandler));
```
lehecka: arent you supposed to call fromServerConnection?

reactivesocket-cpp/test/tcp/TcpServer.cpp
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        virtual void acceptError(const std::exception &ex) noexcept override {
+          std::cout << "acceptError" << ex.what() << "\n";
```
lehecka: in case of the error, do you want to terminate the event loop?
At least the program will terminate..

lehecka:
did you have to fix the reactive-streams-cpp submodule thing to get this to compile?

reactivesocket-cpp/test/tcp/TcpServer.cpp
yschimke: Is there a good formatter for the github project? clang-format?

reactivesocket-cpp/src/tcp/TcpDuplexConnection.h
yschimke: I wasn't certain here to be honest, is that too early?  Should the socket get closed by a destructor once we decide to cancel at the protocol level?

reactivesocket-cpp/src/tcp/TcpDuplexConnection.h
yschimke: I also wasn't sure about this, can the Tcp (unframed) implementation be correctly implemented with request n logic?  I was assuming at the tcp stream layer we would send all data as fast as possible, and let the framed layers do the flow control.

reactivesocket-cpp/src/tcp/TcpDuplexConnection.cpp
yschimke: Took this from here.

https://phabricator.intern.facebook.com/diffusion/FBS/browse/master/fbcode/folly/io/async/AsyncUDPServerSocket.h$146?view=highlighted

My understanding is this allocates a buffer and returns it to the caller.  The caller calls back saying it read X into the buffer and we postallocate and slice off the buffer and return.

reactivesocket-cpp/test/tcp/TcpServer.cpp
yschimke: I installed clang-format as an external tool in CLion.  job done.

reactivesocket-cpp/test/tcp/TcpClient.cpp
yschimke: thanks, works well.

reactivesocket-cpp/src/streams/AllowanceSemaphore.h
yschimke: removed, was using for debugging

reactivesocket-cpp/src/tcp/TcpDuplexConnection.h
yschimke: Does that still obey the setInput docs "The connection MUST NOT assume an ownership of
  /// provided Subscriber."

yschimke:
@lehecka no, I think I'm getting that from an installed version in my system

reactivesocket-cpp/src/tcp/TcpDuplexConnection.h
lehecka: Yes. SubscriberPtr is not like unique_ptr, it doesn't express ownership. I think the Ptr naming is just confusing. It will make sure you will do the right unsubscribe handshake.

reactivesocket-cpp/src/tcp/TcpDuplexConnection.h
lehecka: I think you should do something here. Just imagine someone calls cancel() and keeps the instance for longer.
In your case you will just ignore the cancel call and will keep sending/receiving data. It feels to me that's not intended.

reactivesocket-cpp/src/tcp/TcpDuplexConnection.cpp
lehecka: what memory are you getting in the param of readBufferAvailable? Isnt it the one you preallocated? Or was it allocated by the socket implementation? Maybe it doesn't matter here since IOBuf is just a structure with a pointer to the ref counted buffer and destroying IOBuf doesn't mean it will destroy the ref counted buffer. Please excuse the lack of knowledge of the implementation detail.
How is isBufferMovable used?

reactivesocket-cpp/src/tcp/TcpDuplexConnection.h
yschimke: So my understanding is the call chain will be

FramedDuplexConnection::~FramedDuplexConnection ->
FramedWriter::cancel() ->
TcpSubscriptionBase::cancel()

So I think you are right, closing here seems safe to do.

reactivesocket-cpp/src/tcp/TcpDuplexConnection.cpp
yschimke: "how is babby formed"  ;)

https://our.intern.facebook.com/intern/codex/symbol/fbcode:folly::AsyncReader/ReadCallback/

Even though isBufferMovable is true, I don't think this transport supports it, so we still get called via readDataAvailable.  the data has been written to the buffer we allocated in getReadBuffer.

I think I should fail if inputSubscriber_ isn't set, otherwise nothing is setup to send the data once it is available.

lehecka:
I am going to merge it now as the diff is pretty large already. I am not sure if the travis build is broken already or by this diff, but we can fix it in another iteration.

PR #56, opened 2016-06-23
Framed support classes

yschimke:
asan turned off as its failing when on, need to discuss.

I also added a couple of EXPECT_CALL to avoid noisy warnings. I know it can work without.

### DIFF
--- a/reactivesocket-cpp/src/framed/FramedWriter.h
+++ b/reactivesocket-cpp/src/framed/FramedWriter.h
@@ -1,0 +1,31 @@
+// Copyright 2004-present Facebook.  All rights reserved.
+#pragma once
+
+#include <folly/ExceptionWrapper.h>
+#include <reactivesocket-cpp/src/streams/SmartPointers.h>
+#include "reactivesocket-cpp/src/Payload.h"
+#include "reactivesocket-cpp/src/ReactiveStreamsCompat.h"
+
+namespace reactivesocket {
+
+class FramedWriter : public reactivesocket::Subscriber<Payload>, public reactivesocket::Subscription {
+ public:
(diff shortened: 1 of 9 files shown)


stupaq-fb:
If ASAN build is failing then it's probably not a good idea to merge it. Do you have a paste with ASAN trace?

yschimke:
It doesn't give much that I know how to use

```
==75025==ERROR: AddressSanitizer: heap-use-after-free on address 0x61400000ffe8 at pc 0x00010eb33c91 bp 0x7fff510e0450 sp 0x7fff510e0448
READ of size 8 at 0x61400000ffe8 thread T0
```

And was failing before I made any other changes.

stupaq-fb:
Do you mean master was failing? It looked healthy...

yschimke:
I think these are fixed now, looks like only leaks in the test code and related to calling a method on the heap allocated ref counted mocks after the Object Under Test destructor had already caused the refcount to go to 0.

reactivesocket-cpp/src/framed/FramedWriter.h
```diff
@@ -0,0 +1,31 @@
...
+#pragma once
+
+#include <folly/ExceptionWrapper.h>
+#include <reactivesocket-cpp/src/streams/SmartPointers.h>
+#include "reactivesocket-cpp/src/Payload.h"
+#include "reactivesocket-cpp/src/ReactiveStreamsCompat.h"
+
+namespace reactivesocket {
+
+class FramedWriter : public reactivesocket::Subscriber<Payload>, public reactivesocket::Subscription {
```
stupaq-fb: That looks long, did you run clang-format?

yschimke: it didn't care

PR #58, opened 2016-07-01
[reactivesockets-cpp] add metadata to frames

alexstaytuned:
- Add metadata to all currently supported messages

Note: started using IOBufQueue as it's a better abstraction for chaining together many IOBuf's than just doing it on top of IOBuf.

- Correction: Setup frame doesn't have 'data', only mime types for data/metadata and then the metadata

### DIFF
--- a/reactivesocket-cpp/src/ConnectionAutomaton.cpp
+++ b/reactivesocket-cpp/src/ConnectionAutomaton.cpp
@@ -35,9 +35,9 @@ void ConnectionAutomaton::connect(bool client) {
 
   if (client) {
     // TODO set correct version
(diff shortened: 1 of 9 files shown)


yschimke:
The spec

https://github.com/ReactiveSocket/reactivesocket/blob/master/Protocol.md

and java both have metadata and data for setup

https://github.com/ReactiveSocket/reactivesocket-java/blob/380fd3dc5047717988e5c5ab0a318ca6bdddc98b/reactivesocket-core/src/main/java/io/reactivesocket/Frame.java#L240

alexstaytuned:
Closed on accident. Going to create a new one.

PR #59, opened 2016-07-01
[reactivesockets-cpp] add metadata to frames

alexstaytuned:
- Add metadata to all currently supported messages
Note: started using IOBufQueue as it's a better abstraction for chaining together many IOBuf's than just doing it on top of IOBuf.

### DIFF
--- a/reactivesocket-cpp/src/Frame.cpp
+++ b/reactivesocket-cpp/src/Frame.cpp
@@ -2,6 +2,12 @@
 
 #include "reactivesocket-cpp/src/Frame.h"
 
+#include <bitset>
+#include <memory>
+#include <ostream>
+
+#include <folly/Memory.h>
+#include <folly/Optional.h>
 #include <folly/Singleton.h>
 #include <folly/io/Cursor.h>
 
--- a/reactivesocket-cpp/src/Frame.cpp
+++ b/reactivesocket-cpp/src/Frame.cpp
@@ -90,16 +90,11 @@ folly::Optional<StreamId> FrameHeader::peekStreamId(const folly::IOBuf& in) {
   }
 }
 
-void FrameHeader::serializeInto(folly::io::Appender& app) {
-  app.writeBE(static_cast<uint16_t>(type_));
-  app.writeBE<uint16_t>(flags_);
-  app.writeBE<uint32_t>(streamId_);
-}
-
-void FrameHeader::serializeInto(folly::io::QueueAppender& app) {
-  app.writeBE(static_cast<uint16_t>(type_));
-  app.writeBE<uint16_t>(flags_);
-  app.writeBE<uint32_t>(streamId_);
+template<typename T>
+void FrameHeader::serializeInto(folly::io::detail::Writable<T>& app) {
+  app.template writeBE(static_cast<uint16_t>(type_));
+  app.template writeBE<uint16_t>(flags_);
+  app.template writeBE<uint32_t>(streamId_);
 }
 
 bool FrameHeader::deserializeFrom(folly::io::Cursor& cur) {
@@ -118,12 +113,14 @@ std::ostream& operator<<(std::ostream& os, const FrameHeader& header) {
   return os << header.type_ << "[" << flags << ", " << header.streamId_ << "]";
 }
 
-std::unique_ptr<folly::IOBuf> FrameMetadata::serialize() {
-  auto buf = FrameBufferAllocator::allocate(sizeof(uint32_t));
-  folly::io::Appender app(buf.get(), /* do not grow */ 0);
+constexpr auto MAX_META_LENGTH = std::numeric_limits<int32_t>::max();
+
+void FrameMetadata::serializeInto(folly::io::QueueAppender& app) {
+  // use signed int because the first bit in metadata length is reserved
+  assert(metadataPayload_->length() + sizeof(uint32_t) < MAX_META_LENGTH);
+
   app.writeBE<uint32_t>(static_cast<uint32_t>(metadataPayload_->length()) + sizeof(uint32_t));
-  buf->appendChain(std::move(metadataPayload_));
-  return buf;
+  app.insert(std::move(metadataPayload_));
 }
 
 bool FrameMetadata::deserializeFrom(folly::io::Cursor& cur,
@@ -144,9 +141,12 @@ bool FrameMetadata::deserializeFrom(folly::io::Cursor& cur,
 
 bool FrameMetadata::deserializeFrom(folly::io::Cursor& cur) {
   try {
-    auto length = cur.readBE<uint32_t>() - sizeof(uint32_t);
-    if (length > 0) {
-      cur.clone(metadataPayload_, length);
+    const auto length = cur.readBE<uint32_t>();
+    assert(length < MAX_META_LENGTH);
+    const auto metadataPayloadLength = length - sizeof(uint32_t);
+    
+    if (metadataPayloadLength > 0) {
+      cur.clone(metadataPayload_, metadataPayloadLength);
     } else {
       metadataPayload_.reset();
     }
@@ -166,14 +166,17 @@ std::ostream& operator<<(std::ostream& os, const folly::Optional<FrameMetadata>&
 /// @{
 Payload Frame_REQUEST_SUB::serializeOut() {
   auto queue = FrameBufferAllocator::allocateQueue();
+  const auto metadataPresent = header_.flags_ & FrameFlags_METADATA;
+  const auto bufSize = FrameHeader::kSize + sizeof(uint32_t) +
+    (metadataPresent ? sizeof(uint32_t) : 0);
   auto buf =
-    FrameBufferAllocator::allocate(FrameHeader::kSize + sizeof(uint32_t));
+    FrameBufferAllocator::allocate(bufSize);
   queue->append(std::move(buf));
   folly::io::QueueAppender app(queue.get(), /* do not grow */ 0);
   header_.serializeInto(app);
   app.writeBE<uint32_t>(requestN_);
-  if (header_.flags_ & FrameFlags_METADATA) {
-    app.insert(metadata_->serialize());
+  if (metadataPresent) {
+    metadata_->serializeInto(app);
   }
   if (data_) {
     app.insert(std::move(data_));
@@ -212,15 +215,18 @@ std::ostream& operator<<(std::ostream& os, const Frame_REQUEST_SUB& frame) {
 /// @{
 Payload Frame_REQUEST_CHANNEL::serializeOut() {
   auto queue = FrameBufferAllocator::allocateQueue();
+  const auto metadataPresent = header_.flags_ & FrameFlags_METADATA;
+  const auto bufSize = FrameHeader::kSize + sizeof(uint32_t) +
+    (metadataPresent ? sizeof(uint32_t) : 0);
   auto buf =
-    FrameBufferAllocator::allocate(FrameHeader::kSize + sizeof(uint32_t));
+    FrameBufferAllocator::allocate(bufSize);
   queue->append(std::move(buf));
   folly::io::QueueAppender app(queue.get(), /* do not grow */ 0);
 
   header_.serializeInto(app);
   app.writeBE<uint32_t>(requestN_);
-  if (header_.flags_ & FrameFlags_METADATA) {
-    app.insert(metadata_->serialize());
+  if (metadataPresent) {
+    metadata_->serializeInto(app);
   }
   if (data_) {
     app.insert(std::move(data_));
@@ -259,8 +265,9 @@ std::ostream& operator<<(std::ostream& os, const Frame_REQUEST_CHANNEL& frame) {
 
 /// @{
 Payload Frame_REQUEST_N::serializeOut() {
+  const auto bufSize = FrameHeader::kSize + sizeof(uint32_t);
   auto buf =
-      FrameBufferAllocator::allocate(FrameHeader::kSize + sizeof(uint32_t));
+      FrameBufferAllocator::allocate(bufSize);
   folly::io::Appender app(buf.get(), /* do not grow */ 0);
   header_.serializeInto(app);
   app.writeBE<uint32_t>(requestN_);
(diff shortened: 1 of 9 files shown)


reactivesocket-cpp/src/Frame.cpp
```diff
@@ -99,6 +96,12 @@ void FrameHeader::serializeInto(folly::io::Appender& app) {
   app.writeBE<uint32_t>(streamId_);
 }
 
+void FrameHeader::serializeInto(folly::io::QueueAppender& app) {
```
yschimke: nood c++ question: can this be Writable instead of QueueAppender and Appender?

reactivesocket-cpp/src/Frame.cpp
```diff
@@ -114,34 +117,83 @@ std::ostream& operator<<(std::ostream& os, const FrameHeader& header) {
   std::bitset<16> flags(header.flags_);
   return os << header.type_ << "[" << flags << ", " << header.streamId_ << "]";
 }
+
+std::unique_ptr<folly::IOBuf> FrameMetadata::serialize() {
```
yschimke: I have a concern about allocating small buffers and chaining, rather than allocating the correct size buffer at the start (after computing size) and then writing into that.  I'm interested to hear other viewpoints.

reactivesocket-cpp/src/Frame.cpp
```diff
@@ -114,34 +117,83 @@ std::ostream& operator<<(std::ostream& os, const FrameHeader& header) {
...
+  auto buf = FrameBufferAllocator::allocate(sizeof(uint32_t));
+  folly::io::Appender app(buf.get(), /* do not grow */ 0);
+  app.writeBE<uint32_t>(static_cast<uint32_t>(metadataPayload_->length()) + sizeof(uint32_t));
```
yschimke: nit: I doubt we will hit this, but seems prudent to assert that 

metadataPayload_->length() <= std::numeric_limits<int32_t>::max() - 1

reactivesocket-cpp/src/Frame.cpp
```diff
@@ -351,49 +424,55 @@ std::ostream& operator<<(std::ostream& os, const Frame_KEEPALIVE& frame) {
...
+    return false;
+  }
+  try {
+    version_ = cur.readBE<uint32_t>();
+    keepaliveTime_ = cur.readBE<uint32_t>();
+    maxLifetime_ = cur.readBE<uint32_t>();
+  } catch (...) {
+    return false;
+  }
+  // TODO decode mime types
```
yschimke: Previous code ignored mime types and metadata and assumed remaining was data.  This was wrong, but sort of safe. This reads a length field for metadata from the mimetype (1 byte of mimetype length, 3 bytes of the mimetype) which seems wrong and unsafe.

Could you check the mimetype lengths and skip those at least.

yschimke: I can fix this issue since the case I mentioned was already badly broken.  But can you disable reading metadata for setup frames until I do.  Or fix both in this diff.  Your call.

reactivesocket-cpp/src/Frame.cpp
alexstaytuned: Got it to work with Writable through templates. I also had a overload of serializeInto that need not be there - removed it.

reactivesocket-cpp/src/Frame.cpp
alexstaytuned: I'll add these.

reactivesocket-cpp/src/Frame.cpp
alexstaytuned: Changed the logic to not allocate a separate buffer just for the length.

reactivesocket-cpp/src/Frame.h
```diff
@@ -121,14 +125,16 @@ class FrameMetadata {
   explicit FrameMetadata()
   : metadataPayload_(folly::make_unique<folly::IOBuf>()) {}
 
-  std::unique_ptr<folly::IOBuf> serialize();
+  void serializeInto(folly::io::QueueAppender& app);
```
alexstaytuned: note that we cannot use Writable<T> here because we need a method only defined on QueueAppender.

reactivesocket-cpp/src/Frame.h
```diff
@@ -100,13 +104,38 @@ std::ostream& operator<<(std::ostream&, const FrameHeader&);
...
+                              folly::Optional<FrameMetadata>& metadata);
+  bool deserializeFrom(folly::io::Cursor& cur);
+
+  uint32_t size();
+
+  std::unique_ptr<folly::IOBuf> metadataPayload_;
+};
+std::ostream& operator<<(std::ostream&, const folly::Optional<FrameMetadata>&);
+
+static const folly::Optional<FrameMetadata> FrameMetadata_EMPTY { folly::none };
```
lehecka: this constant is not used anywhere
also I would suggest to make it a static member variable of GrameMetadata class.

reactivesocket-cpp/src/Frame.h
```diff
@@ -122,20 +151,23 @@ class Frame_REQUEST_SUB {
...
-      : Frame_REQUEST_SUB(streamId, flags, 0, std::move(data)) {}
+  Frame_REQUEST_SUB(StreamId streamId, FrameFlags flags, folly::Optional<FrameMetadata> metadata, Payload data)
+      : Frame_REQUEST_SUB(streamId, flags, 0, std::move(metadata), std::move(data)) {}
 
   Payload serializeOut();
   bool deserializeFrom(Payload in);
 
   FrameHeader header_;
   uint32_t requestN_;
+  folly::Optional<FrameMetadata> metadata_;
```
lehecka: It feels to me that using folly::Optional in this case is unnecessary. FrameHeader is storing a unique_ptr which can express nullability.
I would suggest removing Optional from this diff.

reactivesocket-cpp/src/automata/ChannelRequester.cpp
```diff
@@ -40,6 +40,7 @@ void ChannelRequesterBase::onNext(Payload request) {
           streamId_,
           flags,
           static_cast<uint32_t>(initialN),
+          FrameMetadata(),
```
lehecka: FrameMetadata::kEmpty

reactivesocket-cpp/src/Frame.cpp
```diff
@@ -114,34 +117,91 @@ std::ostream& operator<<(std::ostream& os, const FrameHeader& header) {
...
+  const auto metadataPresent = header_.flags_ & FrameFlags_METADATA;
+  const auto bufSize = FrameHeader::kSize + sizeof(uint32_t) +
+    (metadataPresent ? sizeof(uint32_t) : 0);
   auto buf =
-      FrameBufferAllocator::allocate(FrameHeader::kSize + sizeof(uint32_t));
-
-  folly::io::Appender app(buf.get(), /* do not grow */ 0);
+    FrameBufferAllocator::allocate(bufSize);
+  queue->append(std::move(buf));
+  folly::io::QueueAppender app(queue.get(), /* do not grow */ 0);
```
lehecka: I would suggest full names of the variables. app-> appender

reactivesocket-cpp/src/Frame.cpp
```diff
@@ -30,6 +29,10 @@ std::unique_ptr<folly::IOBuf> FrameBufferAllocator::allocateBuffer(
   return folly::IOBuf::createCombined(size);
 }
 
+std::unique_ptr<folly::IOBufQueue> FrameBufferAllocator::allocateQueue() {
```
lehecka: what is the value of this function?
In the case of allocateBuffer, we can actually override the behavior and allocate extra bytes (for message frame size). In the case of allocateQueue, do we plan on doing anything extra? If the answer is no, then I would suggest removing this method and allocate IOBufQueue on stack in the serializeOut methods

reactivesocket-cpp/src/Frame.cpp
```diff
@@ -114,34 +117,91 @@ std::ostream& operator<<(std::ostream& os, const FrameHeader& header) {
   std::bitset<16> flags(header.flags_);
   return os << header.type_ << "[" << flags << ", " << header.streamId_ << "]";
 }
+
+constexpr auto MAX_META_LENGTH = std::numeric_limits<int32_t>::max();
```
lehecka: static constexpr auto kMAX_META_LENGTH
constants should start with k-prefix
symbols which are not relevant outside of the compilation unit should be declared as static

reactivesocket-cpp/src/Frame.cpp
```diff
@@ -114,34 +117,91 @@ std::ostream& operator<<(std::ostream& os, const FrameHeader& header) {
...
+  app.writeBE<uint32_t>(static_cast<uint32_t>(metadataPayload_->length()) + sizeof(uint32_t));
+  app.insert(std::move(metadataPayload_));
+}
+
+bool FrameMetadata::deserializeFrom(folly::io::Cursor& cur,
+                                    const FrameFlags& flags,
+                                    folly::Optional<FrameMetadata>& metadata) {
+  if (flags & FrameFlags_METADATA) {
+    FrameMetadata m;
+    if (! m.deserializeFrom(cur)) {
```
lehecka: nit (formatting): if (!m.deserializeFrom...)

reactivesocket-cpp/src/Frame.h
```diff
@@ -8,15 +8,19 @@
 
 /// Needed for inline d'tors of frames.
 #include <folly/io/IOBuf.h>
+#include <folly/io/IOBufQueue.h>
+#include <folly/Memory.h>
+#include <folly/Optional.h>
+#include <folly/io/Cursor.h>
```
lehecka: you don't have to pull in this header if you are just declaring methods with the folly::Cursor symbol.
it is declared on line 22.

you will need to include the header file in the cpp file where you will call methods off of folly::Cursor.

reactivesocket-cpp/src/Frame.h
```diff
@@ -8,15 +8,19 @@
 
 /// Needed for inline d'tors of frames.
 #include <folly/io/IOBuf.h>
+#include <folly/io/IOBufQueue.h>
+#include <folly/Memory.h>
```
lehecka: is this headed needer in the Frames.h header file?
Can you include it in the cpp file only?

reactivesocket-cpp/src/Frame.cpp
alexstaytuned: Is the convention we're using for constants kUPPERCASE or kCamelCase?

reactivesocket-cpp/src/Frame.h
alexstaytuned: FrameMetadata::kEmpty would have to be copied in the constructors, and because it has a unique_ptr member, copy construction is disabled by default. Does it mean I need to use a && reference or is there a simpler way to pass it around?

alexstaytuned: (With folly::optional, making a constant was easy because copying an optional is easy.)

reactivesocket-cpp/src/Frame.cpp
alexstaytuned: It's there for consistency and to keep allocations together. Otherwise the logic of providing queue options would have to get repeated unnecessarily. Since it's not detrimental performance-wise, I see no harm adding this utility function.

reactivesocket-cpp/src/Frame.h
alexstaytuned: Removed some headers.

reactivesocket-cpp/src/Frame.h
```diff
@@ -5,15 +5,15 @@
 #include <iosfwd>
 #include <limits>
 
-#include <folly/Optional.h>
+/// Needed for inline d'tors of frames.
+#include <folly/io/IOBuf.h>
+#include <folly/io/IOBufQueue.h>
```
alexstaytuned: the two includes above seem to be necessary for the gcc build, otherwise resulting in [these](https://travis-ci.org/ReactiveSocket/reactivesocket-cpp/jobs/142580658) errors

reactivesocket-cpp/src/Frame.h
lehecka: you can define a static method:
static FrameMetadata FrameMetadata::empty();

reactivesocket-cpp/src/Frame.cpp
lehecka: I would suggest keeping this function only if it provides some value. I understand the argument about consistency, but for the reader, its easier to see a local variable on the stack than calling into a function.
You are also adding another allocation for every frame because you are returning unique_ptr<IOBufQueue> which is unnecessary.
Having the IOBufQueue as a local variable on the stack would be enough, I think.

reactivesocket-cpp/src/Frame.cpp
lehecka: good catch.
its kCamelCase
https://our.intern.facebook.com/intern/wiki/CppStyle

reactivesocket-cpp/src/Frame.h
lehecka: that looks like some cpp file needs to include logging.h
its ok. we can fix this in another diff

reactivesocket-cpp/src/Frame.h
```diff
@@ -4,19 +4,19 @@
...
 
 /// Needed for inline d'tors of frames.
 #include <folly/io/IOBuf.h>
+#include <folly/io/IOBufQueue.h>
 
 #include "reactivesocket-cpp/src/Payload.h"
 
 namespace folly {
 template <typename V>
 class Optional;
```
lehecka: I think you dont need this anymore..

reactivesocket-cpp/src/Frame.cpp
```diff
@@ -10,7 +10,6 @@
 #include <folly/Optional.h>
```
lehecka: I think you dont need this anymore..

reactivesocket-cpp/src/Frame.cpp
```diff
@@ -230,46 +300,63 @@ std::ostream& operator<<(std::ostream& os, const Frame_REQUEST_N& frame) {
...
-  header_.serializeInto(app);
+  auto queue = FrameBufferAllocator::allocateQueue();
+  const auto metadataPresent = header_.flags_ & FrameFlags_METADATA;
+  const auto bufSize = FrameHeader::kSize +
+    (metadataPresent ? sizeof(uint32_t) : 0);
+  auto buf = FrameBufferAllocator::allocate(bufSize);
+  queue->append(std::move(buf));
+  folly::io::QueueAppender appender(queue.get(), /* do not grow */ 0);
+  header_.serializeInto(appender);
+  if (metadataPresent && metadata_.metadataPayload_) {
```
lehecka: dont you have to write 0 to the output buffer if metadataPresent and metadata_.metadataPayload_ == nullptr?

I would put the check for metadata_.metadataPayload_ into the metadata_::serializeInto method

alexstaytuned: The "&& metadata_.metadataPayload_" part is only a precaution; the invariant is that we never have metadataPresent != metadata_.metadataPayload_ == nullptr. I used to only have metadataPresent check, but it seemed like a good idea to check both.

I will move the checks into the serialization method.

lehecka: if its an invariant then we should use CHECK(metadata_.metadataPayload_ != nullptr) inside the serialization method.
Can we also verify the invariant in the constructors of the frames?

PR #66, opened 2016-07-08
[reactivesocket-cpp] add fire-and-forget frame

alexstaytuned:
Add support for Fire-and-Forget frames.


src/automata/FireAndForgetRequester.h
```diff
@@ -0,0 +1,55 @@
...
+#include "src/mixins/StreamIfMixin.h"
+
+namespace reactivesocket {
+
+/// Implementation of an automaton that represents a Fire-and-Forget requester.
+/// Note that it's simpler than other automata due to the inherent throwaway
+/// simplicity of Fire-and-Forget semantics. Stream IDs are not tracked.
+///
+/// TODO: we need to add logging similar to other *Requester automata.
+class FireAndForgetRequester: public Subscription {
```
alexstaytuned: For discussion:

I opted out of mixins for FireAndForgetRequester because of its simple semantics. The logic follows the logic in the Java implementation. Are we missing certain properties that would require the mixins? Do we need ExecutorMixin?

src/automata/FireAndForgetRequester.cpp
```diff
@@ -0,0 +1,25 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#include "FireAndForgetRequester.h"
+
+namespace reactivesocket {
+
+void FireAndForgetRequester::request(size_t n) {
```
stupaq-fb: Can you run clang-format, this formatting seems strange.

src/automata/FireAndForgetRequester.h
stupaq-fb: I imagine this could be implemented without ExecutorMixin.

Who is managing lifetime of this object?

alexstaytuned: Nothing is at the moment, I didn't realize that even though we return a reference to the Subscription, we are managing the lifetime using IntrusiveDeleter by ourselves. I'm thinking of adding a dependency on IntrusiveDeleter and calling decrementRefCount() after we send the fire-and-forget request. Makes sense?

stupaq-fb: It looks to me that FNF can be handled without any state on the requester or responder side. This is a completely asynchronous communication and we do not expect to receive anything from the other end. The server, according to the signature of `RequestHandler::handleFireAndForgetRequest` cannot even respond on the same stream.

alexstaytuned: (I should preface this thread with the disclaimed that I'm a beginner in C++)

On the client side, we only need FireAndForgetRequester to exist between onSubscribe returning an instance of FireAndForgetRequester, and the client calling request(N). Since we're not using a unique_ptr here, can we avoid ref-counting and still manage the lifetime of the subscription?

stupaq-fb: > and the client calling request(N)

It's not clear to me why would the client call it. What does that mean, if we never deliver anything on that subscription?

Why are you concerned by ref counting? It's thread-unsafe, hence as fast as any other implementation of memory management over here.

src/ReactiveSocket.cpp
```diff
@@ -75,6 +76,20 @@ void ReactiveSocket::requestSubscription(
   automaton->start();
 }
 
+void ReactiveSocket::requestFireAndForget(
+    Payload request,
+    Subscriber<Payload>& responseSink) {
```
yschimke: Is there a C++ convention for a non instantiable type?  something to indicate that onNext can never be called? Scala would be Nothing.

Or is the benefit of using Payload that we can use the same subscriber classes?

src/RequestHandler.h
```diff
@@ -23,5 +23,9 @@ class RequestHandler {
   virtual void handleRequestSubscription(
       Payload request,
       Subscriber<Payload>& response) = 0;
+
+  /// Handles a new fire-and-forget request sent by the other end.
+  virtual void handleFireAndForgetRequest(
+      Payload request) = 0;
 };
```
alexstaytuned: TODO in this PR: looks like response needs to be provided here.

test/tcp/TcpServer.cpp
```diff
@@ -62,6 +62,10 @@ class ServerRequestHandler : public RequestHandler {
     response.onNext(folly::IOBuf::copyBuffer("from server2"));
     response.onComplete();
   }
+
+  void handleFireAndForgetRequest(Payload request) override {
+    // ignore for now
```
yschimke: Can you just output that the request was received.

src/automata/FireAndForgetRequester.h
```diff
@@ -0,0 +1,47 @@
...
+
+
+namespace reactivesocket {
+
+/// Implementation of an automaton that represents a Fire-and-Forget requester.
+/// Note that it's simpler than other automata due to the inherent throwaway
+/// simplicity of Fire-and-Forget semantics. Stream IDs are not tracked.
+///
+/// TODO: we need to add logging similar to other *Requester automata.
+class FireAndForgetRequester : public Subscription {
```
yschimke: Does this need to be a Subscription? Can this type just have onError and onComplete?

yschimke:
This is looking really good.

src/automata/FireAndForgetRequester.h
alexstaytuned: It's not intuitive to me either why we need to go through the subscriptions/request(N) in fire-and-forget. I'm going by what the Java implementation does: https://github.com/ReactiveSocket/reactivesocket-java/blob/master/reactivesocket-core/src/main/java/io/reactivesocket/internal/Requester.java#L175. (there's also an example in tests https://github.com/ReactiveSocket/reactivesocket-java/blob/master/reactivesocket-core/src/test/java/io/reactivesocket/ReactiveSocketTest.java#L416).

src/automata/FireAndForgetRequester.h
alexstaytuned: The answer to this question largely depends on the discussion we're having with Mateusz above. LMK what you think about that.

src/ReactiveSocket.cpp
```diff
@@ -75,6 +76,20 @@ void ReactiveSocket::requestSubscription(
   automaton->start();
 }
 
+void ReactiveSocket::requestFireAndForget(
```
stupaq-fb: Why does this method allow caller to provide a Subscriber? What that subscriber mean and what kind of data does it expect? Since the signatures of requestFNF and requestSubscription are the same, what is the difference between them?

I wouldn't be concerned by compatibility with Java. ReactiveStreams are not "compatible", because they need to work in a language without a GC. ReactiveSocket intentionally has control flow inversion, as that limits laziness and makes memory management easier and allows to keep more stuff on the stack. Most important to me is for this code to be idiomatic (in C++).

src/automata/FireAndForgetRequester.h
stupaq-fb: I don't see why this class exists in the first place ;D. FNF is second-class member of the protocol -- we happen to have some transport connection, would be nice to pipe some unimportant data over it as well,  instead of establishing another one. What kind of a state per FNF message do we want to maintain if the very name says to forget about this request/message ASAP?

src/ReactiveSocket.cpp
alexstaytuned: It's one thing to not be compatible in the internals. It's another thing to not be compatible from external API point of view. Looking at the Java unit test I linked to, doing something different in C++ implementation would mean that the method named the same (reactiveSocket.fireAndForget()) behaves differently across languages (in Java, one receives a subscription and needs to call request(), in C++ it's immediately dispatched).

To you and Yuri's comments about why FireAndForgetRequester exists - it's only to be on par with what we do in Java. If we make a decision that we want to dispatch immediately, then I agree that it's not necessary.

stupaq-fb: I don't see any value in that yet-another-layer of laziness (that's why I stripped it from the API). If someone wants to be N-times lazy -- lambdas are their friend. We can later on provide a higher-level api, which has no control flow inversion and is as lazy as needed, at a cost of many allocations, indirections, complicated memory management, etc. The lazy API can be built on top of the eager one, stacking in the opposite direction means we always have to pay the cost.

@yschimke ?

src/automata/FireAndForgetRequester.h
yschimke: We could define that this call is synchronous and returns successfully once the data is written to the network. But having the API be async is closer to the pattern of the others.

src/ReactiveSocket.cpp
yschimke: I would be inclined to keep the callback (Subscriber or otherwise), precisely because you often don't care about the result.  With a callback the method could be setup to return immediately and invoke the callback, client can choose to ignore it.  But without the callback we have to wait for it to actually succeed or fail to be written.

stupaq-fb: You won't know whether it was written anywhere (and why would it be helpful?), as everything on the way, including DuplexConnection, TCP, everything else may be "async". What is the information that this callback delivers? What can I do with that fact it was called?

src/ReactiveSocket.cpp
alexstaytuned: I'm going to get rid of it as the Subscriber here would be very superficial. But let me know if you think it is needed.

test/tcp/TcpServer.cpp
```diff
@@ -62,6 +62,11 @@ class ServerRequestHandler : public RequestHandler {
     response.onNext(folly::IOBuf::copyBuffer("from server2"));
     response.onComplete();
   }
+
+  void handleFireAndForgetRequest(Payload request) override {
+    std::cout << "ServerRequestHandler.handleFireAndForgetRequest "
```
stupaq-fb: This is Yuri's stuff, but I'd prefer to use logger for everything. AFAIK, logger is not synchronised with stdout, so this might result in misleading output.

stupaq-fb:
LGTM, I'll merge it after you consider the two nits.

test/tcp/TcpServer.cpp
alexstaytuned: I'm going to change the std::cout usages in the handler for now. There are other places in this class that are using cout. I'll talk to Yuri about whether there's a specific reason for that.

yschimke: No good reason here.  This is short lived test code.

However if you compare to another client we expect users to use, it strictly uses STDOUT, STDERR such that unix pipes work. 

https://github.com/ReactiveSocket/reactivesocket-cli/blob/master/src/main/java/io/reactivesocket/cli/ConsoleOutputHandler.java

PR #69, opened 2016-07-20
Tck parser

lehecka:
This is just the parser utility classes for the tck driver


lehecka:
should we pull gflags as submodule? It seems that travis has issues compiling the sources because it is using old version of gflags: https://github.com/BVLC/caffe/issues/2597

tck-test/client.cpp
```diff
@@ -0,0 +1,35 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#include <gflags/gflags.h>
+#include <glog/logging.h>
+#include "TestSuite.h"
+#include "TestFileParser.h"
+#include "TestInterpreter.h"
+
+DEFINE_string(test_server, "localhost:9898", "ip:port of the test server to runt the client against");
```
yschimke: typo: runt

PR #70, opened 2016-07-22
Basic Stats - reactivesocks, connections/type, frames(read|written), bytes(read|written) 

yschimke:
Will complete more event types before submitting, but want to agree the approach

cc @lehecka @stupaq-fb @jspahrsummers

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,7 +117,9 @@ add_executable(
         src/ReactiveSocket.cpp
         src/ReactiveSocket.h
         test/simple/NullSubscription.cpp
-        test/simple/NullSubscription.h test/simple/StatsPrinter.cpp test/simple/StatsPrinter.h)
+        test/simple/NullSubscription.h
+        test/simple/StatsPrinter.cpp
+        test/simple/StatsPrinter.h)
 
 target_link_libraries(
         tcpclient
@@ -138,7 +140,9 @@ add_executable(
         test/simple/NullSubscription.cpp
         test/simple/NullSubscription.h
         test/simple/CancelSubscriber.cpp
-        test/simple/CancelSubscriber.h test/simple/StatsPrinter.cpp test/simple/StatsPrinter.h)
+        test/simple/CancelSubscriber.h
+        test/simple/StatsPrinter.cpp
+        test/simple/StatsPrinter.h)
 
 target_link_libraries(
         tcpserver
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,7 +64,9 @@ add_library(
   src/ReactiveStreamsCompat.h
   src/RequestHandler.h
   src/tcp/TcpDuplexConnection.cpp
-  src/tcp/TcpDuplexConnection.h)
+  src/tcp/TcpDuplexConnection.h 
+  src/Stats.h 
+  src/Stats.cpp)
 
 target_link_libraries(
   ReactiveSocket
@@ -92,7 +94,8 @@ add_executable(
   test/MockRequestHandler.h
   test/ReactiveSocketTest.cpp
   test/ReactiveStreamsMocksCompat.h
-  test/Test.cpp)
+  test/Test.cpp
+  test/MockStats.h)
 
 target_link_libraries(
   tests
@@ -114,7 +117,7 @@ add_executable(
         src/ReactiveSocket.cpp
         src/ReactiveSocket.h
         test/simple/NullSubscription.cpp
-        test/simple/NullSubscription.h)
+        test/simple/NullSubscription.h test/simple/StatsPrinter.cpp test/simple/StatsPrinter.h)
 
 target_link_libraries(
         tcpclient
@@ -135,7 +138,7 @@ add_executable(
         test/simple/NullSubscription.cpp
         test/simple/NullSubscription.h
         test/simple/CancelSubscriber.cpp
-        test/simple/CancelSubscriber.h)
+        test/simple/CancelSubscriber.h test/simple/StatsPrinter.cpp test/simple/StatsPrinter.h)
 
 target_link_libraries(
         tcpserver
--- a/src/Stats.h
+++ b/src/Stats.h
@@ -1,0 +1,27 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#pragma once
+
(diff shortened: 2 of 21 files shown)


src/Stats.h
```diff
@@ -0,0 +1,27 @@
...
+  virtual void socketCreated() = 0;
+  virtual void socketClosed() = 0;
+  virtual ~Stats() = default;
+};
+
+class NoopStats : public Stats {
+ public:
+  virtual ~NoopStats() = default;
+
+  void socketCreated(){};
```
jspahrsummers: Can you add the `override` keyword to these methods?

CMakeLists.txt
```diff
@@ -64,7 +64,7 @@ add_library(
   src/ReactiveStreamsCompat.h
   src/RequestHandler.h
   src/tcp/TcpDuplexConnection.cpp
-  src/tcp/TcpDuplexConnection.h)
+  src/tcp/TcpDuplexConnection.h src/Stats.h src/Stats.cpp)
```
stupaq-fb: Can we format them one per line? Diff view is less confusing.

src/Stats.h
```diff
@@ -0,0 +1,27 @@
...
+  static NoopStats noop;
+
+  virtual void socketCreated() = 0;
+  virtual void socketClosed() = 0;
+  virtual ~Stats() = default;
+};
+
+class NoopStats : public Stats {
+ public:
+  virtual ~NoopStats() = default;
```
stupaq-fb: Drop?

src/Stats.h
```diff
@@ -0,0 +1,27 @@
...
+#pragma once
+
+namespace reactivesocket {
+// forward decl for NoopStats
+class NoopStats;
+
+// Virtual interface for a stats receiver
+class Stats {
+ public:
+  static NoopStats noop;
```
stupaq-fb: We might regret this. Better to have a static in a method and return reference to it (so the static gets initialised lazily on the first invocation). Or a proper singleton.

test/ReactiveSocketTest.cpp
```diff
@@ -248,20 +249,27 @@ TEST(ReactiveSocketTest, Destructor) {
   auto serverConn = folly::make_unique<InlineConnection>();
   clientConn->connectTo(*serverConn);
 
+  // TODO: since we don't assert anything, should we just use the StatsPrinter
+  // instead?
```
stupaq-fb: Our tests should be silent if they pass, so I would say no.

test/ReactiveSocketTest.cpp
```diff
@@ -311,4 +319,9 @@ TEST(ReactiveSocketTest, Destructor) {
   for (size_t i = 0; i < 2; ++i) {
     clientSock->requestSubscription(originalPayload->clone(), clientInputs[i]);
   }
+
+  //  EXPECT_CALL(clientStats, socketClosed_());
```
stupaq-fb: Why is it commented out?

test/ReactiveSocketTest.cpp
stupaq-fb: Oh, sorry, glog is fine. As long as we don't use cout/cerr directly. But I would actually add assertions for those stats.

test/simple/StatsPrinter.h
```diff
@@ -0,0 +1,17 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#pragma once
+
+#include <src/Stats.h>
+
+namespace reactivesocket {
+// Virtual interface for a stats receiver
```
stupaq-fb: Doesn't seem to be true in this context.

test/tcp/TcpClient.cpp
```diff
@@ -68,6 +69,9 @@ class Callback : public AsyncSocket::ConnectCallback {
 }
 
 int main(int argc, char* argv[]) {
+  FLAGS_logtostderr = true;
```
stupaq-fb: :like:

src/Stats.h
yschimke: I'll try your first suggestion, I wasn't sure about singleton as it seemed to imply another folly dep.

test/ReactiveSocketTest.cpp
yschimke: I couldn't get the mocks behaving nicely.  It's called on destructor but it seems like the mocks run assertions on shutdown, so I was getting errors both ways. :(

src/Stats.h
stupaq-fb: Yep, static in a function should be fine. Although I find it hard to imagine a usage of this static that'd lead to SOIF...

test/ReactiveSocketTest.cpp
stupaq-fb: I don't understand, can you paste gtest failure?

src/Stats.h
yschimke: What does that look like?

src/Stats.h
```diff
@@ -0,0 +1,23 @@
...
+
+#pragma once
+
+namespace reactivesocket {
+// forward decl for NoopStats
+class NoopStats;
+
+class Stats {
+ public:
+  static NoopStats noop;
```
stupaq-fb: That is still a static, isn't it?

yschimke: The diff changes dropped the comments, but my last comment was "What does that look like?"

stupaq-fb: Foo foo() {
  static Foo bar;
  return bar;
}

The difference is that bar is constructed at run time, the first time any code path traverses the line with `static ...`. Otherwise we risk https://isocpp.org/wiki/faq/ctors#static-init-order.

lehecka: Foo**&** foo() {
...
}

test/simple/StatsPrinter.cpp
```diff
@@ -0,0 +1,44 @@
...
+void StatsPrinter::socketCreated() {
+  LOG(INFO) << "socketCreated";
+}
+
+void StatsPrinter::socketClosed() {
+  LOG(INFO) << "socketClosed";
+}
+
+void StatsPrinter::connectionCreated(
+    const char type[4],
```
yschimke: damn Intellij!

PR #71, opened 2016-07-22
make google mock a local file dependency. Make ReactiveStreams use Ex…

tmontgomery:
…ternalProject with a settable URL. Add tests to ctest. Have travis use ctest.


yschimke:
How would this relate to my change to silence the test by implementing a test main?  I'll hold off on that commit until I am up to speed.  https://github.com/ReactiveSocket/reactivesocket-cpp/pull/72

external/googletest
```diff
@@ -1 +1 @@
-Subproject commit ec44c6c1675c25b9827aacd08c02433cccde7780
+Subproject commit d225acc90bc3a8c420a9bcd1f033033c1ccd7fe0
```
yschimke: this is older SHA, do we want to go back? Do we care?

tmontgomery:
@yschimke ctest can run with or without output from googletest. I'll look at the PR and update.

external/googletest
tmontgomery: By pulling googletest out into a local file, this only matters for non-CMake builds.

yschimke:
The output from googletest should still work, the only difference is hopefully that we also support gflags on the command line for the test runner.  Done by implementing our own test runner e.g. a main method that replaces gmock_main.

tmontgomery:
@yschimke PR updated. Now uses the standard test runner. Can use ctest as well.

tmontgomery:
any additional thoughts on this PR? Happy to discuss.

PR #80, opened 2016-08-02
add metadata push

YiqiuLiu:
1. Implemented Frame_METADATA_PUSH class definition in frame.h and class implementation in frame.cpp
2. Add onNextFrame(Frame_METADATA_PUSH&); in ConnectionAutomation.cpp
3. Add metadataPush method in reactiveSocket.cpp
4. Add virtual handler in RequestHandler.h
5. Add handlers in tests directory
6. Add a frame test for Frame_METADATA_PUSH in FrameTest.cpp
7. Add a metadata test in ReactiveSocketTest.cpp


src/Frame.h
```diff
@@ -259,6 +259,23 @@ class Frame_REQUEST_FNF {
...
 std::ostream& operator<<(std::ostream&, const Frame_REQUEST_FNF&);
 
+class Frame_METADATA_PUSH {
+ public:
+  static constexpr bool Trait_CarriesAllowance = false;
+
+  Frame_METADATA_PUSH() {}
+  Frame_METADATA_PUSH(FrameMetadata metadata)
+      : header_(FrameType::METADATA_PUSH, FrameFlags_METADATA, 0),
+      metadata_(std::move(metadata)) {}
```
lehecka: you should call metadata_.checkFlags(flags); inside of the ctor. This validation will make sure that the FrameMetadata is not empty.

src/ReactiveSocket.cpp
```diff
@@ -90,6 +90,12 @@ void ReactiveSocket::requestFireAndForget(Payload request) {
   connection_->onNextFrame(frame);
 }
 
+void ReactiveSocket::metadataPush(Payload metadata) {
+  FrameMetadata metadata_ = FrameMetadata(std::move(metadata));
```
lehecka: local variables should not have the trailing _. This is for class member variables.

src/ConnectionAutomaton.cpp
```diff
@@ -182,6 +183,16 @@ void ConnectionAutomaton::onConnectionFrame(Payload payload) {
       // TODO handle lease logic
       LOG(INFO) << "ignoring setup frame";
       return;
+    case FrameType::METADATA_PUSH: {
+      Frame_METADATA_PUSH frame;
+      if (frame.deserializeFrom(std::move(payload))) {
+        assert(frame.header_.flags_ & FrameFlags_METADATA);
```
yschimke: This logic with flipping the flag and responding is particular to keepalive

> Reception of a KEEPALIVE frame with the (R)espond flag set MUST cause a client or server to send back a KEEPALIVE with the (R)espond flag NOT set. The data in the received KEEPALIVE MUST be echoed back in the generated KEEPALIVE.

For metadata push, we need to allow someone implementing RequestHandler to receive the metadata payload.  e.g. compare 

https://github.com/ReactiveSocket/reactivesocket-java/blob/8408978e638c7e42948017d0e74a37ce115c2aec/reactivesocket-core/src/main/java/io/reactivesocket/RequestHandler.java#L54

and

https://github.com/ReactiveSocket/reactivesocket-cpp/blob/2be2f4f5e74c91506f6be680e57290bfbb214dd4/src/RequestHandler.h

src/RequestHandler.h
```diff
@@ -26,5 +26,8 @@ class RequestHandler {
 
   /// Handles a new fire-and-forget request sent by the other end.
   virtual void handleFireAndForgetRequest(Payload request) = 0;
+
+  /// Handles a new metadata-push sent bu the other end.
```
yschimke: typo: bu

yschimke:
I added support to the reactivesocket-cli project which builds on the java impl.  To test with this client you can do the following

```
$ git clone https://github.com/ReactiveSocket/reactivesocket-cli
$ cd reactivesocket-cli
$ gradle installDist
$ ./build/install/reactivesocket-cli/bin/reactivesocket-cli --metadata -i Hello tcp://localhost:9898
```

src/ReactiveSocket.cpp
```diff
@@ -151,6 +157,14 @@ bool ReactiveSocket::createResponder(
       handler_->handleFireAndForgetRequest(std::move(frame.data_));
       break;
     }
+    case FrameType::METADATA_PUSH: {
```
yschimke: Since metadata push is always on the connection, I don't think you need this here and it should replace the logic you have in ConnectionAutomaton.

yschimke:
This change is looking good, I'm keen to see this land.

src/ReactiveSocket.cpp
YiqiuLiu: If I move this part to ConnectionAutomation, then how can I get handler in that file?

src/Frame.h
YiqiuLiu: What should I put as default flag in ReactiveSocket::metadataPush? FrameFlags_EMPTY or FrameFlags_METADATA?

yschimke: For the first constructor it could be EMPTY, but it will be overwritten later anyway.  For the second, it should always be METADATA as per https://github.com/ReactiveSocket/reactivesocket/blob/master/Protocol.md

TIOLI I would consider changing this constructor to take a Payload and making FrameMetadata more of an implementation detail.

src/ReactiveSocket.cpp
yschimke: Simplest way is in ReactiveSocket::ReactiveSocket, passing it into the ConnectionAutomation constructor. I don't think you have a choice since ConnectionAutomaton::onNext won't call this block for streamId 0.

yschimke:
ping: Just want to confirm you are not waiting on something from me on this PR.

YiqiuLiu:
Hi Yuri, please ignore this update in this PR, I accidentally pushed to the branch.
By the way, I have been in stuck for a few days. Still does not know how to solve this problem.  From your last advice, if I want to have the handler in ConnectionAutomation, I should pass it to the ConnectionAutomation, does that mean to modify the ctor of ConnectionAutomation? I am really confused of what to do, can you give me some directions?

yschimke:
Yes, you would probably need to change the constructor.

> Simplest way is in ReactiveSocket::ReactiveSocket, passing it into the ConnectionAutomation constructor. I don't think you have a choice since ConnectionAutomaton::onNext won't call this block for streamId 0.

YiqiuLiu:
Since handler is a unique pointer in Reactive Socket but if we want to have that in ConnectionAutomation, it will cause a problem. Should we make it shared pointer or try other way to avoid this?

yschimke:
Did something happen with the formatting of some of the files?  You should be able to run devtools/format_all.sh to get back to the correct formatting.

yschimke:
TIOLI: to unblock you while waiting for feedback from others, if a simple reference works for now, that probably minimises the changes while getting the code working.  just put a TODO to fix before we merge this.

YiqiuLiu:
Formatting may be because I did some editing in CLion without setting the format in CLion. I will fix that.

yschimke:
@lehecka thoughts?

lehecka:
I think the problem with the pointer to the RequestHandler can be solved in a different way.
The ConnectionAutomaton has already a pointer (factory_) to the method from ReactiveSocket class. I would rename the variable factory_ to reactiveSocketHandler_ and call into it. The handling of METADATA_PUSH_Frame can be done inside ReactiveSocket::createResponder method which I would rename to respondToInteraction.

src/ReactiveSocket.h
```diff
@@ -86,6 +88,7 @@ class ReactiveSocket {
   bool createResponder(StreamId streamId, Payload& frame);
 
   const std::shared_ptr<ConnectionAutomaton> connection_;
+  // std::unique_ptr<RequestHandler> handler_;
```
yschimke: remove

yschimke:
I think one commented line and a typo and it's good to go.  @lehecka can merge when he is happy, otherwise I will tomorrow.

lehecka:
LGTM. Ready to merge when you address the last comment from Yuri.

src/ConnectionAutomaton.h
```diff
@@ -161,5 +162,6 @@ class ConnectionAutomaton :
   bool isServer_;
   std::unique_ptr<KeepaliveTimer> keepaliveTimer_;
   std::vector<ConnectionCloseListener> closeListeners_;
+  std::unique_ptr<RequestHandler> handler_;
```
lehecka: this is unused, as far as I can see

src/ConnectionAutomaton.h
```diff
@@ -11,6 +11,7 @@
 #include "ReactiveSocket.h"
 #include "src/Payload.h"
 #include "src/ReactiveStreamsCompat.h"
+#include "src/RequestHandler.h"
```
lehecka: this is unnecessary..

src/ConnectionAutomaton.h
YiqiuLiu: Sorry, I should have deleted this.

lehecka:
Good Yiqiu! Thanks!

PR #83, opened 2016-08-03
Add initial support for REQUEST_STREAM

jrahman:
General changelog:
1. Added Frame_REQUEST_STREAM class and implementation
2. Added unit test for Frame_REQUEST_STREAM
3. Updated several interfaces with the new requestStream() and handleRequestStream() API
   a. This included MixinTerminator, MockRequestHandler, and several other classes
4. Fixed two build errors
    a. Missing folly/Conv.h header
    b. Incorrect library name in CMakeList.txt
5. Added implementations of StreamRequester and StreamResponder automata
6. Wired both automata into ReactiveSocket
7. Implemented unit test for two different stream interaction patterns
     a. Run until server sends completion
     b. Run until client cancels
8. Refactored Stream and Subscription Request/Response automata to share base class

As mentioned in the changelog 3 new unit tests were added:
1 for Frame_REQUEST_STREAM and 2 for the RequestStream automata
All compile and pass successfully


src/automata/StreamRequester.cpp
```diff
@@ -0,0 +1,141 @@
...
+    case State::CLOSED:
+      break;
+  }
+}
+
+void StreamRequesterBase::endStream(StreamCompletionSignal signal) {
+  switch (state_) {
+    case State::NEW:
+    case State::REQUESTED:
+      // TODO(jprahman)
```
lehecka: not sure what do you want to add here...

src/automata/StreamRequester.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+void StreamRequesterBase::onNextFrame(Frame_ERROR& frame) {
+  switch (state_) {
+    case State::NEW:
+      // Cannot receive a frame before sending the initial request.
+      CHECK(false);
+      break;
+    case State::REQUESTED:
+      state_ = State::CLOSED;
+      connection_->endStream(streamId_, StreamCompletionSignal::GRACEFUL);
```
lehecka: instead of this line, you want:
Base::onError(
          std::runtime_error(frame.data_->moveToFbString().toStdString()));
      connection_->endStream(streamId_, StreamCompletionSignal::ERROR);

(picking up on the latest changes)

lehecka:
There is a lot of copy/paste code from SubscriptionRequester/Responder. Since the 2 interactions (requestSubscription and requestStream) are almost identical, then we should not duplicate the code but rather share the code. I suggest that SubscriptionRequester/Responder inherits from StreamRequester/Responder and have overrides where necessary or we create a common base class with the common code. I would prefer the first approach.

Also the tests are failing because of memory leaking.

jrahman:
Yeah, I've noticed that both classes are practically identical. Unless there's another reason I'm not aware of, I'd actually prefer making both inherit from a common base class, so that one isn't 'subordinate' to the other and they can evolve independently from one another if both require further specialization. Otherwise if changes occur in the parent class that you don't want in the child, then you end up being forced to override that in the child class, and if you forget to do so, thats an easy bug.

jrahman:
I'll also take a look at the memory leak. Not sure why the new tests are leaking, but the others are not.

jrahman:
Fixed the memory leak in the unit test, turns out I wasn't completing the shutdown handshake in the test as required.

I went ahead and used a shared base class approach because the *Requester::onNext() must instantiate the correct type of Frame (either Frame_REQUEST_SUB or Frame_REQUEST_STREAM). A common base class allows templating the base class with the type of frame to instantiate inside ::onNext() so that each *Requester class does not need to reimplement all of ::onNext()

CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ include_directories(SYSTEM ${OPENSSL_INCLUDE_DIR})
 
 include_directories(${CMAKE_SOURCE_DIR})
 
-include_directories(${CMAKE_CURRENT_BINARY_DIR}/reactivesocket/include)
+include_directories(${CMAKE_CURRENT_BINARY_DIR}/reactivestreams/include)
```
yschimke: what is the reason for this change?

yschimke:
LGTM

yschimke:
FYI tested fine with the java client, after implementing a response in tcp server.

```
$ brew install yschimke/tap/reactivesocket-cli
$ reactivesocket-cli --str -i Hello tcp://localhost:9898
from server
from server2
```

lehecka:
I am not a big fan of StreamSubscriptionRequesterBase being a templated class. There is only 1 spot where the type T is used. 
I suggest creating a pure virtual method in the base class (something like sendRequestFrame) and implementing it in StreamRequesterBase and SubscriptionRequesterBase.
This way we reduce the generated code size and get rid of templates

yschimke:
Do you want me to unmerge, or do a follow up change?

lehecka:
Follow up change would be fine. Less hassle.

jrahman:
Follow-up change in PR #88

PR #87, opened 2016-08-04
[WIP] keep alives and remove as many aborts as possible

yschimke:
I will break up this into 4 PRs e.g.

1. Frame changes (backwards compatible)
2. remove aborts
3. keepalive sending
4. on close support

But any initial feedback welcome @lehecka


CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ find_package(Folly REQUIRED)
 find_package(OpenSSL REQUIRED)
 find_package(Boost REQUIRED COMPONENTS system thread)
 include_directories(SYSTEM ${FOLLY_INCLUDE_DIR})
-include_directories(SYSTEM ${OPENSSL_INCLUDE_DIR})
+include_directories(SYSTEM /usr/local/Cellar/openssl/1.0.2h_1/include)
```
yschimke: will remove, can activate on the command line

src/ConnectionAutomaton.h
```diff
@@ -22,13 +24,18 @@ using StreamId = uint32_t;
...
 ///
-/// It is a responsibility of this startegy to register the responder with the
+/// It is a responsibility of this strategy to register the responder with the
 /// connection automaton and provide it with the initial frame if needed.
 /// Returns true if the responder has been created successfully, false if the
 /// frame cannot start a new stream, in which case the frame (passed by a
 /// mutable referece) must not be modified.
 using StreamAutomatonFactory = std::function<bool(StreamId, Payload&)>;
 
+class ConnectionCloseCallback {
```
lehecka: do you think we can replace this with std::function<void()>?

src/ConnectionAutomaton.cpp
```diff
@@ -172,22 +186,64 @@ void ConnectionAutomaton::onConnectionFrame(Payload payload) {
...
+        // TODO(yschimke) should this be onTerminal
+        cancel();
+      }
+    }
       return;
     default:
-      // TODO(yschimke): check ignore flag and fail
-      assert(false);
+      // TODO(yschimke) make this conditional if we have versioning problems
+      Frame_ERROR errorFrame(
```
lehecka: this is a second time when I see this error frame with the same message.
Should we create static methods for such instances?
Eq.
Frame_ERROR::unexpectedFrame() which would return a new instance of Frame_ERROR?
similarly
Frame_ERROR::badSetupFrame(), 
etc...

For improved readability..

src/ConnectionAutomaton.cpp
```diff
@@ -241,15 +297,44 @@ void ConnectionAutomaton::cancel() {
 }
 /// @}
 
+void ConnectionAutomaton::scheduleKeepalive(const uint32_t keepAliveDelay) {
+  if (keepAliveDelay > 0) {
+    auto eventBase = folly::EventBaseManager::get()->getExistingEventBase();
```
lehecka: A question for discussion...
Grabbing current EventBase instance will work for all our platforms, but in general, should we wrap EventBase+runAfterDelay into a new interface called KeepAliveScheduler and initialize ConnectionAutomaton with an instance to the concrete scheduler? That way we can remove the dependency to EventBase from the core of ReactiveSocket..
What do you think?

src/ConnectionAutomaton.cpp
```diff
@@ -241,15 +297,44 @@ void ConnectionAutomaton::cancel() {
...
+    CHECK(eventBase);
+
+    eventBase->runAfterDelay(
+        [this, keepAliveDelay]() { sendKeepalive(keepAliveDelay); },
+        keepAliveDelay);
+  }
+}
+
+void ConnectionAutomaton::sendKeepalive(const uint32_t keepAliveDelay) {
+  // TODO is this check safe? or needs to check a synchronized shared flag?
```
lehecka: since we are assuming everything runs on the same EventBase, I believe you are safe not to synchronize.

src/ConnectionAutomaton.cpp
```diff
@@ -241,15 +297,44 @@ void ConnectionAutomaton::cancel() {
 }
 /// @}
 
+void ConnectionAutomaton::scheduleKeepalive(const uint32_t keepAliveDelay) {
+  if (keepAliveDelay > 0) {
+    auto eventBase = folly::EventBaseManager::get()->getExistingEventBase();
+    CHECK(eventBase);
+
+    eventBase->runAfterDelay(
+        [this, keepAliveDelay]() { sendKeepalive(keepAliveDelay); },
```
lehecka: is it possible that this instance will be destroyed when the eventBase will execute the callback?
I believe that its possible. In that case we need to capture shared_ptr<ConnectionAutomaton> thisPtr.

src/ReactiveSocket.cpp
```diff
@@ -25,19 +25,29 @@
...
+  //  connection_->onClose(std::move(closeCallback));
+}
+
 std::unique_ptr<ReactiveSocket> ReactiveSocket::fromClientConnection(
     std::unique_ptr<DuplexConnection> connection,
     std::unique_ptr<RequestHandler> handler,
-    Stats& stats) {
+    Stats& stats,
+    uint32_t keepAliveDelay) {
   std::unique_ptr<ReactiveSocket> socket(new ReactiveSocket(
```
lehecka: can we replace calling "new ReactiveSocket" with std::make_unique ?
I like to keep following the guideline that if there is "new", there should be "delete". If I don't have new, then I don't have to search for "delete". Functionally, I know it means the same.

src/ConnectionAutomaton.h
yschimke: I was using std:function originally, switched to this because Justin explained its cheaper.  But maybe I misunderstood something.

src/ReactiveSocket.cpp
yschimke: I thought that was a folly thing until C++14

yschimke:
replaced by individual stacked PRs

PR #88, opened 2016-08-04
Refactor StreamSubscriberRequesterBase to remove template

jrahman:
Refactor StreamSubscriberRequesterBase to not be a template (reduced code footprint), and instead offer a pure virtual method for onNext() for subclasses to override in order to send the correct frame type.

### DIFF
--- a/src/automata/StreamRequester.cpp
+++ b/src/automata/StreamRequester.cpp
@@ -7,6 +7,42 @@
 
 namespace reactivesocket {
 
+void StreamRequesterBase::onNext(Payload request) {
+  switch (state_) {
+    case State::NEW: {
+      state_ = State::REQUESTED;
+      // FIXME: find a root cause of this assymetry; the problem here is that
+      // the Base::request might be delivered after the whole thing is shut
+      // down, if one uses InlineConnection.
+      size_t initialN = initialResponseAllowance_.drainWithLimit(
+          Frame_REQUEST_N::kMaxRequestN);
+      size_t remainingN = initialResponseAllowance_.drain();
+      // Send as much as possible with the initial request.
+      CHECK_GE(Frame_REQUEST_N::kMaxRequestN, initialN);
+      auto flags = initialN > 0 ? FrameFlags_REQN_PRESENT : FrameFlags_EMPTY;
+      Frame_REQUEST_STREAM frame(
+          streamId_,
+          flags,
+          static_cast<uint32_t>(initialN),
+          FrameMetadata::empty(),
+          std::move(request));
+      // We must inform ConsumerMixin about an implicit allowance we have
+      // requested from the remote end.
+      addImplicitAllowance(initialN);
+      connection_->onNextFrame(frame);
+      // Pump the remaining allowance into the ConsumerMixin _after_ sending the
+      // initial request.
+      if (remainingN) {
+        Base::request(remainingN);
+      }
+    } break;
+    case State::REQUESTED:
+      break;
+    case State::CLOSED:
+      break;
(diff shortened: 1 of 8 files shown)


stupaq-fb:
> reduced code footprint

How did you measure that?

jrahman:
It's not measured directly, but the base class template leads to object code bloat. Because each subclass (StreamRequester/SubscriptionRequester) used StreamSubscriptionRequesterBase with different template parameters, the compiler will generate two copies of each method in StreamSubscriptionRequesterBase because each templated version of the parent class is considered a different type. @lehecka requested the change for that reason.

I not certain if the compiler is smart enough in this case to realize the object code is the same for (most) methods in each templated version of the base class to perform deduplication.

Moving away from templates, the compiler will not duplicate object code for each method in the base class (StreamSubscriptionRequesterBase) because the base class is actually the same type in both subclasses.

src/automata/StreamRequester.cpp
```diff
@@ -7,6 +7,42 @@
...
+      // FIXME: find a root cause of this assymetry; the problem here is that
+      // the Base::request might be delivered after the whole thing is shut
+      // down, if one uses InlineConnection.
+      size_t initialN = initialResponseAllowance_.drainWithLimit(
+          Frame_REQUEST_N::kMaxRequestN);
+      size_t remainingN = initialResponseAllowance_.drain();
+      // Send as much as possible with the initial request.
+      CHECK_GE(Frame_REQUEST_N::kMaxRequestN, initialN);
+      auto flags = initialN > 0 ? FrameFlags_REQN_PRESENT : FrameFlags_EMPTY;
+      Frame_REQUEST_STREAM frame(
```
lehecka: I think this change is a great improvement. I would suggest taking one step further in this direction.
Instead of virtual onNext method which is identical for both StreamRequesterBase and SubscriptionRequesterBase, except for the frame type
   Frame_REQUEST_STREAM frame(
vs
   Frame_REQUEST_SUB frame(

you can create a new virtual method called sendRequestFrame which will be called from the onNext method and the different implementations will be:
SubscriptionRequesterBase::sendRequestFrame(...) {
       Frame_REQUEST_SUB frame(
           streamId_,
           flags,
           static_cast<uint32_t>(initialN),
           FrameMetadata::empty(),
           std::move(request));
       connection_->onNextFrame(frame);
}

StreamRequesterBase::sendRequestFrame(...) {
       Frame_REQUEST_STREAM frame(
           streamId_,
           flags,
           static_cast<uint32_t>(initialN),
           FrameMetadata::empty(),
           std::move(request));
       connection_->onNextFrame(frame);
}

src/automata/SubscriptionRequester.cpp
```diff
@@ -7,6 +7,22 @@
...
+    Payload&& request) {
+  Frame_REQUEST_SUB frame(
+      streamId_,
+      flags,
+      static_cast<uint32_t>(initialN),
+      FrameMetadata::empty(),
+      std::move(request));
+  // We must inform ConsumerMixin about an implicit allowance we have
+  // requested from the remote end.
+  addImplicitAllowance(initialN);
```
lehecka: One last nit: do you want to put this line to the onNext method, right before you call sendRequestFrame?
I don't see why moving this line up one spot would be a problem.

jrahman: Updated newest diff to move addImplicitAllowance() to StreamSubscriptionRequesterBase::onNext()

lehecka:
great work! Thank you Jason.

PR #92, opened 2016-08-05
c++14

yschimke:
discuss!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,7 +55,7 @@ ExternalProject_Add(
 find_package(Threads)
 
 # Common configuration for all build modes.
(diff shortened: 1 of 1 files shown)


yschimke:
Travis doesn't seem to support this without additional steps, so I'll look into that unless someone else knows the story?

yschimke:
or alternatively that we don't want this

lehecka:
please check if fbandroid and fbobjc support c++14 before you turn this switch on. Last time I checked (4 months back) fb android didn't support it. It would be extra work to deal with the compile time issues for all new c++14 stuff.

yschimke:
@lehecka stop suggesting std:make_unique then!

lehecka:
sorry, I might have forgotten...
there is also folly::make_unique doing the same thing in C++11.

Maybe I'll just shutup..

PR #93, opened 2016-08-05
close support on ReactiveSocket

yschimke:
just look at last diff

### DIFF
--- a/src/ReactiveSocket.h
+++ b/src/ReactiveSocket.h
@@ -16,15 +16,20 @@ namespace reactivesocket {
 class ConnectionAutomaton;
 class DuplexConnection;
 class RequestHandler;
+class ReactiveSocket;
 enum class FrameType : uint16_t;
 using StreamId = uint32_t;
 
 class KeepaliveTimer {
  public:
+  virtual ~KeepaliveTimer() = default;
+
   virtual std::chrono::milliseconds keepaliveTime() = 0;
   virtual void stop() = 0;
   virtual void start(ConnectionAutomaton* automaton) = 0;
 };
+
+using CloseListener = std::function<void(ReactiveSocket&)>;
 
 // TODO(stupaq): consider using error codes in place of folly::exception_wrapper
 
@@ -65,6 +70,10 @@ class ReactiveSocket {
   void requestSubscription(Payload payload, Subscriber<Payload>& responseSink);
 
   void requestFireAndForget(Payload request);
+
+  void close();
+
+  void onClose(CloseListener listener);
 
  private:
   ReactiveSocket(
--- a/src/folly/FollyKeepaliveTimer.h
+++ b/src/folly/FollyKeepaliveTimer.h
@@ -1,0 +1,32 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#pragma once
+
+#include <folly/io/async/EventBase.h>
+#include <src/ConnectionAutomaton.h>
+#include <src/ReactiveSocket.h>
+
+namespace reactivesocket {
+class FollyKeepaliveTimer : public KeepaliveTimer {
+ public:
+  FollyKeepaliveTimer(
+      folly::EventBase& eventBase,
+      std::chrono::milliseconds period);
+
+  ~FollyKeepaliveTimer();
+
+  std::chrono::milliseconds keepaliveTime() override;
+
+  void schedule();
+
+  void stop() override;
+
+  void start(ConnectionAutomaton* automaton) override;
+
+ private:
+  ConnectionAutomaton* automaton_{nullptr};
+  folly::EventBase& eventBase_;
+  std::shared_ptr<bool> running_;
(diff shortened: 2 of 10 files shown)


src/ReactiveSocket.h
```diff
@@ -15,9 +16,19 @@ namespace reactivesocket {
 class ConnectionAutomaton;
 class DuplexConnection;
 class RequestHandler;
+class ReactiveSocket;
 enum class FrameType : uint16_t;
 using StreamId = uint32_t;
 
+class KeepaliveTimer {
```
lehecka: you need the virtual destructor here

src/ReactiveSocket.h
```diff
@@ -15,9 +16,19 @@ namespace reactivesocket {
...
+ public:
+  virtual std::chrono::milliseconds keepaliveTime() = 0;
+  virtual void stop() = 0;
+  virtual void start(ConnectionAutomaton* automaton) = 0;
+};
+
+using CloseListener = std::function<void(ReactiveSocket*)>;
```
lehecka: lets have the param as ReactiveSocket&
this way the client code won't have to think about nullability of the param.

test/tcp/TcpClient.cpp
```diff
@@ -27,6 +28,46 @@ class Callback : public AsyncSocket::ConnectCallback {
     std::cout << "TODO error" << ex.what() << " " << ex.getType() << "\n";
   }
 };
+
+class FollyKeepaliveTimer : public KeepaliveTimer {
```
lehecka: do you think we should put the FollyKeepaliveTimer under src/<something>? it will be commonly used..

test/tcp/TcpClient.cpp
```diff
@@ -27,6 +28,46 @@ class Callback : public AsyncSocket::ConnectCallback {
...
+        [this, running]() {
+          if (*running) {
+            automaton_->sendKeepalive();
+            schedule();
+          }
+        },
+        keepaliveTime().count());
+  }
+
+  void stop() override {
```
lehecka: should we cal stop from the destructor too?

test/tcp/TcpClient.cpp
```diff
@@ -27,6 +28,46 @@ class Callback : public AsyncSocket::ConnectCallback {
...
+
+class FollyKeepaliveTimer : public KeepaliveTimer {
+ public:
+  explicit FollyKeepaliveTimer(folly::EventBase& eventBase)
+      : eventBase_(eventBase) {
+    running_ = std::make_shared<bool>(false);
+  }
+
+  std::chrono::milliseconds keepaliveTime() override {
+    return std::chrono::milliseconds(5000);
```
lehecka: let's parametrize this from the constructor with some default value

test/tcp/TcpClient.cpp
```diff
@@ -27,6 +28,46 @@ class Callback : public AsyncSocket::ConnectCallback {
...
+  };
+
+  void schedule() {
```
lehecka: can you add a comment somewhere that this method as well as close method are supposed to be executed on the same thread so we don't have race condition around the value of running?
Maybe a general comment for the class in the header file would be enough..

src/ReactiveSocket.h
yschimke: OK, that will still allow pointer equality, right?

test/tcp/TcpClient.cpp
yschimke: yep, will do.

test/tcp/TcpClient.cpp
yschimke: ok, will do.

test/tcp/TcpClient.cpp
yschimke: what about asserts that fail if you don't?

src/ReactiveSocket.h
lehecka: yep

test/tcp/TcpClient.cpp
lehecka: those asserts would involve remembering current thread at least.. I am not sure it that's what you want... 
Our implementation has such assumptions on a lot of places. I would be fine with just a comment, but am open to anything.

src/folly/FollyKeepaliveTimer.h
```diff
@@ -0,0 +1,26 @@
...
+ public:
+  FollyKeepaliveTimer(
+      folly::EventBase& eventBase,
+      std::chrono::milliseconds period);
+
+  std::chrono::milliseconds keepaliveTime() override;
+
+  void schedule();
+
+  void stop() override;
```
lehecka: Destructor should call stop too.

lehecka:
Please address the last comment and feel free to check it in. LGTM

yschimke:
@lehecka which comment? I thought I had.

PR #96, opened 2016-08-17
main components of the tck driver.

lehecka:
please feel free to comment.


tck-test/TestFileParser.cpp
```diff
@@ -44,15 +41,28 @@ void TestFileParser::parseCommand(const std::string& command) {
...
+    currentTest_.setSucceeding(parameters[0] == "pass");
+    return;
+  }
 
   TestCommand newCommand(std::move(parameters));
   if (!newCommand.valid()) {
-    LOG(ERROR) << "invalid command on line " << currentLine_ << ": " << command
-               << " (ignoring)";
+    LOG(ERROR) << "invalid command on line " << currentLine_ << ": " << command;
+    throw std::runtime_error("unknown commad in the test");
```
maxwellsayles: command

tck-test/TestFileParser.cpp
```diff
@@ -44,15 +41,28 @@ void TestFileParser::parseCommand(const std::string& command) {
...
   } else {
     currentTest_.addCommand(std::move(newCommand));
   }
 }
 
+void TestFileParser::addCurrentTest() {
+  if (currentTest_.empty()) {
```
maxwellsayles: nit: may as well reverse the truth since there's no test defined.

tck-test/TestFileParser.cpp
```diff
@@ -44,15 +41,28 @@ void TestFileParser::parseCommand(const std::string& command) {
...
+    // no test defined
+  } else {
+    testSuite_.addTest(std::move(currentTest_));
+    DCHECK(currentTest_.empty());
```
yschimke: Is there a reason this isn't just CHECK?

lehecka: the std::move(currentTest_) on the line above is supposed to steal the guts from that instance and leave it in the empty state. The DCHECK is just a sanity check that it actually does that. If it works in debug build, it will work always.

yschimke:
Passing in clang, failing with gcc

https://travis-ci.org/ReactiveSocket/reactivesocket-cpp/builds/152846839

PR #98, opened 2016-08-17
fix tests for setup frame

yschimke:


yschimke:
1 more test to fix :(

PR #99, opened 2016-08-18
fix the tests

yschimke:
ReactiveSocket tests expect setup frame, DuplexConnection not so much


yschimke:
@lehecka merging to fix the build, can you give me a retroactive shipit.

PR #100, opened 2016-08-19
avoid leaking Frame

yschimke:
Use std::string instead of FrameType in stats API

### DIFF
--- a/src/ConnectionAutomaton.cpp
+++ b/src/ConnectionAutomaton.cpp
@@ -9,6 +9,7 @@
 #include <folly/io/IOBuf.h>
 #include <glog/logging.h>
 #include <iostream>
+#include <sstream>
 
 #include "src/AbstractStreamAutomaton.h"
 #include "src/DuplexConnection.h"
@@ -122,7 +123,7 @@ void ConnectionAutomaton::onNextFrame(Payload frame) {
     return;
   }
   if (pendingWrites_.empty() && writeAllowance_.tryAcquire()) {
-    connectionOutput_.onNext(std::move(frame));
+    writeFrame(std::move(frame));
   } else {
     // We either have no allowance to perform the operation, or the queue has
     // not been drained (e.g. we're looping in ::request).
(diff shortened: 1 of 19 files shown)


src/ConnectionAutomaton.cpp
```diff
@@ -322,10 +330,19 @@ void ConnectionAutomaton::handleUnknownStream(
...
 
 void ConnectionAutomaton::onClose(ConnectionCloseListener listener) {
   closeListeners_.push_back(listener);
 }
+
+void ConnectionAutomaton::writeFrame(Payload frame) {
+  std::stringstream ss;
+  ss << FrameHeader::peekType(*frame);
+
+  stats_.frameWritten(ss.str());
```
lehecka: nit: I think you would be fine using folly::to<std::string>(...)

test/simple/StatsPrinter.cpp
```diff
@@ -34,11 +34,11 @@ void StatsPrinter::bytesRead(size_t bytes) {
   LOG(INFO) << "bytesRead " << bytes;
 }
 
-void StatsPrinter::frameWritten(FrameType type) {
+void StatsPrinter::frameWritten(const std::string& type) {
```
lehecka: nit: please consider param name "frameType"

src/ConnectionAutomaton.cpp
```diff
@@ -322,10 +330,19 @@ void ConnectionAutomaton::handleUnknownStream(
...
       FrameFlags_KEEPALIVE_RESPOND, folly::IOBuf::create(0));
-  connectionOutput_.onNext(pingFrame.serializeOut());
+  writeFrame(pingFrame.serializeOut());
 }
 
 void ConnectionAutomaton::onClose(ConnectionCloseListener listener) {
   closeListeners_.push_back(listener);
 }
+
+void ConnectionAutomaton::writeFrame(Payload frame) {
```
lehecka: nit: please consider naming this method outputFrame. It would be clearer to the me (reader) where is the frame going..

lehecka:
other than few minor comments, it LGTM

yschimke:
holding off on the folly change until we agree definitely on more folly deps

PR #101, opened 2016-08-19
Payload = (data, metadata)

lehecka:
The main change is the new class Payload which stores data + metadata.
We were using alias Payload = std::unique_ptr<IOBuf> before. This alias was used by the ReactiveSocket API as well as DuplexConnection API. I had to separate those two. 
For all (low level) places where Payload should stay std::unique_ptr<IOBuf> I  replaced the type with std::unique_ptr<IOBuf>.

I also did some cleanup mainly in Frame.cpp where there was FrameMetadata class which no longer is needed and I removed duplicated code.


CMakeLists.txt
```diff
@@ -119,6 +119,7 @@ add_library(
   src/mixins/StreamIfMixin.h
   src/NullRequestHandler.cpp
   src/NullRequestHandler.h
+  src/Payload.cpp
```
yschimke: Does this need a similar change to TARGETS?

src/Frame.cpp
```diff
@@ -594,17 +363,14 @@ Payload Frame_KEEPALIVE::serializeOut() {
   return queue.move();
 }
 
-bool Frame_KEEPALIVE::deserializeFrom(Payload in) {
+bool Frame_KEEPALIVE::deserializeFrom(std::unique_ptr<folly::IOBuf> in) {
   folly::io::Cursor cur(in.get());
-  if (!header_.deserializeFrom(cur)) {
-    return false;
-  }
   assert((header_.flags_ & FrameFlags_METADATA) == 0);
```
yschimke: this assert looks premature

src/Payload.cpp
```diff
@@ -0,0 +1,94 @@
...
+  metadata = deserializeMetadataFrom(cur, flags);
+  data = deserializeDataFrom(cur);
+}
+
+std::ostream& operator<<(std::ostream& os, const Payload& payload) {
+  return os << "[metadata: \""
+            << (payload.metadata
+                    ? folly::to<std::string>(
+                          payload.metadata->computeChainDataLength())
+                    : "<null>")
```
yschimke: I don't think these two <null> should be quoted.  otherwise you can't tell the difference with "<null>" e.g. a 6 char string and nullptr.

test/InlineConnectionTest.cpp
```diff
@@ -54,29 +53,33 @@ TEST(InlineConnectionTest, PingPong) {
...
+      .WillOnce(Invoke([&]() {
+        output[1]->onComplete(); // "Unsubscribe handshake".
+        inputSub[1]->cancel(); // Close the other direction.
+      }));
   EXPECT_CALL(input[0], onComplete_())
       .InSequence(s); // This finishes the handshake.
-  EXPECT_CALL(outputSub[0], cancel_()).InSequence(s).WillOnce(Invoke([&]() {
-    output[0]->onComplete(); // "Unsubscribe handshake".
-  }));
+  EXPECT_CALL(outputSub[0], cancel_())
```
yschimke: how are you formatting this? this seems incorrect.

src/mixins/LoggingMixin.h
```diff
@@ -65,8 +65,11 @@ class LoggingMixin : public Base {
   }
 
   void onNext(Payload payload) {
-    Base::logPrefix(LOG(INFO)) << "onNext(<"
-                               << payload->computeChainDataLength() << ">)";
+    Base::logPrefix(LOG(INFO))
+        << "onNext(<d:"
+        << (payload.data ? payload.data->computeChainDataLength() : 0) << ", m:"
```
yschimke: don't you already have a Payload tostring for this?

src/automata/StreamSubscriptionRequesterBase.cpp
```diff
@@ -113,8 +113,8 @@ void StreamSubscriptionRequesterBase::onNextFrame(Frame_ERROR& frame) {
       break;
     case State::REQUESTED:
       state_ = State::CLOSED;
-      Base::onError(
-          std::runtime_error(frame.data_->moveToFbString().toStdString()));
+      Base::onError(std::runtime_error(
+          frame.payload_.data->moveToFbString().toStdString()));
```
yschimke: Can you add a Payload::dataAsString?

yschimke:
overall the change looks really positive, few comments.  It's a tough review though.

test/InlineConnectionTest.cpp
lehecka: ./devtools/format_all.sh

src/automata/StreamSubscriptionRequesterBase.cpp
lehecka: will add moveDataToString. I want to make sure that the reader is aware of the destruction of the payload.data field.

lehecka:
thanks for the review. I know this is a large change.

yschimke:
Remove the .dsstore file

PR #102, opened 2016-08-22
Cleaning up frame ownership when passed around

lehecka:
We used to allocate frames and send them down as reference arguments to methods. Most of the methods were stealing (std::move) the guts from the instances. For this case we should send the frames down as r-value references or values to cleanly define the ownership.
I decided for the r-value references to avoid accidental copies.


yschimke:
LGTM

PR #105, opened 2016-08-23
NoopStats non copyable

yschimke:


src/Stats.cpp
```diff
@@ -19,10 +19,22 @@ class NoopStats : public Stats {
...
+    return singleton;
+  }
+
+ protected:
+  NoopStats() = default;
+  ~NoopStats() = default;
+
+ private:
+  NoopStats(const NoopStats& other) = delete; // non construction-copyable
+  NoopStats& operator=(const NoopStats&) = delete; // non copyable
```
lehecka: I would also make this type non movable

PR #106, opened 2016-08-23
ConnectionAutomaton: Output frame

lehecka:
ConnectionAutomaton was not respecting allowance for writing frames to DuplexConnection.
This change introduces only one public method for writing frames to output which will enqueue frames if we are not allowed to write to DuplexConnection.
Also in this change we will make sure that frames are written and output in order.


src/ConnectionAutomaton.cpp
```diff
@@ -55,7 +55,7 @@ void ConnectionAutomaton::connect() {
         "",
         "",
         Payload());
-    onNextFrame(frame.serializeOut());
+    outputFrameOrEnqueue(frame.serializeOut());
```
yschimke: This connect method sticks out, especially since we probably want it to take a SetupFrameData with mimetype and lease modes eventually.

PR #107, opened 2016-08-23
virtual inheritance for FramedDuplexConnection

lehecka:
To enable creating decorated version of FramedDuplexConnection.


yschimke:
Can you cleanup your commits, maybe trash master and get a clean copy from origin?

PR #113, opened 2016-08-25
initial resume machinery

tmontgomery:
Only for discussion (DO NOT MERGE). Incomplete. No tests written yet.

Some basic handling of RESUME and RESUME_OK for idle connections and tracking of positions.

### DIFF
--- a/src/ReactiveSocket.cpp
+++ b/src/ReactiveSocket.cpp
@@ -35,6 +35,7 @@ ReactiveSocket::ReactiveSocket(
     bool isServer,
     std::unique_ptr<DuplexConnection> connection,
     std::unique_ptr<RequestHandler> handler,
+    ResumeSocketListener resumeListener,
     Stats& stats,
     std::unique_ptr<KeepaliveTimer> keepaliveTimer)
     : connection_(new ConnectionAutomaton(
@@ -44,6 +45,9 @@ ReactiveSocket::ReactiveSocket(
               this,
               std::placeholders::_1,
               std::placeholders::_2),
+          [&](const ResumeIdentificationToken& token) {
+             return resumeListener(*this, token);
+          },
           stats,
           isServer)),
       handler_(std::move(handler)),
@@ -60,6 +64,7 @@ std::unique_ptr<ReactiveSocket> ReactiveSocket::fromClientConnection(
       false,
       std::move(connection),
       std::move(handler),
+      [](ReactiveSocket&, const ResumeIdentificationToken&) { return false; },
       stats,
       std::move(keepaliveTimer)));
   socket->connection_->connect();
@@ -90,11 +95,13 @@ std::unique_ptr<ReactiveSocket> ReactiveSocket::fromClientConnection(
 std::unique_ptr<ReactiveSocket> ReactiveSocket::fromServerConnection(
     std::unique_ptr<DuplexConnection> connection,
     std::unique_ptr<RequestHandler> handler,
-    Stats& stats) {
+    Stats& stats,
+    ResumeSocketListener resumeListener) {
   std::unique_ptr<ReactiveSocket> socket(new ReactiveSocket(
       true,
       std::move(connection),
       std::move(handler),
+      resumeListener,
       stats,
       std::unique_ptr<KeepaliveTimer>(nullptr)));
   socket->connection_->connect();
@@ -266,4 +273,9 @@ void ReactiveSocket::close() {
 void ReactiveSocket::onClose(CloseListener listener) {
   connection_->onClose([listener, this]() { listener(*this); });
 }
+
+void ReactiveSocket::resumeFromSocket(ReactiveSocket &socket) {
+  connection_->resumeFromAutomaton(*socket.connection_);
+}
+
 }
--- a/src/ReactiveSocket.cpp
+++ b/src/ReactiveSocket.cpp
@@ -278,4 +278,10 @@ void ReactiveSocket::resumeFromSocket(ReactiveSocket &socket) {
   connection_->resumeFromAutomaton(*socket.connection_);
 }
 
+void ReactiveSocket::tryClientResume(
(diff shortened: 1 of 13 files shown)


yschimke:
Look great so far.  I like the incremental approach and we might find the minimal idle only mode works well in practice.  e.g. for a mostly client/server subscription, if the Gateway knows the connection is down, just hold onto the last event rather than sending.

So I'd be interested to see what that looks like.  How does either side know the status of the ReactiveSocket is temporarily disconnected?

lehecka:
I like the change. So far its pretty straightforward and simple to follow.
I guess the next commit will add new methods to ReactiveSocket, like tryResume(std::unique_ptr<DuplexConnection>), etc.
Also we need to modify the behavior around when the existing DuplexConnection terminates, we should just keep going if resumable==true.

tmontgomery:
The key parts to look at next are how to keep the Resume state and Handlers around and look them up when doing a RESUME from the server perspective.

tmontgomery:
The server side logic for handling ReactiveSocket to ReactiveSocket based on token is in. Next is the client side tryResume() logic. Then I will add some tests and start hammering stuff out.

tmontgomery:
still work in progress. But resume TCP client and server are starting to send/receive RESUME and RESUME_OK

src/ReactiveSocket.cpp
```diff
@@ -259,11 +268,26 @@ bool ReactiveSocket::createResponder(
...
 
 void ReactiveSocket::onClose(CloseListener listener) {
   connection_->onClose([listener, this]() { listener(*this); });
 }
+
+void ReactiveSocket::resumeFromSocket(ReactiveSocket &socket) {
+  connection_->resumeFromAutomaton(*socket.connection_);
+}
+
+void ReactiveSocket::tryClientResume(
```
lehecka: since this is tryXXX method, how do we let the caller know whether this succeeded or failed?

src/ReactiveSocket.h
```diff
@@ -32,6 +32,8 @@ class KeepaliveTimer {
 
 using CloseListener = std::function<void(ReactiveSocket&)>;
 
+using ResumeSocketListener = std::function<bool(ReactiveSocket& newSocket, const ResumeIdentificationToken& token)>;
```
lehecka: Do we want to expose the ResumeIdentificationToken all the way to the client code or is it just an internal identifier of the socket?

src/ResumeCache.cpp
```diff
@@ -0,0 +1,12 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#include "src/ResumeCache.h"
+
+namespace reactivesocket {
+
+bool ResumeCache::retransmitFromPosition(
+    position_t initialPosition, ConnectionAutomaton &connection, const RetransmitFilter &filter) {
```
lehecka: How do you envision using the RetransmitFilter?

test/resume/TcpResumeServer.cpp
```diff
@@ -0,0 +1,185 @@
...
+        folly::make_unique<ServerRequestHandler>();
+
+    std::unique_ptr<ReactiveSocket> rs = ReactiveSocket::fromServerConnection(
+        std::move(framedConnection),
+        std::move(requestHandler),
+        stats_,
+        [&](ReactiveSocket& socket, const ResumeIdentificationToken& token)
+        {
+            std::cout << "resume callback" << std::endl;
+            socket.resumeFromSocket(*reactiveSockets_[0]);
```
lehecka: I didn't quite get why we are resuming a socket from a different socket...
it feels equivalent to me to reactiveSocket = std::move(reactiveSocket) which is not allowed..

src/ResumeCache.cpp
tmontgomery: The RetransmitFilter is the best way I can envision now for allowing individual streams to be ERRORed or CANCELed during resume.

test/resume/TcpResumeServer.cpp
tmontgomery: Due to the way the API works and the internal objects, we need a new ReactiveSocket on the server side (from fromServerConnection) to resume on. i.e. a new connection comes in and we don't know until a RESUME or SETUP is seen if it will be a resume or not. So, the old ReactiveSocket is an old session. And the new ReactiveSocket is a resumed session. Which means we need to move the streams and resume state to the new ReactiveSocket.

Also, the user only has ReactiveSocket to associate with a Resume Token.

src/ReactiveSocket.h
tmontgomery: I think it would be better to let the application track the tokens and associations. So, that means it would be external.

src/ReactiveSocket.cpp
tmontgomery: It's an async operation. The RESUME_OK will dictate whether it is possible to resume or not.

tmontgomery:
The basic logic for resuming idle connections (RESUME and RESUME_OK) seems to be working correctly. Exchange of positions and checks seem to function fine.

We need to sort out a number of issues, though, API-wise that are related to resuming operation.

- The individual streams (Server side) must not be closed when a connection closes if resume is desired. But must time out. Should this be individual stream time outs or just a total timeout?
- Should the application be the one responsible for explicitly determining ReactiveSocket lifetime (Server side)?
- What is the best means to inform the Server side application of the Resume ID Token in the SETUP? Should it be a property of the ReactiveSocket? or just a callback?
- Should the application (Server side) be responsible for associating an "old" ReactiveSocket state with a "new" (freshly connected and sent RESUME) ReactiveSocket? Or should this just be handled internally? If handled internally, then the entire ownership model of the ReactiveSocket will likely need to be changed.

These are all server side questions, I think. The client side doesn't have to do as much.

tmontgomery:
This could be merged now.

PR #117, opened 2016-08-30
Add support for RequestResponse

Cristinag:
Summary:
A REQUEST_RESPONSE frame is sent when subscriber calls ReactiveSocket::requestResponse(), and the response payload is delivered back to the subscriber only after it requests it with a request() call.
We send back one single frames from the publisher to the subscriber, containing the payload and also the COMPLETE flag;

Test Plan:
Added ReactiveSocket test.

Also tested java subscriber -> C++ publisher with:
publisher: ./tcpServer
subscriber: reactivesocket-cli --rr -i Hello tcp://localhost:9898

### DIFF
--- a/src/Frame.h
+++ b/src/Frame.h
@@ -38,7 +38,7 @@ enum class FrameType : uint16_t {
   SETUP = 0x0001,
   LEASE = 0x0002,
   KEEPALIVE = 0x0003,
-  // REQUEST_RESPONSE = 0x0004,
+  REQUEST_RESPONSE = 0x0004,
   REQUEST_FNF = 0x0005,
   REQUEST_STREAM = 0x0006,
   REQUEST_SUB = 0x0007,
@@ -217,6 +217,29 @@ class Frame_REQUEST_CHANNEL : public Frame_REQUEST_Base {
   Frame_REQUEST_CHANNEL(StreamId streamId, FrameFlags flags, Payload payload)
       : Frame_REQUEST_CHANNEL(streamId, flags, 0, std::move(payload)) {}
 };
+
+class Frame_REQUEST_RESPONSE {
+ public:
+  static constexpr bool Trait_CarriesAllowance = false;
+
+  Frame_REQUEST_RESPONSE() = default;
+  Frame_REQUEST_RESPONSE(StreamId streamId, FrameFlags flags, Payload payload)
+      : header_(
+            FrameType::REQUEST_RESPONSE,
+            flags | payload.getFlags(),
+            streamId),
+        payload_(std::move(payload)) {
+    payload_.checkFlags(header_.flags_); // to verify the client didn't set
+    // METADATA and provided none
+  }
+
+  std::unique_ptr<folly::IOBuf> serializeOut();
+  bool deserializeFrom(std::unique_ptr<folly::IOBuf> in);
+
+  FrameHeader header_;
+  Payload payload_;
+};
+std::ostream& operator<<(std::ostream&, const Frame_REQUEST_RESPONSE&);
 
 class Frame_REQUEST_FNF {
  public:
--- a/src/automata/RequestResponseRequester.cpp
+++ b/src/automata/RequestResponseRequester.cpp
@@ -12,8 +12,6 @@ void RequestResponseRequesterBase::subscribe(Subscriber<Payload>& subscriber) {
   // FIXME
   // Subscriber::onSubscribe is delivered externally, as it may attempt to
   // synchronously deliver Subscriber::request.
-  // Q: why do we want to avoid Subscriber::request being delivered
-  // synchronously?
 }
 
 void RequestResponseRequesterBase::onNext(Payload request) {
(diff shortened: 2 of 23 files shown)


src/Frame.h
```diff
@@ -218,6 +218,29 @@ class Frame_REQUEST_CHANNEL : public Frame_REQUEST_Base {
...
+  static constexpr bool Trait_CarriesAllowance = false;
+
+  Frame_REQUEST_RESPONSE() = default;
+  Frame_REQUEST_RESPONSE(StreamId streamId, FrameFlags flags, Payload payload)
+      : header_(
+            FrameType::REQUEST_RESPONSE,
+            flags | payload.getFlags(),
+            streamId),
+        payload_(std::move(payload)) {
+    payload_.checkFlags(header_.flags_); // to verify the client didn't set
```
yschimke: This comment points out that we should probably set metadata flag completely based on the presence or not.  Generally this is an existing problem, but REQUEST_FNF also does "flags | payload.getFlags()" while other frame types don't.

test/FrameTest.cpp
```diff
@@ -205,6 +205,19 @@ TEST_F(FrameTest, Frame_LEASE) {
   EXPECT_EQ(numberOfRequests, frame.numberOfRequests_);
 }
 
+TEST_F(FrameTest, Frame_REQUEST_RESPONSE) {
+  uint32_t streamId = 42;
+  FrameFlags flags = FrameFlags_METADATA;
```
yschimke: maybe more interesting that if we don't set this flag, that it gets correctly set.

yschimke:
Change looks great to me, you might get a more thorough C++ feedback from @lehecka

yschimke:
BTW just tested Java <-> Java and it seems to work

Server:

```
$ reactivesocket-cli -i "Server" --server tcp://localhost:9898
Client
```

Client:

```
$ reactivesocket-cli -i "Client" --rr tcp://localhost:9898
Server
```

src/Frame.h
lehecka: agree. The metadata flag should be sent only by presence of payload.metadata. Just need to make sure the user can't set it from outside.
We don't have to treat this in this PR.

src/automata/RequestResponseRequester.cpp
```diff
@@ -0,0 +1,125 @@
...
+  switch (state_) {
+    case State::NEW: {
+      state_ = State::REQUESTED;
+      Frame_REQUEST_RESPONSE frame(
+          streamId_, FrameFlags_EMPTY, std::move(std::move(request)));
+      connection_->outputFrameOrEnqueue(frame.serializeOut());
+      break;
+    }
+    case State::REQUESTED:
+      break;
```
lehecka: since we allow only one payload to be sent up, should we error out in this case and also on CLOSED?

src/automata/RequestResponseRequester.cpp
```diff
@@ -0,0 +1,125 @@
...
+      // Spontaneous ::endStream signal means an error.
+      DCHECK(StreamCompletionSignal::GRACEFUL != signal);
+      state_ = State::CLOSED;
+      break;
+    case State::CLOSED:
+      break;
+  }
+  // Q: don't really understand what the comment below means or if it applies
+  // in this case as well?
+  // FIXME: switch on signal and verify that callsites propagate stream errors
```
lehecka: I don't understand that one either. Feel free to keep it in and we will get to it later. Since you copy/pasted it, we will get more context once we will tackle this issue.

src/automata/RequestResponseRequester.h
```diff
@@ -0,0 +1,80 @@
...
+class exception_wrapper;
+}
+
+namespace reactivesocket {
+
+enum class StreamCompletionSignal;
+
+/// Implementation of stream automaton that represents a RequestResponse
+/// requester
+class RequestResponseRequesterBase : public MixinTerminator {
```
lehecka: maybe you want to inherit from LoggingMixin<MixinTerminator>

test/FrameTest.cpp
Cristinag: (y) as lehecka suggested, I'll do this in another PR

yschimke: Great, it's not urgent so create a one line issue to track otherwise.

Cristinag:
@yschimke I think the reason testing with reactivesocket-cli  fails currently is that the Java client expects one RESPONSE frame with the COMPLETE flag set, while the C++ server sends back two frames (RESPONSE + COMPLETE). 

 Talked with @lehecka  and it seems from the protocol definition that we should send back only one frame, so working on fixing that now.

src/automata/RequestResponseRequester.cpp
Cristinag: I think for State::REQUESTED we can error out, however State::CANCEL seems possible if the subscriber cancels the subscription immediately. e.g. 

 ReactiveSocket::requestResponse() 
     ->Subscriber::onSubscribe() calls (Subscription)RequestResponseRequester::cancel()
    -> RequestResponseRequester::onNext() // state is canceled

PR #118, opened 2016-08-30
added x86_64 version of SPSC ring buffer

tmontgomery:
Initial commit of One-To-One ringbuffer for resume cache and for new type of DuplexConnection for testing.

### DIFF
--- a/src/concurrent/OneToOneRingBuffer.h
+++ b/src/concurrent/OneToOneRingBuffer.h
@@ -10,22 +10,22 @@ namespace reactivesocket {
 
 namespace RingBufferDescriptor {
 
-constexpr static const size_t CACHE_LINE_LENGTH = 64;
+constexpr static const size_t kCACHE_LINE_LENGTH = 64;
 
 #pragma pack(push)
 #pragma pack(4)
 struct RingBufferDescriptorDefn {
-  std::uint8_t prePad[(2 * CACHE_LINE_LENGTH)];
+  std::uint8_t prePad[(2 * kCACHE_LINE_LENGTH)];
   std::uint64_t tailPosition;
-  std::int8_t tailPositionPad[(2 * CACHE_LINE_LENGTH) - sizeof(std::uint64_t)];
+  std::int8_t tailPositionPad[(2 * kCACHE_LINE_LENGTH) - sizeof(std::uint64_t)];
   std::uint64_t headCachePosition;
   std::int8_t
-      headCachePositionPad[(2 * CACHE_LINE_LENGTH) - sizeof(std::uint64_t)];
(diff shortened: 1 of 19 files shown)


lehecka:
haven't really looked deeply yet, but one thing we should do is to run devtools/format_all.sh on the change.

tmontgomery:
@lehecka done

src/concurrent/OneToOneRingBuffer.h
```diff
@@ -0,0 +1,236 @@
...
+
+#include <cstddef>
+
+#include "src/concurrent/atomic64.h"
+
+namespace reactivesocket {
+
+namespace RingBufferDescriptor {
+
+constexpr static const size_t CACHE_LINE_LENGTH = 64;
```
lehecka: following fb coding guidelines.. (if you feel we shouldn't follow them, please ignore)
constants should start with k... prefix: kCACHE_LINE_LENGTH

src/concurrent/OneToOneRingBuffer.h
```diff
@@ -0,0 +1,236 @@
...
+            {
+                head = getUInt64Volatile(headPosition_);
+                headIndex = static_cast<uint32_t>(head) & mask;
+
+                if (requiredCapacity > headIndex)
+                {
+                    return RingBufferDescriptor::INSUFFICIENT_SPACE;
+                }
+
+                putUInt64Ordered(headCachePosition_, head);
```
lehecka: not being a big expert on memory models...
do we need to use these specialized read/write methods instead of using std::atomic?

And from a higher level perspective. The whole RS-cpp implementation is written in the model that there is one thread executing the code. We chose single threaded model. With that regards, how much simpler would the OneToOneRingBuffer class look like if we don't have to deal with concurrency? Would you agree that the complexity associated with concurrency would not be used?

src/concurrent/atomic64_gcc_x86_64.h
```diff
@@ -0,0 +1,77 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
```
lehecka: do we have to provide clang version of this file for the clang compiler? FB is using clang for a lot of build process tools, including running tests.

src/concurrent/OneToOneRingBuffer.h
tmontgomery: sure. Can do that.

src/concurrent/atomic64_gcc_x86_64.h
tmontgomery: clang should be fine anyway. The same checks for platform and compiler of gcc work with clang, thankfully.

yschimke:
As a rewrite of existing code, with unit tests and not currently actively used.  I'm going to take the finer details on faith.

PR #119, opened 2016-09-02
Get createManagedInstance to work with Subscriber<T> where T != Payload

maxwellsayles:
With -Werror and -Woverloaded-virtual, we cannot use `createManagedInstance` with a `Subscriber<T>` where `T != Payload`.

test/simple/SubscriberNonPayloadTest.cpp will fail without the changes to MemoryMixin.h

Other mixins also refer to Payload directly, but I can fix those up in subsequent diffs.

### DIFF
--- a/src/mixins/MemoryMixin.h
+++ b/src/mixins/MemoryMixin.h
@@ -30,7 +30,7 @@ namespace reactivesocket {
 /// implicitly specified as virtual, depending on whether the Base class
 /// is implementing the (virtual) methods of the
 /// Subscription or the Subscriber interface.
-template <typename Base>
+template <typename Base, typename Payload = Payload>
 class MemoryMixin : public Base {
   static_assert(
       std::is_base_of<IntrusiveDeleter, Base>::value,
@@ -126,10 +126,11 @@ class WithIntrusiveDeleter {
 
 } // namespace details
 
-template <typename Base, typename... TArgs>
+template <typename Base, typename Payload = Payload, typename... TArgs>
 Base& createManagedInstance(TArgs&&... args) {
   auto* instance =
-      new MemoryMixin<typename details::WithIntrusiveDeleter<Base>::T>(
+      new MemoryMixin<typename details::WithIntrusiveDeleter<Base>::T,
+                      Payload>(
           std::forward<TArgs>(args)...);
   return *instance;
 }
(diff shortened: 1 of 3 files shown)


src/mixins/MemoryMixin.h
```diff
@@ -30,7 +30,7 @@ namespace reactivesocket {
 /// implicitly specified as virtual, depending on whether the Base class
 /// is implementing the (virtual) methods of the
 /// Subscription or the Subscriber interface.
-template <typename Base>
+template <typename Base, typename Payload = Payload>
```
yschimke: If I understand the intention of your change, shouldn't this be?

`typename T = Payload`

Reactive Streams isn't tied to Payload https://github.com/ReactiveSocket/reactive-streams-cpp/blob/master/include/ReactiveStreams.h

maxwellsayles: Sure `typename T = Payload`.  The issue is on line 71 below where we have

  void onNext(Payload payload) {

I want `Payload` to be generic.   `typename T = Payload` might be more explicit and then I'll change `Payload` elsewhere.

src/mixins/MemoryMixin.h
```diff
@@ -126,10 +126,11 @@ class WithIntrusiveDeleter {
 
 } // namespace details
 
-template <typename Base, typename... TArgs>
+template <typename Base, typename Payload = Payload, typename... TArgs>
 Base& createManagedInstance(TArgs&&... args) {
   auto* instance =
-      new MemoryMixin<typename details::WithIntrusiveDeleter<Base>::T>(
+      new MemoryMixin<typename details::WithIntrusiveDeleter<Base>::T,
+                      Payload>(
```
maxwellsayles: Actually, when you run this with

> cmake -DCMAKE_BUILD_TYPE=DEBUG

I get
```
  ==7521==ERROR: LeakSanitizer: detected memory leaks
```
This is true even before the changes this diff introduces.

More generally, I think something like
```
return *(new MyClass);
```
leads to an awkward API and memory model that's difficult to manage.

test/simple/SubscriberNonPayloadTest.cpp
```diff
@@ -0,0 +1,30 @@
...
+ public:
+  ~FooSubscriber() override = default;
+  void onSubscribe(Subscription&) override {}
+  void onNext(Foo) override {}
+  void onComplete() override {}
+  void onError(folly::exception_wrapper) override {}
+};
+
+int main(int argc, char** argv) {
+auto& m = createManagedInstance<FooSubscriber, Foo>();
```
maxwellsayles: Looks like despite being "managed", this might still be leaking.  This is the case even before the changes in this diff.

PR #120, opened 2016-09-02
createManagedInstance causes Subscriber to be leaked.

maxwellsayles:
This is a very simple test binary.  It just creates a Subscriber using `createManagedInstance`.  This instance appears to be leaked when I build and run it with

> cmake ../ -DCMAKE_BUILD_TYPE=DEBUG && make

```
$ ./leaked_subscriber

=================================================================
==9764==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 24 byte(s) in 1 object(s) allocated from:
    #0 0x7f37b1362532 in operator new(unsigned long) (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x99532)
        #1 0x40a010 in FooSubscriber& reactivesocket::createManagedInstance<FooSubscriber>() (/home/maxwellsayles/github/reactivesocket-cpp/build/leaked_subscriber+0x40a010)
	    #2 0x409362 in main /home/maxwellsayles/github/reactivesocket-cpp/test/simple/LeakedSubscriber.cpp:20
	        #3 0x7f37b01cd82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)

SUMMARY: AddressSanitizer: 24 byte(s) leaked in 1 allocation(s).
```

I'm probably doing it wrong though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -289,4 +289,18 @@ target_link_libraries(
 
 add_dependencies(tcpresumeserver gmock)
 
+add_executable(
+  leaked_subscriber
+  test/simple/LeakedSubscriber.cpp
+  )
+
+target_link_libraries(
+  leaked_subscriber
+  ReactiveSocket
+  ${FOLLY_LIBRARIES}
+  ${CMAKE_THREAD_LIBS_INIT}
+  )
+
+add_dependencies(leaked_subscriber ReactiveSocket ReactiveStreams)
+
 # EOF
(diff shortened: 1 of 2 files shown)


yschimke:
So the point of this feature is to make the object lifecycle match the events, so it will get leaked unless it sees those terminal events.  See https://github.com/ReactiveSocket/reactivesocket-cpp/blob/master/src/mixins/MemoryMixin.h

For anything more involved I'll defer to @stupaq-fb and @lehecka

maxwellsayles:
@yschimke, thanks.  I didn't look at the code closely enough.

PR #127, opened 2016-09-16
propogate errors

yschimke:
Two changes 

1) TCP Duplex connection treats a EOF or write error as an EOF.  
2) ConsumerMix propogates a CONNECTION_ERROR via onError instead of OnComplete


src/mixins/ConsumerMixin.h
```diff
@@ -48,8 +48,14 @@ class ConsumerMixin : public Base {
  protected:
   /// @{
   void endStream(StreamCompletionSignal signal) {
-    // FIXME: switch on signal and verify that callsites propagate stream errors
-    consumingSubscriber_.onComplete();
+    switch (signal) {
+      case  StreamCompletionSignal::CONNECTION_ERROR:
```
stupaq-fb: Can you reformat?

src/tcp/TcpDuplexConnection.cpp
```diff
@@ -64,7 +66,7 @@ void TcpDuplexConnection::readDataAvailable(size_t len) noexcept {
 }
 
 void TcpDuplexConnection::readEOF() noexcept {
-  inputSubscriber_.onComplete();
+  inputSubscriber_.onError(std::runtime_error("connection closed"));
```
stupaq-fb: Is it possible to have write err and then read EOF? In this case this will SEGV, right?

test/tcp/TcpServer.cpp
```diff
@@ -72,16 +72,16 @@ class ServerRequestHandler : public DefaultRequestHandler {
 
   void handleFireAndForgetRequest(Payload request) override {
     LOG(INFO) << "ServerRequestHandler.handleFireAndForgetRequest " << request
-              << "\n";
+              << std::endl;
```
stupaq-fb: Why is this needed?

test/tcp/TcpServer.cpp
```diff
@@ -72,16 +72,16 @@ class ServerRequestHandler : public DefaultRequestHandler {
...
   }
 
   void handleMetadataPush(std::unique_ptr<folly::IOBuf> request) override {
     LOG(INFO) << "ServerRequestHandler.handleMetadataPush "
-              << request->moveToFbString() << "\n";
+              << request->moveToFbString() << std::endl;
```
stupaq-fb: Ditto

test/tcp/TcpServer.cpp
```diff
@@ -72,16 +72,16 @@ class ServerRequestHandler : public DefaultRequestHandler {
...
   }
 
   void handleSetupPayload(ConnectionSetupPayload request) override {
-    LOG(INFO) << "ServerRequestHandler.handleSetupPayload " << request << "\n";
+    LOG(INFO) << "ServerRequestHandler.handleSetupPayload " << request << std::endl;
```
stupaq-fb: Ditto.

src/tcp/TcpDuplexConnection.cpp
yschimke: checking inputSubscriber_ is enough, right?

stupaq-fb: I believe so.

test/tcp/TcpServer.cpp
lehecka: we shouldn't be writing new lines into LOG statements, it would make grep filtering impossible.

yschimke: That is gone already, just showing because of github reviews AFAIK.  I will merge this is it passes tests.

PR #129, opened 2016-09-19
updates for cleaner build wrt C++14 and update gmock/gtest to 1.8 release

tmontgomery:
set std to c++14 and update to gmock 1.8.0.


tmontgomery:
FYI. the std=c++14 change was necessary for me based on updated folly (9/6/2016?) via brew on MacOS.

PR #130, opened 2016-09-20
temporary fix for lifetime of FramedReader

lehecka:
There is a problem with the FramedReader which is calling onNext in the loop parsing incoming frames. If one of the frame is refused and ConnectionAutomaton closes the DuplexConnection, the caller and callee are released and the instance of FramedReader is destroyed while still in the loop.
This fix is fixing the problem by making sure the instance is not destroyed while in the method.

I added a unit test for this case. I verified that without the fix there is SEGFAULT. With the fix there is still a problem with use-after-free in the unit test. The problem is not possible to address properly without changing the memory management around passing and referencing subscriptions and duplex connection. Will address this in the near future.
For now the problem is fixed.


yschimke:
I think we should have a rethink around the lifecycle events and memory management for reactivesocket-cpp.  Even perhaps starting with some base assumptions e.g. could a single threaded design using clean shared_ptrs (std or alike) make sense here?

Mainly because I can't see how to get to a cleaner model by adding more code.  But maybe someone else can suggest.

In that context, fixing this actual bug with a temporary fix seems appropriate to me.

tmontgomery:
@yschimke I am looking at this exact thing off and on. There is no need to complicate things further with our own SmartPointer, IMO.

PR #135, opened 2016-09-26
use folly::to instead of std::to_string

yschimke:
build compatibility fix

### DIFF
--- a/src/ConnectionAutomaton.cpp
+++ b/src/ConnectionAutomaton.cpp
@@ -6,6 +6,7 @@
 
 #include <folly/ExceptionWrapper.h>
 #include <folly/Optional.h>
+#include <folly/String.h>
 #include <folly/io/IOBuf.h>
 #include <glog/logging.h>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


stupaq-fb:
Please don't. It only makes removing folly dependency harder and doesn't change anything. What would be the reason to change it?

yschimke:
We already use it

```
$ git grep folly::to
src/ConnectionAutomaton.cpp:        Frame_ERROR::invalid("unknown stream " + folly::to<std::string>(streamId))
src/Payload.cpp:                    ? folly::to<std::string>(
src/Payload.cpp:                                 ? folly::to<std::string>(
tck-test/TypedCommands.h:    return folly::to<int>(command_.params().at(1));
tck-test/TypedCommands.h:    return folly::to<int>(command_.params().at(3));
tck-test/TypedCommands.h:    return folly::to<size_t>(command_.params().at(3));
test/framed/FramedWriterTest.cpp:            folly::to<std::string>(
test/framed/FramedWriterTest.cpp:            folly::to<std::string>(
```

And the existing code doesn't appear to work in our android build

stupaq-fb:
Oh, right. std::to_string is unimplemented, I remember hacking it on top of stringstream in RocketSpeed.

Issue #121, opened 2016-09-09
Create unit tests for TCP DuplexConnection


yschimke:

yschimke:
I suggest using MockAsyncSocket etc to test this.

PR #146, opened 2016-10-13
Add basic unit tests for TcpDuplexConnection

foghina:
Not sure this is enough to close out #121, but it's a start.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,6 +179,7 @@ add_executable(
   test/Test.cpp
   test/MockStats.h
   test/concurrent/OneToOneRingBufferTest.cpp
+  test/tcp/TcpDuplexConnectionTest.cpp
 )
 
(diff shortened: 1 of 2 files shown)


foghina:
Looks like folly needs to be updated in `reactivesocket-cpp-dependencies`.

yschimke:
Leave this here, we need a fresh build of homebrew folly.  Once that is available, I'll be more keen to update the travis version also.

foghina:
I'm using the latest homebrew folly, and it seems fresher than whatever travis has (i.e. this test passes). Or are you waiting for an even newer version?

yschimke:
I'm looking for a fresher version, e.g. about 3 days old.

benjchristensen:
@foghina thanks for getting involved and contributing!

lehecka:
as there is no progress on this. I will close this PR.

PR #147, opened 2016-10-15
new Memory model

lehecka:
This change is proposing a new memory model.

The old memory model was based on terminating signals (onComplete/onError for Subscriber and cancel for Subscription) where the instance destroyed itself (delete this) when receiving the signal. The main advantage of such approach was that we would guarantee that on terminating signal the instance is destroyed. Also in this model we don't have to deal with circular dependencies. Over the last 6 months of using the api we gained some experience and these are some of the main drawbacks of such memory model:
* memory management issues: it was very easy for the user to fall in a trap of "use-after-free" issues. We had several non-trivial bugs on iOS and fbcode.
* the memory model was not obvious from the API signatures. It was hard for the newcomers to learn and understand the model especially around corner cases.
* user had to use custom smart pointer wrappers (SubscriptionPtr<T>, SubscriberPtr<T>) to do the termination handshake right. Otherwise they would often cause memory corruption.
* non-trivial chaining of Subscribers was hard to implement. This is mostly due to the fact that from any method the user can send a terminating signal (eq. from onSubscribe the user can call subscription::cancel) which would start destroying instances while their methods are still on the stack.
* even simple unit tests had often memory bugs, especially use-after-free or memory corruption. This was mainly because the API wasn't clear as to whether the object has to be allocated on stack vs. heap.

We decided to change the memory model which would be easier to use. This change is proposing a memory model with the following properties:
* the memory model should be obvious from the public API.
* we should use standard tools like std::shared_ptr and std::unique_ptr which everybody understands
* terminating signal and object lifetime should be separated. So even if the user will use the API incorrectly (eq. wrong terminating sequence) the program will not corrupt memory. 
* circular dependencies should be broken in terminating method (onComplete/onError/cancel).

Please review the change.


stupaq-fb:
Hmm, sounds interesting (and I very much agree with current model being hard to understand for newcomers), but I cannot see the link :/.


On 14 Oct 2016, at 18:00, Ondrej Lehecka <notifications@github.com<mailto:notifications@github.com>> wrote:


This change is proposing a new memory model.

The old memory model was based on terminating signals (onComplete/onError for Subscriber and cancel for Subscription) where the instance destroyed itself (delete this) when receiving the signal. The main advantage of such approach was that we would guarantee that on terminating signal the instance is destroyed. Also in this model we don't have to deal with circular dependencies. Over the last 6 months of using the api we gained some experience and these are some of the main drawbacks of such memory model:

  *   memory management issues: it was very easy for the user to fall in a trap of "use-after-free" issues. We had several non-trivial bugs on iOS and fbcode.
  *   the memory model was not obvious from the API signatures. It was hard for the newcomers to learn and understand the model especially around corner cases.
  *   user had to use custom smart pointer wrappers (SubscriptionPtr, SubscriberPtr) to do the termination handshake right. Otherwise they would often cause memory corruption.
  *   non-trivial chaining of Subscribers was hard to implement. This is mostly due to the fact that from any method the user can send a terminating signal (eq. from onSubscribe the user can call subscription::cancel) which would start destroying instances while their methods are still on the stack.
  *   even simple unit tests had often memory bugs, especially use-after-free or memory corruption. This was mainly because the API wasn't clear as to whether the object has to be allocated on stack vs. heap.

We decided to change the memory model which would be easier to use. This change is proposing a memory model with the following properties:

  *   the memory model should be obvious from the public API.
  *   we should use standard tools like std::shared_ptr and std::unique_ptr which everybody understands
  *   terminating signal and object lifetime should be separated. So even if the user will use the API incorrectly (eq. wrong terminating sequence) the program will not corrupt memory.
  *   circular dependencies should be broken in terminating method (onComplete/onError/cancel).

Please review the change.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147

Commit Summary

  *   it compiles
  *   add tests passing
  *   everything compiles
  *   new unit tests enabled and working
  *   Merge branch 'master' into memoryModel
  *   comment updates

File Changes

  *   M CMakeLists.txt<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-0> (3)
  *   M src/ConnectionAutomaton.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-1> (32)
  *   M src/ConnectionAutomaton.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-2> (19)
  *   M src/DuplexConnection.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-3> (4)
  *   M src/NullRequestHandler.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-4> (30)
  *   M src/NullRequestHandler.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-5> (12)
  *   M src/ReactiveSocket.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-6> (64)
  *   M src/ReactiveSocket.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-7> (10)
  *   M src/RequestHandler.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-8> (10)
  *   M src/automata/ChannelRequester.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-9> (4)
  *   M src/automata/ChannelRequester.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-10> (2)
  *   M src/automata/RequestResponseRequester.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-11> (4)
  *   M src/automata/RequestResponseRequester.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-12> (2)
  *   M src/framed/FramedDuplexConnection.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-13> (13)
  *   M src/framed/FramedDuplexConnection.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-14> (9)
  *   M src/framed/FramedReader.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-15> (21)
  *   M src/framed/FramedReader.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-16> (24)
  *   M src/framed/FramedWriter.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-17> (6)
  *   M src/framed/FramedWriter.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-18> (9)
  *   M src/mixins/ConsumerMixin.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-19> (4)
  *   M src/mixins/ExecutorMixin.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-20> (20)
  *   M src/mixins/IntrusiveDeleter.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-21> (69)
  *   M src/mixins/LoggingMixin.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-22> (12)
  *   M src/mixins/MemoryMixin.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-23> (190)
  *   M src/mixins/MixinTerminator.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-24> (4)
  *   M src/mixins/PublisherMixin.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-25> (4)
  *   M src/mixins/SinkIfMixin.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-26> (4)
  *   M src/tcp/TcpDuplexConnection.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-27> (19)
  *   M src/tcp/TcpDuplexConnection.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-28> (11)
  *   M tck-test/TestInterpreter.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-29> (38)
  *   M tck-test/TestInterpreter.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-30> (6)
  *   M tck-test/TestSubscriber.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-31> (12)
  *   M tck-test/TestSubscriber.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-32> (2)
  *   M test/ConnectionAutomatonTest.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-33> (351)
  *   M test/InlineConnection.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-34> (65)
  *   M test/InlineConnection.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-35> (19)
  *   M test/InlineConnectionTest.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-36> (49)
  *   M test/MockRequestHandler.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-37> (26)
  *   M test/ReactiveSocketTest.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-38> (191)
  *   M test/ReactiveStreamsMocksCompat.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-39> (12)
  *   M test/framed/FramedReaderTest.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-40> (72)
  *   M test/framed/FramedWriterTest.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-41> (99)
  *   M test/resume/TcpResumeClient.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-42> (8)
  *   M test/resume/TcpResumeServer.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-43> (17)
  *   M test/simple/PrintSubscriber.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-44> (4)
  *   M test/simple/PrintSubscriber.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-45> (4)
  *   D test/streams/AllowanceSemaphoreTest.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-46> (66)
  *   M test/streams/Mocks.h<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-47> (107)
  *   D test/streams/OwnershipTest.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-48> (81)
  *   D test/streams/SmartPointersTest.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-49> (125)
  *   M test/tcp/TcpServer.cpp<https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147/files#diff-50> (18)

Patch Links:

  *   https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147.patch
  *   https://github.com/ReactiveSocket/reactivesocket-cpp/pull/147.diff

lehecka:
Github disabled my account because they thought I am a bot. 
It should be all fine now.
Please review.

src/RequestHandler.h
```diff
@@ -16,24 +16,24 @@ class RequestHandler {
...
   virtual void handleRequestStream(
       Payload request,
-      Subscriber<Payload>& response) = 0;
+      const std::shared_ptr<Subscriber<Payload>>& response) = 0;
 
   /// Handles a new inbound Subscription requested by the other end.
   virtual void handleRequestSubscription(
       Payload request,
-      Subscriber<Payload>& response) = 0;
+      const std::shared_ptr<Subscriber<Payload>>& response) = 0;
```
yschimke: For my education: Why is this better than just a `Subscriber<Payload>&`, it doesn't seem like nullability comes into play here.  Does this mean, here is a Subscriber that you can't delete, but may disappear underneath you?

test/ConnectionAutomatonTest.cpp
```diff
@@ -17,151 +17,218 @@
...
-// TODO: the following tests are functionally correct and validates the
-//       invalid frame functionality. But with the current memory model
-//       around DuplexConnection, its Subcribers and InlineConnection
-//       it is impossible to make the destruction and cleanup of the test
-//       clean and without use-after-free
-// Both tests in this file use InlineConnection which will be destructed
-// inside InlineConnection::setInput() when it calls inputSink.onSubscribe().
-// TODO(lehecka): will fix this in the next iteration, with a different
-//                memory model
+static std::unique_ptr<folly::IOBuf> makeInvalidFrameHeader() {
```
yschimke: nice!

src/mixins/IntrusiveDeleter.h
```diff
@@ -1,34 +1,35 @@
...
+//
+//#include <cassert>
+//#include <cstddef>
+//#include <memory>
+//
+//namespace reactivesocket {
+//
+///// Deletes itself using the first time the reference count, initially one,
+///// drops to zero.
+//class IntrusiveDeleter {
```
maxwellsayles: Without this, is there any chance of circular references between `std::shared_ptr<>`s ?  What I'm imagining is a `ReactiveSocket` having a `Subscriber`, which has the `ReactiveSocket`.  Would these get freed up on close?

src/ConnectionAutomaton.h
```diff
@@ -50,7 +50,8 @@ class ConnectionAutomaton :
     /// Registered as an input in the DuplexConnection.
     public Subscriber<std::unique_ptr<folly::IOBuf>>,
     /// Receives signals about connection writability.
-    public Subscription {
+    public Subscription,
+    public std::enable_shared_from_this<ConnectionAutomaton> {
```
jspahrsummers: YMMV, but I've tried using `enable_shared_from_this` in the past, and it seems _very_ hard to get right. In particular, the restriction that there must already be an `std::shared_ptr` to `this` is difficult to satisfy in all cases—and getting it wrong is undefined behavior (not necessarily a crash).

If you do want to pursue this approach, I would suggest making all constructors `private` and exposing a `static` method which returns the `std::shared_ptr` correctly.

src/DuplexConnection.h
```diff
@@ -36,13 +36,13 @@ class DuplexConnection {
   /// lifetime of the connection. The connection MUST NOT assume an ownership of
   /// provided Subscriber.
```
jspahrsummers: Looks like there are some comments about memory management that need updates.

src/RequestHandler.h
jspahrsummers: Nullability is incidental here—the real benefit is that if the implementor wants to store `response`, it can do so with proper memory management (refcounting).

stupaq-fb:
So let me put down the company a bit.

This only works if terminal signals are synchronous. Consider a Subscriber that operates on a different thread and the following pseudo-code:

// s is a shared_ptr to the Subscriber as above.
s->onNext(...);
...
s->onComplete();
// s is the last reference to the subscriber.
s.reset()

The problem here is that if s is asynchronous, it processes signals at some point after they are delivered. So at the time shared_ptr's ref count drops to zero, onComplete might not have been processed yet and the signal will arrive after d'tor of the subscriber is called.

One could address it it multiple ways.

1) Send a copy of shared pointer to a subscriber with every signal. This is deadly inefficient and makes the interface nasty, as onNext, onComplete, etc. must accept that shared_ptr.

2) Make subscriber hold a weak_ptr to itself. This is as deadly inefficient as above, except we no longer bastardise the interface.

3) Make d'tor "asynchronous" by sending a special signal and letting the Subscriber decide when to deallocate itself.

As you might have noticed, 3) is exactly the old memory model and above reasoning have been discussed on the original diff. I've chosen to piggy back on onComplete/cancel for this "asynchronous destruction" because of Occam's razor. We can separate them, if that's helpful.

TL;DR I don't think there is another way of solving the problem, but I agree the exists a different *syntax* for the old memory management model that would (maybe) be less error-prone.

Lastly, why asynchrony is so important, why not lock on destruction? Well, that would be tricky. We would have to lock entire chain of subscribers. Also, a similar counter example exists for a race between onNext and destruction, which means we would have to make every signal synchronous...

tmontgomery:
The usual method to handle the asynchronous case is what is done with RCU or QSBR by procrastinating (or deferring). A few ways to handle it that are relatively clean.

For efficiency, it is a bad idea to free on a different thread than the allocation. So, one possibility is to use an "Acceptor" for SETUP/RESUME and thus handler creation (as has been discussed). That thread can then be the procrastination cleanup thread as well. It would be the DuplexConnection reading thread, IIRC.

lehecka:
Let me answer most of the fundamental questions here:

In the proposed model I don't think we have a restriction that terminal signals have to be delivered synchronously. Let's say we want to execute terminal signal on a different thread (eq. onComplete). We can have 2 ways to do that:
1. let the user code marshal the call onto the right thread:
// assuming std::shared_ptr<Subscriber> subscriber = std::make_shared<MySubscriber>() earlier in the code
...
thread2->eventBase()->runInEventBase([subscriber](){
  subscriber->onComplete();
}
subscriber.reset(); // its safe to release the pointer..

In this case the user code has to make sure the thread2 is calling into a living object so it explicitly creates copy of subscriber shared_ptr. 
(this is pretty standard multithreaded code, nothing special for ReactiveSocket)

2. implementation of onComplete will marshal on the right thread:
The implementation will do something like this:
void onComplete() {
  auto thisPtr = shared_from_this();
  thread2->eventBase()->runInEventBase([thisPtr](){
    thisPtr->onComplete();
  }
}

// from thread1:
subscriber->onComplete();
subscriber.reset(); // its safe to release the pointer..

In this case the implementation will create a copy of shared_ptr from this. By having the subscriber class inherit from std::enable_shared_from_this we will enable the implementation to create the copy of the shared_ptr to this.

The current proposal is using the second approach.

BTW: The asynchronous termination is exercised and verified in ReactiveSocketConcurrencyTest.

src/ConnectionAutomaton.h
lehecka: Thanks for the comment here. I found it also confusing when I am using classes deriving from enable_shared_from_this and allocate them on the stack and then be surprised it is throwing execptions about bad weak_ptr. I was going to do what you are suggesting but I didn't like the duplicated code for the duplicated creator methods.
Having the API accepting only shared_ptrs of objects, the user is forced to allocate them all on heap so I never fell in the trap of bad weak_ptr in any of the unit tests.

src/mixins/IntrusiveDeleter.h
lehecka: The typical circular reference in the system happens when you implement a class implementing Subscriber and Subscription interfaces and have the class be a middle man between 2 subscribers.
I tested that scenario and verified that since we are releasing smart pointers on the terminating signal, we break any circular references.

src/mixins/MemoryMixin.h
```diff
@@ -32,105 +36,107 @@ namespace reactivesocket {
...
+//
 
 template <typename Base, typename T = Payload, typename... TArgs>
-Base& createManagedInstance(TArgs&&... args) {
-  auto* instance =
-      new MemoryMixin<typename details::WithIntrusiveDeleter<Base>::T, T>(
-          std::forward<TArgs>(args)...);
-  return *instance;
+std::shared_ptr<Base> createManagedInstance(TArgs&&... args) {
+  return std::make_shared<Base>(std::forward<TArgs>(args)...);
```
lehecka: I agree. I will get rid of the method completely in another diff which will be some more cleanup. I want to minimize the amount of modified lines. The change is huge already.

stupaq-fb:
Oh, right I missed the trick with `enable_shared_from_this`, that is clever. With this in mind, this approach seems to be isomorphic to what I described as the second option, since `enable_shared_from_this` is practically a `weak_ptr` to this.

My concerns are twofold then:
* inability to *not* use `shared_ptr`,
* performance.

I definitely buy the argument that it's harder to use-after-free in this approach (but easier to leak). I think, given our use cases, this is a totally fair trade-off. You have my bow.

I do approve. My understanding is that this change is hard to merge due to it depending on the other repo, so merge when it makes sense to you.

PR #154, opened 2016-10-20
passing Executor instance to ExecutorMixin

lehecka:
This change allows specifying Executor instance in the constructor of ExecutorMixin.
The default Executor is still QueuedImmediateExecutor so the the default behavior stays untouched.
The Executor is used to synchronize access to Subscriber and Subscription methods.  
[current behavior]
In the case when the Subscriber/Subscription is executing client code callback and the client code tries to call back into the Subscriber/Subscription again the QueuedImmediateExecutor queues such call. Once the client code returns back to Subscriber/Subscription code, the queued methods are executed one by one.

We can achieve the same behavior with the EventBase Executor which we can now pass into the constructor of ExecutorMixin. And what we get as a bonus is that we can call methods of Subscriber/Subscription from any thread and they will be automatically marshalled to the EventBase thread.

The intent of this change it to remove all boilerplate code such as:

eventBase->runInEventBaseThread([](){
  subscription->request(n)
  subscription->cancel()
  subscriber->onNext(p);
  subscriber->onComplete();
  subscriber->onError();
}

and allow direct call to the methods without worrying about concurrency issues.

Please note that only Subscriber/Subscription methods will be safe to call from different threads. The methods of ReactiveSocket instance should be still called from the same thread which created the instance. (If we want to we can cover concurrent access to the RS instance methods in another diff);

This should simplify our iOS an gateway code.
@jspahrsummers, @MartinSix

Please review.

I added extensive unit tests to test access of the methods from a different thread.
I also used the change to simplify the tck-test library.


src/Payload.h
```diff
@@ -33,7 +33,7 @@ struct Payload {
       const std::string& metadata = std::string());
 
   explicit operator bool() const {
-    return data != nullptr && metadata != nullptr;
+    return data != nullptr || metadata != nullptr;
```
yschimke: nice catch

src/ReactiveSocket.cpp
```diff
@@ -271,10 +322,24 @@ bool ReactiveSocket::createResponder(
...
+                {connection_, streamId},
+                executor ? *executor : defaultExecutor()};
+            automaton = new RequestResponseResponder(params);
+            connection_->addStream(streamId, *automaton);
+            return *automaton;
+          });
+      handler_->handleRequestResponse(
+          std::move(frame.payload_), subscriberFactory);
+      // we need to create a responder to at least close the stream
+      if (!automaton) {
```
yschimke: why would this ever be unset?

jspahrsummers:
I like the premise here—that `Subscriber` and `Subscription` methods should be safe to call from any thread—but the mixin-based approach (still) seems a little too magical.

The `folly::Executor& executor = defaultExecutor()` default parameter makes a lot of sense to me. Could we use that in all public methods, and then ask implementors of `Subscriber` to override protected methods that we guarantee will be invoked in the correct place?

src/ReactiveSocket.cpp
```diff
@@ -59,9 +62,14 @@ ReactiveSocket::ReactiveSocket(
       keepaliveTimer_(std::move(keepaliveTimer)),
       resumeSocketListener_(resumeListener) {}
 
+folly::Executor& ReactiveSocket::defaultExecutor() {
```
yschimke: It seems like in this case you have to make the calls from the correct thread to have correctness.  Is there a folly executor decorator or something similar that can do those checks for us?

lehecka: I didnt find any class like this which would give us the checks whether we are on the right thread.
I would implement such checks in later diff. I assume those would be only debug checks.

lehecka:
To Martin:
You are exactly right. This is the direction I am walking towards. I want to get rid of the ExecutorMixin and MemoryMixin and collapse their responsibilities into SubscriberBase/SubscriptionBase classes which would provide the conveniences or you with a clean and easy to understand api.
To Justin and Martin: bare with me. This is only the first change of a series.

Essentially in this change I introduced a way to provide Executor to the constructor and have the public methods (onNext, onComplete, onError, cancel, request) do the right thing.

I will simplify the inheritance in the following diffs.

stupaq-fb:
The approach with base class is not composable, so maybe, if mixins are too implicit, a wrapper classes would be better? Mixins are like wrappers resolved at compile time (less code, etc.).

Instead of MixinExecutor, we could have MixinWrapper, that takes an owning pointer to a Subscriber/Subscription and exposes the same API. It does the same thing as mixin: switches context to executor and then calls corresponding signal.

If we use base classes instead, then there is no way of composing one wrapper with another, since every wrapper changes the API slightly.

martinsix:
Why do they have to be composable? Are there that many different combinations?

stupaq-fb:
I noticed I said "MixinWrapper", I meant "WrapperExecutor"...

Good question, @martinsix. I don't know that, but wrapper has barely any addition cost over virtual inheritance (and both are more costly than mixins in terms of perf and allocation overhead), so I would go with the one that gives the most in terms of flexibility.

Wrapper is closer to shared_ptr, hence might be more intuitive. Also, one can decide at runtime whether to use wrappers or not. This allows for usage of the same Subscriber/Subscription implementation without automatic memory management, which is convenient in tests.

Both mixins and wrappers provide separation of concerns: you write your application logic in one class and can add "automatic" memory management or not. This is not the case if the Subscriber/Subscription that contains application logic is inherited from.

lehecka:
landing this as tests are passing and to unblock myself in completing the effort of delivering the new memory model.

PR #157, opened 2016-10-26
introducing SubscriberBase and SubscriptionBase

lehecka:
SubscriberBase and SubscriptionBase are new base classes designed for users implementing their own Subscriber/Subscription classes. The base classes allow to specify Executor in the constructor. 
* All calls to onSubscribe/onNext/onComplete/onError and request/cancel are synchronized via provided or default executor. The Executor makes sure there will be only one of the call on the stack. This is keeping our current design which is very useful for correctness and safety.
* If users provides an EventBase as an Executor, all above calls will be automatically marshaled onto the provided EventBase. (please remove all runInEventBaseThread wrappers from your code and create Subscribers with the appropriate Executor!)

Please see how the new classes are used in the test and tcp duplex connection


src/EnableSharedFromThis.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include <memory>
+
+namespace reactivesocket {
+
+class EnableSharedFromThisVirtualBase
+    : public std::enable_shared_from_this<EnableSharedFromThisVirtualBase> {};
+
+template <typename T>
+class EnableSharedFromThisBase
```
jspahrsummers: Why is this class necessary?

src/Executor.cpp
```diff
@@ -0,0 +1,37 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#include <folly/futures/QueuedImmediateExecutor.h>
+#include "src/SubscriberBase.h"
+#include "src/mixins/ExecutorMixin.h"
+
+namespace reactivesocket {
+
+// just instantiating of the template here
```
jspahrsummers: Why?

src/Executor.cpp
```diff
@@ -0,0 +1,37 @@
...
+ExecutorBase::ExecutorBase(folly::Executor& executor, bool startExecutor)
+    : executor_(executor) {
+  if (!startExecutor) {
+    pendingSignals_ = folly::make_unique<PendingSignals>();
+  }
+}
+
+void ExecutorBase::start() {
+  if (pendingSignals_) {
+    auto movedSignals = folly::makeMoveWrapper(std::move(pendingSignals_));
```
jspahrsummers: I thought we didn't need `MoveWrapper` anymore?

src/mixins/ExecutorMixin.h
```diff
@@ -19,12 +19,43 @@
...
+folly::Executor& defaultExecutor();
+
+class ExecutorBase {
+ public:
+  explicit ExecutorBase(
+      folly::Executor& executor = defaultExecutor(),
+      bool startExecutor = true);
+
+  /// We start in a queueing mode, where it merely queues signal
+  /// deliveries until ::start is invoked.
```
jspahrsummers: Why do we want to queue initially?

src/mixins/ExecutorMixin.h
```diff
@@ -19,12 +19,43 @@
...
+  ///
+  /// Calling into this method may deliver all enqueued signals immediately.
+  void start();
+
+ protected:
+  template <typename F>
+  void runInExecutor(F&& func) {
+    if (pendingSignals_) {
+      pendingSignals_->emplace_back(func);
```
jspahrsummers: Both uses of `func` in here should be `std::forward<F>(func)`.

src/mixins/ExecutorMixin.h
```diff
@@ -19,12 +19,43 @@
...
+    } else {
+      executor_.add(std::move(func));
+    }
+  }
+
+ private:
+  using PendingSignals = std::vector<std::function<void()>>;
+  std::unique_ptr<PendingSignals> pendingSignals_;
```
jspahrsummers: Why is this allocated on the heap? The vector can be moved into the executor lambda regardless.

src/SubscriberBase.h
```diff
@@ -0,0 +1,55 @@
...
+#include "src/ReactiveStreamsCompat.h"
+#include "src/mixins/ExecutorMixin.h"
+
+namespace reactivesocket {
+
+template <typename T>
+class SubscriberBaseT : public Subscriber<T>,
+                        public EnableSharedFromThisBase<SubscriberBaseT<T>>,
+                        public virtual ExecutorBase {
+  virtual void onSubscribeImpl(std::shared_ptr<Subscription> subscription) = 0;
```
jspahrsummers: Is default visibility `protected`? I didn't realize that—maybe worth adding it explicitly?

src/SubscriberBase.h
```diff
@@ -0,0 +1,55 @@
...
+  virtual void onCompleteImpl() = 0;
+  virtual void onErrorImpl(folly::exception_wrapper ex) = 0;
+
+ public:
+  using ExecutorBase::ExecutorBase;
+
+  void onSubscribe(std::shared_ptr<Subscription> subscription) override final {
+    runInExecutor(std::bind(
+        &SubscriberBaseT::onSubscribeImpl,
+        this->shared_from_this(),
```
jspahrsummers: How does `this->` change the semantics?

src/SubscriberBase.h
```diff
@@ -0,0 +1,55 @@
...
+  virtual void onSubscribeImpl(std::shared_ptr<Subscription> subscription) = 0;
+  virtual void onNextImpl(T payload) = 0;
+  virtual void onCompleteImpl() = 0;
+  virtual void onErrorImpl(folly::exception_wrapper ex) = 0;
+
+ public:
+  using ExecutorBase::ExecutorBase;
+
+  void onSubscribe(std::shared_ptr<Subscription> subscription) override final {
+    runInExecutor(std::bind(
```
jspahrsummers: I don't think there's much reason to prefer `std::bind` over a lambda, especially because lambdas are usually easier to understand.

src/SubscriptionBase.h
```diff
@@ -0,0 +1,31 @@
...
+                         public virtual ExecutorBase {
+  virtual void requestImpl(size_t n) = 0;
+  virtual void cancelImpl() = 0;
+
+ public:
+  using ExecutorBase::ExecutorBase;
+
+  void request(size_t n) override final {
+    runInExecutor(
+        std::bind(&SubscriptionBase::requestImpl, shared_from_this(), n));
```
jspahrsummers: These uses of `shared_from_this()` could lead to some very surprising behavior if a derived class wants to call one of these methods during construction.

src/EnableSharedFromThis.h
lehecka: I want user to be able to implement a class which inherits both from SubscriberBase and SubscriptionBase. The class needs to inherit std::enable_shared_from_this only once in order to work properly. Unfortunately std::enable_shared_from_this can't be inherited virtually because of some implementation details of that class.
EnableSharedFromThis is essentially a workaround which will enable you to inherit std::enable_shared_from_this only once in your class hierarchy.

src/Executor.cpp
lehecka: I don't want the template to be instantiated in every compilation unit. It is unnecessary and just adds more work for compiler and linker. That's why I declared the template to be extern. Now Executor.cpp is the only place which will instantiate the template (it could be any cpp file) so the code is generated only once.

src/mixins/ExecutorMixin.h
lehecka: I want this base class to be used for the implementation reactive socket automata where we need the functionality. We want to make some calls to the object right after the construction which can eventually call users callback, but we don't want to process user signals until we say start (the instance is fully initialized and ready).
I should write more about this in the documentation.
Good thing is that adds no overhead if not used.

src/mixins/ExecutorMixin.h
lehecka: good catch.

src/mixins/ExecutorMixin.h
lehecka: this is because we are differentiating a case when pendingSignals_ is null or pendingSignals_ is empty. Please, see method start().
If we wanted to save the allocation, we would add a boolean value signaling whether we are still just accumulating user signals or executing them right away.

src/SubscriberBase.h
lehecka: I think that's a personal preference. std::bind does exactly the same with fever lines of code.

src/SubscriberBase.h
lehecka: I know this looks strange. But without this-> the code would not compile. The reason why is that the name resolution in the templated code is excluding dependent base (base class dependent on a template argument, in our case it is EnableSharedFromThisBase<SubscriberBaseT<T>>) from lookup:
https://gcc.gnu.org/wiki/VerboseDiagnostics#dependent_base

src/SubscriptionBase.h
lehecka: In the constructor the internal weak_ptr is not set yet. It is set by the shared_ptr afterwards.
So the behavior that you will always get is exception bad_weak_ptr. Unfortunately that's what we have to live with ;(

martinsix: I think that's ok as long as it throws an error.

src/EnableSharedFromThis.h
jspahrsummers: Can you please document that?

src/SubscriptionBase.h
jspahrsummers: Only C++17 _guarantees_ that an error is thrown. An implementation is allowed to do anything it wants before then—and that could very well happen on iOS or Android.

jspahrsummers:
I requested changes in the hopes that the documentation could be updated before merging—sorry for not making that clear. 😕

src/mixins/ExecutorMixin.h
jspahrsummers: We don't really _need_ to differentiate those cases. Enqueuing an empty work item on the `Executor` isn't a big enough deal to complicate this code, IMO.

lehecka: I don't want to leave this comment unanswered...
We do need the differentiation to know whether the incoming signal should be queued or it should be processed.
If pendingSignals_ is null, we know to process the signal. if it is not null, we queue it.
If pendingSignals_.empty() == true, it means its the first signal to queue.

src/SubscriberBase.h
lehecka: sure, I will change this to lambdas. It is also much easier to put a breakpoint into lambda to see it was executed.

PR #159, opened 2016-10-27
terminating streams with onError when connection is closed

lehecka:


src/Common.cpp
```diff
@@ -0,0 +1,29 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#include "src/Common.h"
+#include <glog/logging.h>
+#include <folly/io/IOBuf.h>
+#include "src/AbstractStreamAutomaton.h"
+
+namespace reactivesocket {
+
+static const char* getTerminatingSignalErrorMessage(int terminatingSignal) {
```
jspahrsummers: Why doesn't this accept the type `StreamCompletionSignal`?

src/Common.cpp
```diff
@@ -0,0 +1,29 @@
...
+  switch ((StreamCompletionSignal)terminatingSignal) {
+    case StreamCompletionSignal::CONNECTION_END:
+      return "connection closed";
+    case StreamCompletionSignal::CONNECTION_ERROR:
+      return "connection error";
+    case StreamCompletionSignal::ERROR:
+      return "general error";
+    case StreamCompletionSignal::GRACEFUL:
+      DCHECK(false) << "throwing exception for GRACEFUL termination?";
```
jspahrsummers: Can this still return a string saying "graceful termination," just in case we need to debug it in production?

lehecka: yes, it can

yschimke:
Can we land this also before importing, this hit me

PR #163, opened 2016-11-02
reenabling unit tests in ConnectionAutomatonTest.cpp

lehecka:
This is fixing ASAN crash for the tests and also memory leak in the test due a to circular dependency.


lehecka:
merging to unblock myself with other changes.

PR #165, opened 2016-11-09
One terminal signal for SubscriberBase users

lehecka:
from the code:

//    once the user cancels the subscriber via the subscription::cancel, it is
//    guaranteed that no more callbacks will be called on the subscriber. It is safe for the client
//    to do cleanup and releasing of resources. User doesn't have to call
//    subscription::cancel in on{Complete,Error} methods.

I left most of the documentation in the code.

### DIFF
--- a/src/SubscriberBase.h
+++ b/src/SubscriberBase.h
@@ -9,6 +9,18 @@
 
 namespace reactivesocket {
 
+//
+// SubscriberBase is designed to be a general purpose base class for
+// implementors of Subscriber<T> interface. It provides 2 key features:
+// 1. ExecutorBase: You can specify the Executor you want to use to execute the
+//    on{Subscribe,Next,Complete,Error} methods. Eq. by providing EventBase
+//    as an executor, you can call the methods from any thread and the
+//    marshaling to the right EventBase thread will happen automatically.
+// 2. once the user cancels the subscriber via the subscription, it is
+//    garanteed that no more callbacks will be called. It is safe for the client
+//    to do cleanup and releasing of resources. User doesn't have to call
+//    subscription::cancel in on{Complete,Error}.
+//
 template <typename T>
 class SubscriberBaseT : public Subscriber<T>,
                         public EnableSharedFromThisBase<SubscriberBaseT<T>>,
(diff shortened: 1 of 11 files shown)


src/SubscriberBase.h
```diff
@@ -18,34 +30,115 @@ class SubscriberBaseT : public Subscriber<T>,
   virtual void onCompleteImpl() = 0;
   virtual void onErrorImpl(folly::exception_wrapper ex) = 0;
 
+  // used to be able to cancel subscription immediatelly, making sure we dont
+  // deliver any other signals after that
+  // also to break the reference cycle involving storing subscription pointer
+  // for the users of the SubscriberBase
+  class SubscriptionShim : public Subscription {
```
jspahrsummers: Why doesn't this class wrap a `Subscription` instead of a subscriber? Seems like it makes more sense to have a `Subscription` which accepts an `EventBase`, and ensures that cancellation makes other kinds of messages invalid.

src/SubscriberBase.h
```diff
@@ -18,34 +30,115 @@ class SubscriberBaseT : public Subscriber<T>,
...
   }
+
+ private:
+  // Once the subscription is cancelled we will no longer deliver any
+  // other signals.
+  // Scheduling the calls is not atomic operation so it may very well happen
+  // that 2 threads race for sending onNext and onComplete. We need to make sure
+  // that once the terminating signal is delivered we no longer try to deliver
+  // onNext.
+  std::atomic<bool> cancelled_;
```
jspahrsummers: This seems to only be used from the `Executor`, meaning this doesn't really need to be atomic.

(If it _is_ modified from elsewhere, there's a race in `onNext`.)

src/SubscriberBase.h
lehecka: can you elaborate more on the proposal? What would be the inheritance hierarchy?
currently the structure is this:
ClientsMySubscriber : derives from SubscriberBase

I didn't want to have something like
CancellationDecorator : derives from (ClientMySubscriber : derives from SubscriberBase)

I wanted to have the client code class to be the terminal leaf in the inheritance.

jspahrsummers:
We talked about this a bit offline. I'm happy for this to go in as-is (although I think it'd benefit from review from, e.g., @yschimke), as long as we follow it up with some refactoring that aims to do something like this:

 * Instead of reaching for a private atomic flag of `SubscriberBase`, move cancellation functionality into a new kind of `CancellationToken` class, which can be shared between the subscriber and subscription.
 * Create a new class of `Subscription` which wraps any other `Subscription` with this kind of atomic cancellation behavior shown here, so we build that through composition instead of inheritance.
 * Instead of inheriting the functionality of `ExecutorBase`, turn it into a new kind of `folly::Executor`, which wraps another `Executor` with queuing behaviors. The `SubscriberBase` can create this `QueuedExecutor` internally (so users don't have to), then pass it to the new subscription class.

src/SubscriberBase.h
```diff
@@ -18,34 +30,116 @@ class SubscriberBaseT : public Subscriber<T>,
...
   }
 
   void onComplete() override final {
     auto thisPtr = this->shared_from_this();
-    runInExecutor([thisPtr]() { thisPtr->onCompleteImpl(); });
+    runInExecutor([thisPtr]() {
+      if (!thisPtr->cancelled_.exchange(true)) {
+        thisPtr->onCompleteImpl();
+
+        DCHECK(thisPtr->originalSubscription_);
```
yschimke: This seems like it should be either a CHECK (to avoid segfault below) or add an IF below

src/SubscriberBase.h
```diff
@@ -18,34 +30,116 @@ class SubscriberBaseT : public Subscriber<T>,
...
   void onError(folly::exception_wrapper ex) override final {
     auto movedEx = folly::makeMoveWrapper(std::move(ex));
     auto thisPtr = this->shared_from_this();
-    runInExecutor(
-        [thisPtr, movedEx]() mutable { thisPtr->onErrorImpl(movedEx.move()); });
+    runInExecutor([thisPtr, movedEx]() mutable {
+      if (!thisPtr->cancelled_.exchange(true)) {
+        thisPtr->onErrorImpl(movedEx.move());
+
+        DCHECK(thisPtr->originalSubscription_);
```
yschimke: same as above? CHECK or add an if

PR #167, opened 2016-11-12
initial resume API for SETUP/RESUME

tmontgomery:
I've been iterating on the API for resume. Feedback welcome.

The hideous method of "resumeFromSocket" and technique has been removed. Now RequestHandler has a `handleResume` method that is called instead.

The state for the connection is kept in a `StreamState` object (name is not very good, but can't think of a better one. Open to suggestions). This state is now created by the return value of `RequestHandler::handleSetupPayload` and returned in `RequestHandler::handleResume`.

This is not perfect, but it seems cleaner. By bringing the stream state out into its own object, it can be managed easier. It also decouples it from the ReactiveSocket. RequestHandler is maybe not the perfect place to put SETUP and RESUME logic, but we could break those out into their own handler. Would love opinions on that.

The aspects of marking dirty/clean are not included in this PR. I wanted to get feedback on this first and get it sorted then include what I have been playing with for dirty/clean.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,7 +163,8 @@ add_library(
   src/EnableSharedFromThis.h
   src/SubscriberBase.h
   src/SubscriptionBase.h
-  src/Executor.cpp)
+  src/Executor.cpp
+  src/StreamState.h)
 
 target_link_libraries(
   ReactiveSocket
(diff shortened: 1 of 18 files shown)


lehecka:
I like the direction. 
One thing to address it to include pendingWrites_ in the StreamState.
It would be also nice to make the default behavior of RequestHandler such that we dont have to override the handleSetupPayload method. There are just a lot of places which are returning new instance of StreamState class.

PR #177, opened 2016-11-18
making ReactiveSocket::createResponder method static

lehecka:
For my next PR I want to be able to release the instance of ReactiveSocket while in the RequestHandler::handleRequestXXX.
If I do it today I will end up destroying the instance while in the ReactiveSocket::createResponder method in on the stack and accessing member variables.
This change is turning the method into the static method which will guarantee its not accessing any member variables.

### DIFF
--- a/src/ConnectionAutomaton.h
+++ b/src/ConnectionAutomaton.h
@@ -22,6 +22,8 @@ enum class FrameType : uint16_t;
 enum class StreamCompletionSignal;
 using StreamId = uint32_t;
 
+class ConnectionAutomaton;
+
 /// Creates, registers and spins up responder for provided new stream ID and
 /// serialised frame.
 ///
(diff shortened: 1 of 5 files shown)


src/ReactiveSocket.cpp
```diff
@@ -192,6 +193,8 @@ void ReactiveSocket::metadataPush(std::unique_ptr<folly::IOBuf> metadata) {
 }
 
 bool ReactiveSocket::createResponder(
+    const std::shared_ptr<RequestHandlerBase>& handler,
```
jspahrsummers: Looks like this method doesn't actually retain `handler` in any way, so it seems like it could just be a `const RequestHandlerBase&`.

src/ConnectionAutomaton.h
```diff
@@ -30,8 +32,10 @@ using StreamId = uint32_t;
 /// Returns true if the responder has been created successfully, false if the
 /// frame cannot start a new stream, in which case the frame (passed by a
 /// mutable referece) must not be modified.
-using StreamAutomatonFactory =
-    std::function<bool(StreamId, std::unique_ptr<folly::IOBuf>)>;
+using StreamAutomatonFactory = std::function<bool(
+    ConnectionAutomaton& connection,
```
jspahrsummers: Conversely to my comment below, this seems like a good thing to turn into a `shared_ptr`, because there's nothing in the factory API which requires that the responder be created synchronously, right?

PR #180, opened 2016-11-22
stop the keepalive timer ASAP and hold shared_ptr to ConnectionAutomaton

yschimke:


src/folly/FollyKeepaliveTimer.cpp
```diff
@@ -27,21 +27,21 @@ void FollyKeepaliveTimer::schedule() {
...
+// must be called from the same thread as start
 void FollyKeepaliveTimer::stop() {
   *running_ = false;
 }
 
-// must be called from the thread as stop
-void FollyKeepaliveTimer::start(ConnectionAutomaton* automaton) {
-  automaton_ = automaton;
+// must be called from the same thread as stop
+void FollyKeepaliveTimer::start(const std::shared_ptr<ConnectionAutomaton> connection) {
```
lehecka: param ahould be either
const std::shared_ptr<ConnectionAutomaton>& connection
or
std::shared_ptr<ConnectionAutomaton> connection (and then moved to connection_)

src/folly/FollyKeepaliveTimer.h
```diff
@@ -21,10 +21,10 @@ class FollyKeepaliveTimer : public KeepaliveTimer {
 
   void stop() override;
 
-  void start(ConnectionAutomaton* automaton) override;
+  void start(const std::shared_ptr<ConnectionAutomaton> connection) override;
 
  private:
-  ConnectionAutomaton* automaton_{nullptr};
+  std::shared_ptr<ConnectionAutomaton> connection_{nullptr};
```
lehecka: no need to initialize it to nullptr.

PR #185, opened 2016-11-23
initial draft of keepalive closing

yschimke:
An initial cut.  I'd be happy to land this with a test and probably a flag for actually closing.

- should closing be optional behaviour?
- should the acceptable delay be customisable, as opposed to each keepalive period?
- do we need keepalive data to be definable?
- should we check this on clients?
- should server close 
- if so, shoudl we use the keepalive time in the setup frame

Todo

- Unit tests

### DIFF
--- a/src/folly/FollyKeepaliveTimer.cpp
+++ b/src/folly/FollyKeepaliveTimer.cpp
@@ -27,8 +27,18 @@ void FollyKeepaliveTimer::schedule() {
   eventBase_.runAfterDelay(
       [this, running]() {
         if (*running) {
-          connection_->sendKeepalive();
-          schedule();
+          if (pending_) {
+            stop();
+
+            connection_->outputFrameOrEnqueue(
+                Frame_ERROR::connectionError("no response to keepalive")
+                    .serializeOut());
+            connection_->disconnect();
+          } else {
+            connection_->sendKeepalive();
+            pending_ = true;
+            schedule();
+          }
         }
       },
       keepaliveTime().count());
@@ -46,5 +56,9 @@ void FollyKeepaliveTimer::start(
   *running_ = true;
 
   schedule();
+}
+
+void FollyKeepaliveTimer::keepaliveReceived() {
+  pending_ = false;
 }
 }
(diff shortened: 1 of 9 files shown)


yschimke:
cc @lehecka @lexs

src/folly/FollyKeepaliveTimer.h
```diff
@@ -23,10 +23,13 @@ class FollyKeepaliveTimer : public KeepaliveTimer {
...
   void start(const std::shared_ptr<ConnectionAutomaton>& connection) override;
 
+  void keepaliveReceived() override;
+
  private:
   std::shared_ptr<ConnectionAutomaton> connection_;
   folly::EventBase& eventBase_;
   std::shared_ptr<bool> running_;
   std::chrono::milliseconds period_;
+  bool pending_{false};
```
lehecka: I suggest making this boolean member an atomic<bool> because you can potentially call it from a different thread, or add DCHECK in keepaliveReceived that it is executed on the same eventBase_.
This is only for cleanness.

src/folly/FollyKeepaliveTimer.cpp
```diff
@@ -27,8 +27,18 @@ void FollyKeepaliveTimer::schedule() {
   eventBase_.runAfterDelay(
       [this, running]() {
         if (*running) {
-          connection_->sendKeepalive();
-          schedule();
+          if (pending_) {
+            stop();
+
+            connection_->outputFrameOrEnqueue(
```
lehecka: I am not a big fan on spreading the behavior of failed keepalive over here.
Do you mind creating a new method ConnectionAutomaton::keepaliveFailed and putting the {outputFrameOrEnqueue, disconnect} logic there?

yschimke: Yep, will do.  I wonder whether it should be a terminateWithPrejudice method.  I'm never sure this error frame actually gets sent before termination, and its repeated inside ConnectionAutomaton also.

lehecka: the frame gets written out if we have the allowance, otherwise it is lost and connection closed.

lehecka:
to your questions in the description...
I like your current implementation. I wouldn't spend any more time trying to fine tune it now.
We can add al those things you are mentioning later.

yschimke:
Mocking ConnectionAutomaton seems problematic.  Any suggestions? Split out a sub interface (sendFrame/Keepalive + disconnectWithError)?

PR #186, opened 2016-11-23
Enforce setup frame

yschimke:


yschimke:
don't wait on this, I'll add tests before putting up for review.  I'll use this for testing internally as a patch.

yschimke:
@lehecka Ping me when you are online, I wanted to discuss how to unit test these without refactoring the StreamAutomatonFactory.  The logic is split awkwardly between ReactiveSocket and ConnectionAutomaton, and the ConnectionAutomaton tests don't use the real factory.

lehecka:
feel free to ping me when you see me online...
I think you can write the test such that you create two linked DuplexConnections. Then create ReactiveSocket instance from one of the DuplexConnections and send other than setup frame to the  RS via the DuplexConnection. You will expect error frame and closing of RS.

Another test can be about sending multiple setup frames The second setup should be ignored.

Look at test/ReactiveSocketConcurrencyTest.cpp line 444 thru 537. You can copy paste stuff from there and simplify it for this.

lehecka:
this code change drifted away. Enforcing the setup payload is going to be solved differently with a different state machine.

PR #187, opened 2016-11-23
Attempt Upgrade to gcc 4.9

benjchristensen:
Experimenting with Travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,15 +14,15 @@ matrix:
           packages:
             - g++-4.9
       env: COMPILER=g++-4.9 CXX=g++-4.9
-    - compiler: clang
-      addons:
-        apt:
-          sources:
-            - ubuntu-toolchain-r-test
-            - llvm-toolchain-precise-3.6
-          packages:
-            - clang-3.6
-      env: COMPILER=clang++-3.6 CXX=clang++-3.6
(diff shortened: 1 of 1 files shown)


benjchristensen:
@lehecka This seems to work for gcc4.9, can you take a look? Do you want clang as well (which isn't working)?

benjchristensen:
Seems like clang is just timing out ... and my attempt to extend the timeout failed :-/

yschimke:
We unfortunately need support for both clang and gcc, as both are used internally.

lehecka:
I know this is going to take a lot of tries to figure this out. Thank you for doing this!
I tried this and ended up convincing travis to use gcc 4.9, but the build was messed up enough that the tests froze and never completed: https://github.com/lehecka/reactivesocket-cpp/commit/17a592c7d4ed3ab11c97162325a633cdcdefc95d

lehecka:
the latest attempt from me is here: https://travis-ci.org/lehecka/reactivesocket-cpp/builds/178380361

it looks like it all compiles, but gets stuck in executing tests.
Maybe we can disable tests and see if they are all messed up or just one test is messed up..

lehecka:
closing in favor of https://github.com/ReactiveSocket/reactivesocket-cpp/pull/253

PR #191, opened 2016-11-25
cleanup imports

yschimke:
CLion "optimise imports"

What do people think?


jspahrsummers:
:+1: once it builds

lexs:
I'm a big fan of "include everything you use" but C++ is a terrible language for that (before we have modules) so ¯\_(ツ)_/¯.

PR #193, opened 2016-12-01
cleanup TCP resume client and server to generate UUIDs for tokens and…

tmontgomery:
… store StreamState by token. Modify clean/dirty callbacks to be more useful. Add status callback to tryClientResume.


src/ConnectionAutomaton.cpp
```diff
@@ -293,6 +310,17 @@ void ConnectionAutomaton::onConnectionFrame(
       }
       return;
     }
+    case FrameType::ERROR: {
+      Frame_ERROR frame;
+      if (frame.deserializeFrom(std::move(payload))) {
+        if (frame.errorCode_ == ErrorCode::CONNECTION_ERROR && !isServer_ && isResumable_ && resumeStatusCallback_) {
+          resumeStatusCallback_->onConnectionError(std::runtime_error(frame.payload_.moveDataToString()));
```
lehecka: I would check that we can set resumeStatusCallback_ only if !isServer_ && isResumable_ and then on this place I would only check the resumeStatusCallback_ pointer.

Also it feels that since you received a frame, that means the transport layer is probably working just fine.
We should probably just message Resume failure.

The underlaying DuplexConnection might terminate with cancel, onComplete or onError. That is a transport layer failure which we should hook up to and message it as network failure.

BTW, does onConnectionError mean resume error?

src/ConnectionAutomaton.cpp
```diff
@@ -98,9 +101,11 @@ void ConnectionAutomaton::closeDuplexConnection(folly::exception_wrapper ex) {
 }
 
 void ConnectionAutomaton::reconnect(
-    std::unique_ptr<DuplexConnection> newConnection) {
+  std::unique_ptr<DuplexConnection> newConnection,
+  std::unique_ptr<ClientResumeStatusCallback> statusCallback) {
   disconnect();
   connection_ = std::shared_ptr<DuplexConnection>(std::move(newConnection));
+  resumeStatusCallback_ = std::shared_ptr<ClientResumeStatusCallback>(std::move(statusCallback));
```
lehecka: you probably don't want to allow more than one resume operation if flight, so I would say you want to CHECK(!resumeStatusCallback)

src/ClientResumeStatusCallback.h
```diff
@@ -0,0 +1,24 @@
...
+public:
+
+  // Called when a RESUME_OK frame is received during resuming operation
+  virtual void onResumeOk() = 0;
+
+  // Called when an ERROR frame with CONNECTION_ERROR is received during resuming operation
+  virtual void onConnectionError(folly::exception_wrapper ex) = 0;
+};
+
+class ClientResumeStatusCallbackDefault : public ClientResumeStatusCallback {
```
lehecka: just a nit, but are you used to naming things in the opposite order? Or why not just DefaultClientResumeStatusCallback
(To be consistent, I would also prefer word Handler instead of Callback, because we have RequestHandler and not RequestCallback. But will leave it up to you)

src/ClientResumeStatusCallback.h
```diff
@@ -0,0 +1,24 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#pragma once
+
+#include <folly/ExceptionWrapper.h>
+
+namespace reactivesocket {
+
+class ClientResumeStatusCallback {
```
lehecka: class with virtual methods should have a virtual destructor. That is a lint rule in FB...

src/ConnectionAutomaton.cpp
```diff
@@ -262,14 +267,22 @@ void ConnectionAutomaton::onConnectionFrame(
...
+
+            if (streamState_->resumeCache_->isPositionAvailable(frame.position_, streamId)) {
+                requestHandler_->handleCleanResume(streamId, errorStream);
             } else {
-              it.second->onDirtyResume();
+                requestHandler_->handleDirtyResume(streamId, errorStream);
+            }
+
+            if (errorStream) {
+                it = streamState_->streams_.erase(it);
```
lehecka: I believe you should be sending endStream signal to the AbstractAutomaton to make sure it will do a proper cleanup of subscribers/subscriptions.

src/ConnectionAutomaton.cpp
lehecka: I would also not allow server RS instance to execute this method. Also not resumable RS should fail in this call.
(CHECK)

src/ClientResumeStatusCallback.h
```diff
@@ -0,0 +1,24 @@
...
+public:
+
+  // Called when a RESUME_OK frame is received during resuming operation
+  virtual void onResumeOk() = 0;
+
+  // Called when an ERROR frame with CONNECTION_ERROR is received during resuming operation
+  virtual void onConnectionError(folly::exception_wrapper ex) = 0;
```
lehecka: should the destructor automatically call onConnectionError if onResumeOk or onConnectionError were never called?

src/ConnectionAutomaton.cpp
```diff
@@ -293,6 +310,17 @@ void ConnectionAutomaton::onConnectionFrame(
       }
       return;
     }
+    case FrameType::ERROR: {
+      Frame_ERROR frame;
+      if (frame.deserializeFrom(std::move(payload))) {
+        if (frame.errorCode_ == ErrorCode::CONNECTION_ERROR && !isServer_ && isResumable_ && resumeStatusCallback_) {
+          resumeStatusCallback_->onConnectionError(std::runtime_error(frame.payload_.moveDataToString()));
+          resumeStatusCallback_.reset();
```
lehecka: how about the case when ReactiveSocket instance is released while the resumeStatusCallback is in flight? Maybe the disconnect method or dtor should fail the resumeStatusCallback_

src/RequestHandler.h
```diff
@@ -22,6 +23,25 @@ class SubscriberFactory {
       folly::Executor& executor) = 0;
 };
 
+// Object to hold state for erroring a stream during resume callbacks
+class ErrorStream {
```
lehecka: at this moment I don't see much purpose for this class. folly::exception_wrapper has also explicit operator bool(). What is this class providing opposite to folly::exception_wrapper?

src/mixins/MixinTerminator.h
```diff
@@ -39,26 +39,20 @@ class MixinTerminator {
     Parameters() = default;
     Parameters(
         const std::shared_ptr<ConnectionAutomaton>& _connection,
-        StreamId _streamId,
-        std::shared_ptr<RequestHandlerBase> _handler)
```
lehecka: like!

test/resume/TcpResumeClient.cpp
```diff
@@ -31,6 +32,39 @@ class Callback : public AsyncSocket::ConnectCallback {
...
+
+  void onConnectionError(folly::exception_wrapper ex) override {
+    std::cout << "resume connection error " << ex.what() << "\n";
+  }
+};
+
+class GuidGenerator {
+public:
+  // method to generate a UUID for the Resume Token
+  void static generateAndFill(ResumeIdentificationToken& token) {
```
lehecka: should we turn ResumeIdentificationToken into a class and integrate methods of GuidGenerator into it?

src/ConnectionAutomaton.cpp
tmontgomery: Agreed. Wasn't sure what the correct thing to do here was.

src/ClientResumeStatusCallback.h
tmontgomery: ok

src/ConnectionAutomaton.cpp
tmontgomery: Same with a client in the middle of SETUP?

src/ClientResumeStatusCallback.h
tmontgomery: it could. But that would seem odd since the application decided to have the RS go out of scope.

src/ClientResumeStatusCallback.h
tmontgomery: Naming is hard. I'm open to using Handler and can pre-pend Default for consistency.

test/resume/TcpResumeClient.cpp
tmontgomery: Probably. Also, if we make the token variable length, it should be an object.

src/RequestHandler.h
tmontgomery: No need for exception_wrapper. Just need a simple boolean.

tmontgomery:
I left out calling resume callback on RS dtor as well as Resume Token change to a class. Should discuss those more, IMO.

lehecka:
can I close this PR as outdated?

lehecka:
closing in favor of https://github.com/ReactiveSocket/reactivesocket-cpp/pull/246
where the rest of the cleanup code got ported to.

PR #194, opened 2016-12-01
pull in change to leak the FrameAllocator

yschimke:
Leak the FrameAllocator to avoid shutdown destructor races

cc @lehecka @jspahrsummers

### DIFF
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -6,12 +6,7 @@
 
 #include <folly/Memory.h>
 #include <folly/Optional.h>
-#include <folly/Singleton.h>
 #include <folly/io/Cursor.h>
-
-namespace {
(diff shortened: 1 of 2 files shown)


lehecka:
I don't think this is a good change.
Why would you generate frames when the application is being teared down? If there is a timing issue in shutting down threads, it should be resolved somewhere else.

Lets open this issues and discuss it better. I feel there should be a better fix for the issue than this.

lehecka:
Not sure why ASAN doesn't complain here, but we are essentially making memory leaking detection very hard since we are leaking by design.
I would like to find a different fix for the original problem.
@yschimke can you share some details about the original problem?

jspahrsummers:
Will follow up on chat.

yschimke:
I wanted to keep the two repos, in sync.  no objection to a more appropriate change.

PR #197, opened 2016-12-01
Fix PublisherMixin requiring onSubscribe() to be called synchronously

martinsix:
Right now, PublisherMixin swallows the initial request size if onSubscribe() is not called synchronously before reaching https://github.com/ReactiveSocket/reactivesocket-cpp/blob/master/src/ReactiveSocket.cpp#L335

We need to store the initial request n in this case, and only request after onSubscribe() was called

### DIFF
--- a/src/mixins/PublisherMixin.h
+++ b/src/mixins/PublisherMixin.h
@@ -27,6 +27,9 @@ class PublisherMixin : public Base {
   void onSubscribe(std::shared_ptr<Subscription> subscription) {
     DCHECK(!producingSubscription_);
     producingSubscription_.reset(std::move(subscription));
+    if (initialRequestN_) {
+      producingSubscription_.request(initialRequestN_);
+    }
   }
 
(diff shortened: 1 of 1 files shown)


src/mixins/PublisherMixin.h
```diff
@@ -67,12 +70,14 @@ class PublisherMixin : public Base {
...
-      if (size_t n = frame.requestN_) {
+    if (size_t n = frame.requestN_) {
+      if (producingSubscription_) {
         producingSubscription_.request(n);
+      } else {
+        initialRequestN_ += n;
       }
-      Base::onNextFrame(std::move(frame));
     }
+    Base::onNextFrame(std::move(frame));
```
martinsix: PublisherMixin is the last mixin in the chain, so this doesn't really matter, but IMO it's more correct to always call Base::onNextFrame(), regardless of whether we have a subscription here or not

lehecka:
nice work. Thank you Martin!

PR #200, opened 2016-12-06
Unit test

lehecka:
This is adding unit tests for for lazy subscription to server streams.
I also added debug logging (VLOG) for the async signals queuing and executing. It helps with debugging issues when the queued signals are not delivered.

clang-formatting got in the way...


lehecka:
since this is not adding or changing functionality, I am going to merge the change right away.

PR #204, opened 2016-12-14
unexpected frames will error out ReactiveSocket

lehecka:
the expected frames for each automaton are explicitly declared as overriden onNextFrame methods. The default implementation (for the not overriden cases) in the StreamAutomatonBase will take down the ReactiveSocket with the ERROR frame as prescribed by the protocol definition.


src/automata/RequestResponseResponder.cpp
```diff
@@ -76,6 +76,10 @@ void RequestResponseResponder::onNextFrame(Frame_CANCEL&& frame) {
   }
 }
 
+void RequestResponseResponder::onNextFrame(Frame_REQUEST_RESPONSE&& frame) {
+  processRequestN(frame);
```
yschimke: I assume this is correct, just looks counter-intuitive.

src/automata/StreamAutomatonBase.cpp
```diff
@@ -0,0 +1,54 @@
...
+}
+
+void StreamAutomatonBase::onBadFrame() {
+  connection_->disconnectWithError(
+      Frame_ERROR::invalid(streamId_, "bad frame"));
+}
+
+void StreamAutomatonBase::onUnexpectedFrame() {
+  DCHECK(false) << "onUnexpectedFrame";
+  connection_->disconnectWithError(Frame_ERROR::unexpectedFrame());
```
yschimke: How does this jive with?

> Handling the Unexpected
> 
> This protocol attempts to be very lenient in processing of received frames and SHOULD ignore conditions that do not make sense given the current context.

src/mixins/PublisherMixin.h
```diff
@@ -93,27 +98,17 @@ class PublisherMixin : public Base {
         initialRequestN_ += n;
       }
     }
-    Base::onNextFrame(std::move(frame));
   }
 
-  void onNextFrame(Frame_REQUEST_RESPONSE&& frame) {
+  void processRequestN(const Frame_REQUEST_RESPONSE&) {
```
yschimke: makes sense, naming just a bit confusing. TIOLI

yschimke:
LGTM - a couple of questions and shipit when you are ready.  still interested in the answers though

src/automata/StreamAutomatonBase.cpp
lehecka: I will ask Ben what these words mean exactly. We can ignore the frames, but it feels to me it can exploit some unwanted behavior as the buggy implementation can execute some piece of code which doesn't make sense in the context.

PR #206, opened 2016-12-15
implementing disconnecting of ReactiveSocket

lehecka:
this diff allows a new state of ReactiveSocket instance when the instance is disconnected from the network and the streams stay alive.

new unit test included

### DIFF
--- a/src/Common.cpp
+++ b/src/Common.cpp
@@ -18,7 +18,7 @@ static const char* getTerminatingSignalErrorMessage(int terminatingSignal) {
       return "reactive socket closed";
     case StreamCompletionSignal::GRACEFUL:
       DCHECK(false) << "throwing exception for GRACEFUL termination?";
-      return "gracefull termination";
+      return "graceful termination";
     default:
       return "stream interrupted";
   }
--- a/src/Common.cpp
+++ b/src/Common.cpp
@@ -14,12 +14,37 @@ static const char* getTerminatingSignalErrorMessage(int terminatingSignal) {
       return "connection error";
     case StreamCompletionSignal::ERROR:
       return "general error";
+    case StreamCompletionSignal::SOCKET_CLOSED:
+      return "reactive socket closed";
     case StreamCompletionSignal::GRACEFUL:
       DCHECK(false) << "throwing exception for GRACEFUL termination?";
       return "gracefull termination";
     default:
       return "stream interrupted";
   }
+}
+
+std::ostream& operator<<(std::ostream& os, StreamCompletionSignal signal) {
+  switch (signal) {
+    case StreamCompletionSignal::GRACEFUL:
+      return os << "GRACEFUL";
+    case StreamCompletionSignal::ERROR:
+      return os << "ERROR";
+    case StreamCompletionSignal::INVALID_SETUP:
+      return os << "INVALID_SETUP";
+    case StreamCompletionSignal::UNSUPPORTED_SETUP:
+      return os << "UNSUPPORTED_SETUP";
+    case StreamCompletionSignal::REJECTED_SETUP:
+      return os << "REJECTED_SETUP";
+    case StreamCompletionSignal::CONNECTION_ERROR:
+      return os << "CONNECTION_ERROR";
+    case StreamCompletionSignal::CONNECTION_END:
+      return os << "CONNECTION_END";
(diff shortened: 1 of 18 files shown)


src/Common.cpp
```diff
@@ -14,6 +14,8 @@ static const char* getTerminatingSignalErrorMessage(int terminatingSignal) {
       return "connection error";
     case StreamCompletionSignal::ERROR:
       return "general error";
+    case StreamCompletionSignal::SOCKET_CLOSED:
+      return "reactive socket closed";
     case StreamCompletionSignal::GRACEFUL:
       DCHECK(false) << "throwing exception for GRACEFUL termination?";
       return "gracefull termination";
```
yschimke: typo

src/Common.cpp
```diff
@@ -22,6 +24,29 @@ static const char* getTerminatingSignalErrorMessage(int terminatingSignal) {
...
+      return os << "REJECTED_SETUP";
+    case StreamCompletionSignal::CONNECTION_ERROR:
+      return os << "CONNECTION_ERROR";
+    case StreamCompletionSignal::CONNECTION_END:
+      return os << "CONNECTION_END";
+    case StreamCompletionSignal::SOCKET_CLOSED:
+      return os << "SOCKET_CLOSED";
+  }
+  // this should be never hit because the switch is over all cases
+  LOG(FATAL) << "unknown StreamCompletionSignal=" << (int)signal;
```
yschimke: Are these meant to be https://github.com/ReactiveSocket/reactivesocket/blob/master/Protocol.md#error-codes or just similarly named but internal only?

yschimke: Actually I guess that is still ErrorCode

src/ConnectionAutomaton.cpp
```diff
@@ -41,19 +42,25 @@ ConnectionAutomaton::ConnectionAutomaton(
...
-  // signal
-  // which would call disconnect
+  // signal which would call disconnect/close
   if (connection_) {
     // This may call ::onSubscribe in-line, which calls ::request on the
     // provided
     // subscription, which might deliver frames in-line.
-    // it can also call onComplete which will call disconnect() and reset the
+    // it can also call onComplete which will call disconnect/close and reset
+    // the
```
yschimke: nit: weird line breaks

src/ConnectionAutomaton.cpp
```diff
@@ -109,10 +121,14 @@ void ConnectionAutomaton::closeDuplexConnection(folly::exception_wrapper ex) {
...
+}
 
-  stats_.socketClosed();
-  onClosed_();
+void ConnectionAutomaton::closeWithError(Frame_ERROR&& error) {
+  VLOG(4) << "closeWithError "
+          << error.payload_.data->cloneAsValue().moveToFbString();
+
+  outputFrameOrEnqueue(error.serializeOut());
+  close(folly::exception_wrapper(), StreamCompletionSignal::ERROR);
```
yschimke: folly::exception_wrapper() seems like it can be interpreted multiple ways

https://github.com/facebook/folly/blob/master/folly/ExceptionWrapper.h#L72

```
 * // Thread1
 * void doSomethingCrazy() {
 *   int rc = doSomethingCrazyWithLameReturnCodes();
 *   if (rc == NAILED_IT) {
 *     globalExceptionWrapper = exception_wrapper();
 *   } else if (rc == FACE_PLANT) {
 *     globalExceptionWrapper = make_exception_wrapper<FacePlantException>();
```

https://github.com/facebook/folly/blob/master/folly/ExceptionWrapper.h#L155

yschimke: Maybe everyone else apart from me was already clear that this is effectively NO_ERROR, instead of UNDEFINED_ERROR

yschimke:
Nothing blocking from me, a few questions.

src/Common.cpp
lehecka: This is internal enum we use to send down when closing the streams. Based on this value every stream then calls methods like onComplete/onError, etc.
The protocol error codes are defined in Frame.h, enum ErrorCode.

src/ConnectionAutomaton.cpp
lehecka: Yeah, we have been using exception_wrapper() (default ctor) to indicate no error. This could be found on multiple places in our codebase and also in other FB code.

PR #208, opened 2016-12-16
moving AllowanceSemaphore and SmartPointers from reactive-streams-cpp to reactivesocket-cpp

lehecka:


lehecka:
since this is only copy paste of files and fixing include path I am going to merge without reviewing

PR #210, opened 2016-12-16
Fix FramedWriter tests' use of stack allocated IOBuf

alexmalyshev:
These tests prepend two nonempty IOBufs onto an empty, default constructed
IOBuf, and crash under UBSAN when trying to coalesce the chain.

From my reading, the only thing that's legal to do with a default constructed
IOBuf is to move another IOBuf into it, none of the coalescing code can actually
handle a (nullptr, nullptr) IOBuf.


lehecka:
thanks for the fix!

PR #211, opened 2016-12-20
Expose DuplexConnection

martinsix:
We'd like to be able to access the underlying duplexconnection after
creating the reactivesocket, because the connection may contain
properties that we'd like to use. An example would be a HTTP header
that we don't want to store separately, when it's already stored with
the connection.


src/ReactiveSocket.h
```diff
@@ -107,6 +107,8 @@ class ReactiveSocket {
       std::unique_ptr<DuplexConnection> newConnection,
       const ResumeIdentificationToken& token);
 
+  const DuplexConnection& getDuplexConnection();
```
lehecka: please change this to pointer instead of a reference return type. When the RS instance is disconnected it doesn't have DuplexConnection associated.

lehecka: other that that, LGTM, feel free to merge.

martinsix:
I just realized that the connectionAutomaton doesn't store the duplexconnection any more. Is there a better way to expose connection properties on the socket?

lehecka:
ConnectionAutomaton is holding a reference to FrameTransport and FrameTransport is holding the reference to DuplexConnection. You can still expose the DuplexConnection _pointer_ out of RS class.

Please remember, that accessing DuplexConnection should be done only on the same executor (thread/EventBase). I would also suggest adding this comment in the header file.

lehecka:
I will take care of this PR.

lehecka:
closing in favor of https://github.com/ReactiveSocket/reactivesocket-cpp/pull/224

PR #212, opened 2016-12-20
Automatically cancel Subscription in ~SubscriberBase

jspahrsummers:
I expected this to already be the case while testing some behaviors built on top of RS, and was surprised to find out that it's not. I'm not sure if we ever saw any bad behavior without this, but it seems safe and like the proper thing to do.

### DIFF
--- a/src/SubscriberBase.h
+++ b/src/SubscriberBase.h
@@ -83,6 +83,14 @@ class SubscriberBaseT : public Subscriber<T>,
       bool startExecutor = true)
       : ExecutorBase(executor, startExecutor), cancelled_(false) {}
 
+  virtual ~SubscriberBaseT() {
+    if (!cancelled_ && originalSubscription_) {
+      runInExecutor([subscription = std::move(originalSubscription_)] {
+        subscription->cancel();
+      });
+    }
+  }
+
   void onSubscribe(std::shared_ptr<Subscription> subscription) override final {
     auto thisPtr = this->shared_from_this();
     runInExecutor([thisPtr, subscription]() {
--- a/src/SubscriberBase.h
+++ b/src/SubscriberBase.h
@@ -85,7 +85,8 @@ class SubscriberBaseT : public Subscriber<T>,
 
   virtual ~SubscriberBaseT() {
     if (!cancelled_ && originalSubscription_) {
-      runInExecutor([subscription = std::move(originalSubscription_)] {
+      auto subscription = std::move(originalSubscription_);
+      runInExecutor([subscription] {
         subscription->cancel();
       });
     }
--- a/test/SubscriberBaseTest.cpp
+++ b/test/SubscriberBaseTest.cpp
@@ -159,3 +159,13 @@ TEST(SubscriberBaseTest, SubscriptionCancel) {
   std::shared_ptr<Subscriber<int>> ptr = subscriber;
   ptr->onSubscribe(originalSubscription);
 }
+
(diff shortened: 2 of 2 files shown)


src/SubscriberBase.h
```diff
@@ -83,6 +83,15 @@ class SubscriberBaseT : public Subscriber<T>,
       bool startExecutor = true)
       : ExecutorBase(executor, startExecutor), cancelled_(false) {}
 
+  virtual ~SubscriberBaseT() {
+    if (!cancelled_ && originalSubscription_) {
```
lehecka: At first I thought the destructor is fine, but as I am thinking about how did you get into that I see a few issues.

If you called onSubscribe then you have to call a terminating on{Complete,Error}. If you didn't that is a broken contract and a bug. If you would call the terminating signal, you wouldn't have a valid pointer in the destructor.

From the threading correctness I am not sure the code is thread safe. We know that reading the pointer value or modifying the pointer is not thread safe. The destructor can be executed on any thread, but originalSubscription_ can be safely modified only in runInExecutor (like in the other methods). Since there could be 2 threads modifying the pointer close enough in time and there is no memory barrier (or lock) you might be reading stale value from processor cache. The only safe way to do this automatic cancellation is passing thisPtr inside the lambda, which makes no sense since you are in destructor and the instance is being destroyed.
I can't help it, but there is probably no safe way of doing it.

I would suggest that the only safe thing we can do in the destructor is (D)CHECK(! originalSubscription_); to make verify the user terminated the subscription..

What do you think?

jspahrsummers: Let me back up and describe my problem, and then maybe we can approach this a different way.

I have a class inheriting `SubscriberBaseT` which receives a `Subscription`. In the event that the _subscriber_ is destroyed, its subscription should cancel—otherwise, it might continue running, right? The problem is, `SubscriptionBase` is not cancellable when the `SubscriberBaseT` is being destructed, because it attempts to lock its `weak_ptr`, and it's already too late to do that.

Maybe an alternative way to solve this would be to make `SubscriptionBase` cancellable even if the parent `SubscriberBaseT` is no longer alive, but I'm not sure how best to achieve that.

What do you think?

lehecka: When you say the SubscriberBaseT is about to be destroyed (because nobody is using it anymore) that should mean somebody sent a terminating signal to that instance prior releasing the shared_ptr. And if I am looking correctly at the code, if you call any of the {cancel, onComlete, onError} then the subscription is always going to be cancelled.
If the Subscription is not cancelled then the code never sent the terminating signal. And based on the contract, this is a bug.
I agree that if you lost the SubscriberBaseT instance already, nobody will be able to call cancel on the subscription because the executor is also gone.

jspahrsummers: Where is it guaranteed that the `SubscriberBaseT` will be strongly retained until it receives a terminating signal? `SubscriptionBase` doesn't guarantee this, for example.

lehecka: Not sure if I understand the question well. RS hold shared_pr to the Subscriber and makes sure it will call on{Error,Complete}. The implementation of that method will schedule on{Error,Complete}Impl to be executed on Executor and will capture thisPtr, a hard shared_ptr reference, in the lambda parameter.
Am I answering your question?

jspahrsummers: Okay, fair enough. I was doing this in unit tests, where I have a `Subscriber` which is never passed to a `ReactiveSocket`—in those circumstances, it's possible for a `Subscription` to go uncancelled, but I guess that's just an artifact of my test and not something that will happen in prod?

PR #214, opened 2016-12-21
Separate

lehecka:
Splitting implementation of Subscriber and Subscription in ConnectionAutomaton into a new separate class FrameTransport.
Thanks to this we will be able to disconnect and reconnect new connections (represented by FrameTransport). This is a necessary refactoring diff for the following diff with resumability bits.


src/ReactiveSocket.cpp
```diff
@@ -203,40 +197,35 @@ bool ReactiveSocket::createResponder(
...
-        //            frame").serializeOut());
-        //        disconnect();
+      if (!connection.deserializeFrameOrError(
+              frame, std::move(serializedFrame))) {
+        return;
       }
+      if (frame.header_.flags_ & FrameFlags_LEASE) {
+        // TODO(yschimke) We don't have the correct lease and wait logic above
+        // yet
+        LOG(WARNING) << "ignoring setup frame with lease";
```
yschimke: should we change this now?

The spec is pretty clear:

`If the server does NOT accept the contents of the SETUP frame, the server MUST send back a SETUP_ERROR and then close the connection.`

But reactivesocket-java used leases, so it was useful to ignore.

yschimke:
The change looks really nice, ship when you are happy with it

src/ReactiveSocket.cpp
yschimke: n.b. this can obviously wait, but you are changing this code anyway and fix should be pretty safe and not affect us.  Up to you.

src/ConnectionAutomaton.cpp
```diff
@@ -50,40 +48,41 @@ ConnectionAutomaton::~ConnectionAutomaton() {
...
+  // the connection may also get closed before the method returns so
+  // we need to check again
+  auto outputFrames = streamState_->moveOutputFrames();
+  if (frameTransport_) {
+    for (auto& frame : outputFrames) {
+      outputFrame(std::move(frame));
+    }
     onConnected_();
+  } else {
+    LOG_IF(WARNING, !outputFrames.empty()) << "transport closed, throwing away "
```
lehecka: I would ideally use LOG_IF_EVERY_MS, but that is fbcode local macro. Any suggestions what should I use here?

src/ReactiveSocket.cpp
lehecka: lets change this in another PR. This one is quite large already.

src/ConnectionAutomaton.cpp
yschimke: I mainly don't think you care in a server that is processing a huge volume of traffic with asynchronous network events and payload deliveries.  it's very inevitable and uninteresting.

VLOG?

PR #215, opened 2016-12-21
Resume complete

lehecka:
This is a complete implementation resumability for intermittent disconnections.

Also there is a new constructor for Reactive socket:
auto socket = ReactiveSocket::disconnectedClient(...);
auto socket = ReactiveSocket::disconnectedServer(...);

which allows clients to start creating streams (requestXXX)

and then connecting the socket via connectServer, connectClient.

A few of the more granular signals are commented out since they are not used at the moment.

The change has been tested on the tcpresume{client,server}.


yschimke:
Does tcpresumeclient talk to normal tcpserver and vice/versa?

lehecka:
tcpresumeclient typically talks to tcpresumeserver and they exchange expected messages. It would connect to tcpserver, but it is not gonna exchange the expected messages.

yschimke:
Does that indicate a need to carefully update both client and server at the same time. Mostly interested that they don't cause segfaults

lehecka:
oh, I see what you mean. 
I didn't break the compatibility so we don't need to upgrade both servers and clients at the same time.

PR #218, opened 2017-01-04
Executor in ReactiveSocket instance

lehecka:
This is the first diff of a series to complete the following:
* ReactiveSocket class needs to operate on its own Executor. It can no longer "inherit" the executor from the underlaying DuplexConnection. The resume operation replaces the old DuplexConnection with a new DuplexConnection which might be operating on a different thread (executor). Switching executors is not possible because of all possible threading issues.
* In this diff, the application code provides Executor to the constructor of ReactiveSocket. We can then remove the executor parameter from all the requestXXX methods.
* In the next diff I want to make all public methods of ReactiveSocket (including dtor) to be thread safe with respect to the executor. To be concrete, if you provide EventBase as an executor, the instance will be thread safe. This will allow us to remove the rest of executeInEventBase([]{ socket->bla() }) workarounds.
* At the end the resume operation will be possible even for DuplexConnection from a different thread.


src/ConnectionAutomaton.cpp
```diff
@@ -60,8 +60,7 @@ void ConnectionAutomaton::connect(
     std::shared_ptr<FrameTransport> frameTransport,
     bool sendingPendingFrames) {
   CHECK(isDisconnectedOrClosed());
-  CHECK(frameTransport);
-  CHECK(!frameTransport->isClosed());
+  CHECK(frameTransport && !frameTransport->isClosed());
```
jspahrsummers: There's an advantage to leaving this separate—we'll get better crash reports if the condition is ever violated.

src/Executor.cpp
```diff
@@ -2,17 +2,19 @@
...
 #include <folly/futures/QueuedImmediateExecutor.h>
+#include <folly/io/IOBuf.h>
 #include "src/StackTraceUtils.h"
 #include "src/SubscriberBase.h"
 
 namespace reactivesocket {
 
 // just instantiating of the template here
 template class SubscriberBaseT<Payload>;
+template class SubscriberBaseT<folly::IOBuf>;
```
jspahrsummers: I might've asked this before (sorry if so), but why are you instantiating them here, again?

src/ReactiveSocket.cpp
```diff
@@ -35,14 +35,15 @@ ReactiveSocket::ReactiveSocket(
     bool isServer,
     std::shared_ptr<RequestHandlerBase> handler,
     Stats& stats,
-    std::unique_ptr<KeepaliveTimer> keepaliveTimer)
+    std::unique_ptr<KeepaliveTimer> keepaliveTimer,
+    folly::Executor& executor)
     : handler_(handler),
       connection_(std::make_shared<ConnectionAutomaton>(
-          [handler](
+          [this, handler](
```
jspahrsummers: Is it guaranteed that the `ConnectionAutomaton` will be destroyed before the `ReactiveSocket`? Since they're both `shared_ptr`s, it seems possible that the automaton could outlive the socket in some pathological cases.

src/ReactiveSocket.cpp
```diff
@@ -241,15 +243,13 @@ void ReactiveSocket::createResponder(
...
-      ReactiveSocketSubscriberFactory subscriberFactory(
-          [&](folly::Executor* executor) {
-            ChannelResponder::Parameters params = {
-                {connection.shared_from_this(), streamId, handler},
-                executor ? *executor : defaultExecutor()};
-            automaton = std::make_shared<ChannelResponder>(params);
-            connection.addStream(streamId, automaton);
-            return automaton;
-          });
+      ReactiveSocketSubscriberFactory subscriberFactory([&](folly::Executor*) {
```
jspahrsummers: Can the parameter here be removed entirely, from the type?

src/ReactiveSocket.h
```diff
@@ -51,45 +51,47 @@ class ReactiveSocket {
       ConnectionSetupPayload setupPayload = ConnectionSetupPayload(),
       Stats& stats = Stats::noop(),
       std::unique_ptr<KeepaliveTimer> keepaliveTimer =
-          std::unique_ptr<KeepaliveTimer>(nullptr));
+          std::unique_ptr<KeepaliveTimer>(nullptr),
+      folly::Executor& executor = defaultExecutor());
```
jspahrsummers: I don't think this argument should be defaulted. In our mobile code, won't we always want to use Liger's `EventBase` here? It seems like something that the caller should always carefully consider.

src/ReactiveSocket.h
```diff
@@ -106,25 +108,28 @@ class ReactiveSocket {
...
       const ResumeIdentificationToken& token,
-      std::unique_ptr<DuplexConnection> newConnection,
+      std::shared_ptr<FrameTransport> frameTransport,
       std::unique_ptr<ClientResumeStatusCallback> resumeCallback);
 
   bool tryResumeServer(
       std::shared_ptr<FrameTransport> frameTransport,
       ResumePosition position);
 
+  folly::Executor& executor() {
```
jspahrsummers: This method can be `const`.

src/Executor.h
```diff
@@ -25,6 +25,9 @@ class ExecutorBase {
 
  protected:
   void runInExecutor(folly::Func func);
+  folly::Executor& executor() {
```
jspahrsummers: This method can be `const`.

src/ConnectionAutomaton.cpp
lehecka: the check violation should be caught during testing and development. I don't expect this to be ever hit in the production so my thinking was having one check statement generates less code and less overhead.

src/Executor.cpp
lehecka: The template is declared as extern in the header file, to prevent instantiation of the template in every compilation unit. It has to be instantiated once somewhere. I can add a new cpp file to instantiate the template there, but instead I added it here. If you have a preference where to do it, please let me know.

src/ReactiveSocket.cpp
lehecka: Theoretically it is possible that ConnectionAutomaton will outlive ReactiveSocket, but even in that case ConnectionAutomaton will never call back into this method. 
Either way I am changing this in the next diff, because I also dont like "this" to be captured here. So the problem will go away and in the meantime it will not cause any harm.

src/ReactiveSocket.cpp
lehecka: this Factory thing will all go away, including the Executor argument. I didn't want to cause too much distracting code churn in this diff.

src/ReactiveSocket.h
lehecka: then transitively the returned value will be const Executor& (because it is reference member variable) which won't allow you to call methods in the executor.

src/Executor.h
lehecka: same comment here.

src/ConnectionAutomaton.cpp
jspahrsummers: I don't think overhead is a real concern here, and the amount of code should be infinitesimal. Since these are the kinds of lines that show up in crash logs, I really think it's important to have as precise a root cause as possible—a combined binary condition like this one means that we'll have two possible causes instead of one, and so makes it harder to debug.

src/Executor.cpp
jspahrsummers: That sounds fine, I guess. It just seems a bit inconvenient that the template will have to be explicitly instantiated for each new parameter type. Oh well.

lehecka: the explicit instantiation is related only to SubscriberBaseT<Payload> and SubscriberBaseT<folly::IOBuf>. For any other template param type you would get the normal instantiation

src/ConnectionAutomaton.cpp
lehecka: will address in the next clenaup diff

PR #220, opened 2017-01-05
ReactiveSocket interface

yschimke:
Clients wanting to mock ReactiveSocket face a tough ask without an interface.  While technically possible its not as elegant and the other option is mocking DuplexConnection or using Inline connection, both of which are much lower level abstractions.

### DIFF
--- a/src/ReactiveSocket.h
+++ b/src/ReactiveSocket.h
@@ -13,149 +13,62 @@
 #include "src/Stats.h"
 
 namespace folly {
-class Executor;
+  class Executor;
 }
 
 namespace reactivesocket {
+  class ClientResumeStatusCallback;
+  class ConnectionAutomaton;
+  class DuplexConnection;
+  class FrameTransport;
 
-class ClientResumeStatusCallback;
-class ConnectionAutomaton;
-class DuplexConnection;
-class FrameTransport;
-class RequestHandler;
-class ReactiveSocket;
+  class ReactiveSocket {
+  public:
+    virtual ~ReactiveSocket() = default;
 
-folly::Executor& defaultExecutor();
+    virtual void requestStream(
+        Payload payload,
+        const std::shared_ptr<Subscriber<Payload>>& responseSink) = 0;
 
-// TODO(stupaq): Here is some heavy problem with the recursion on shutdown.
-// Giving someone ownership over this object would probably lead to a deadlock
-// (or a crash) if one calls ::~ReactiveSocket from a terminal signal handler
-// of any of the connections. It seems natural to follow ReactiveStreams
-// specification and make this object "self owned", so that we don't need to
-// perform all of the cleanup from a d'tor. We could give out a "proxy" object
-// that (internally) holds a weak reference (conceptually, not std::weak_ptr)
-// and forbids any interactions with the socket after the shutdown procedure has
-// been initiated.
-class ReactiveSocket {
- public:
-  ReactiveSocket(ReactiveSocket&&) = delete;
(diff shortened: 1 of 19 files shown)


src/ReactiveSocket.h
```diff
@@ -69,53 +57,53 @@ class ReactiveSocket {
       std::unique_ptr<RequestHandlerBase> handler,
       Stats& stats = Stats::noop());
 
-  std::shared_ptr<Subscriber<Payload>> requestChannel(
+  virtual std::shared_ptr<Subscriber<Payload>> requestChannel(
```
yschimke: I'll remove the virtuals as override is already there

lexs:
It's problematic because clients (eg ConnectionManager) will call `fromClientConnection()` et all which are static methods. I'm not sure how you'd make that mockable?

yschimke:
@lexs my full diff has a factory which actually works nicely and allows more flexibility, it also removes concerns like the KeepAliveTimer from the ConnectionManager

class ReactiveSocketFactory {
 public:
  virtual ~ReactiveSocketFactory() = default;
  virtual std::unique_ptr<::reactivesocket::ReactiveSocketInterface> create(
      std::unique_ptr<::reactivesocket::DuplexConnection> connection,
      ::reactivesocket::ResumeIdentificationToken resumeToken) = 0;
};

yschimke:
But generally even assuming that there are issues for some callers, there will be some code that is given a ReactiveSocket and should be able to test against that interface.

This stuff is trivial in java land e.g. https://github.com/ReactiveSocket/reactivesocket-java/blob/7f3cf71fcd616442a68e151da6629a216632ce13/reactivesocket-core/src/test/java/io/reactivesocket/test/util/MockReactiveSocket.java

or Mockito

lehecka:
Similarly to Java implementation, I would suggest to make all public methods of ReactiveSocket class virtual so that you can create mocks and override what you want. I don't see much value in having the extra interface. Sending ReactiveSocketInterface as parameter everywhere instead of ReactiveSocket doesn't look very pretty ;(
If you see a value in having the interface and if you plan on using it on places where the virtual methods would not be enough, then feel free to check it in.

yschimke:
@lehecka I guess that's really my point, that's what I want.  The Java version has

`public interface ReactiveSocket {`

and impls like

```
/**
 * Server side ReactiveSocket. Receives {@link Frame}s from a
 * {@link ClientReactiveSocket}
 */
public class ServerReactiveSocket implements ReactiveSocket {

/**
 * Client Side of a ReactiveSocket socket. Sends {@link Frame}s
 * to a {@link ServerReactiveSocket}
 */
public class ClientReactiveSocket implements ReactiveSocket {
```

Plus LoadBalancer and BackupRequestSocket.

So if we want to match the Java I would drop the Interface suffix, and call the implementation class something else, e.g. StandardReactiveSocket.  I'm happy to make this change.  Would you prefer that?  I'd like to get this right.

lehecka:
That's not a bad idea. I don't see much benefit to that, but I also don't see any harm to it. We aren't at the feature parity with Java yet so I am not sure what different implementations of the interface we will have, but maybe one day we will have more than one implementation. I will leave the decision up to you and will accept you diffs.

src/ReactiveSocket.h
```diff
@@ -13,149 +13,65 @@
 #include "src/Stats.h"
 
 namespace folly {
-class Executor;
+  class Executor;
 }
 
 namespace reactivesocket {
+  class ClientResumeStatusCallback;
```
lehecka: just currious, did you run ./devtools/format_all.sh?

PR #221, opened 2017-01-05
Make ResumeIdentificationToken::toString() return hex representation

lexs:
This allows me to print this in a nice way :)

Tested by creating:
```
ResumeIdentificationToken({0xab, 0xba, 0xff, 0x12, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0});
```

And verify it prints as:
```
ConnectionManager.cpp  I  Trying to resume socket, token: abbaff12000000000000000000000000
```

### DIFF
--- a/src/Common.cpp
+++ b/src/Common.cpp
@@ -7,6 +7,12 @@
 
 namespace reactivesocket {
 
+namespace {
+
+const char* HEX_CHARS = { "0123456789abcdef" };
+
+}
+
 static const char* getTerminatingSignalErrorMessage(int terminatingSignal) {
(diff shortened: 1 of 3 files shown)


src/Common.cpp
```diff
@@ -80,11 +86,12 @@ ResumeIdentificationToken ResumeIdentificationToken::fromString(
 }
 
 std::string ResumeIdentificationToken::toString() const {
```
yschimke: Can you add a test?

lexs: Added a test, rather not make this final for now in case @lehecka has plans. I just wanted it printable.

PR #223, opened 2017-01-06
Allocate StreamState::resume{Cache_,Tracker_} inline

alexmalyshev:
No reason for these to be unique_ptr.

### DIFF
--- a/src/StreamState.h
+++ b/src/StreamState.h
@@ -21,15 +21,13 @@ using StreamId = uint32_t;
 
 class StreamState {
  public:
-  StreamState()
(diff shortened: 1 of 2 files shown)


src/StreamState.h
```diff
@@ -21,15 +21,13 @@ using StreamId = uint32_t;
 
 class StreamState {
  public:
-  StreamState()
-      : resumeTracker_(new ResumeTracker()), resumeCache_(new ResumeCache()) {}
-
+  StreamState() = default;
```
lehecka: Please remove the ctor and dtor completely. The compiler will generate them automatically.

PR #227, opened 2017-01-10
fixing degenerated onNext method in Requester automatons

lehecka:
the onNext method name was very confusing, colliding with ReactiveStreams onNext.
It was also historically accidently implemented as async method (scheduling a callback on the executor).
This PR fixes that problem and removed the confusion in the situation when we are processing the initial payload.


CMakeLists.txt
```diff
@@ -203,7 +203,8 @@ add_executable(
   test/ReactiveSocketResumabilityTest.cpp
   test/SmartPointersTest.cpp
   test/AllowanceSemaphoreTest.cpp
-  test/ResumeIdentificationTokenTest.cpp)
+  test/ResumeIdentificationTokenTest.cpp
+  test/PayloadTest.cpp.cpp)
```
yschimke: too many .cpp

src/automata/RequestResponseRequester.cpp
```diff
@@ -110,31 +104,35 @@ void RequestResponseRequester::onNextFrame(Frame_ERROR&& frame) {
...
-      //       after receiving it. Add unit tests for this.
-      if (frame.header_.flags_ & FrameFlags_COMPLETE) {
-        state_ = State::CLOSED;
-        end = true;
-      }
+      state_ = State::CLOSED;
       break;
     case State::CLOSED:
+      // should not be receiving frames when closed
+      // if we ended up here, we broke some internal invariant fo the class
```
yschimke: typo: fo

src/automata/RequestResponseRequester.cpp
```diff
@@ -110,31 +104,35 @@ void RequestResponseRequester::onNextFrame(Frame_ERROR&& frame) {
 }
 
 void RequestResponseRequester::onNextFrame(Frame_RESPONSE&& frame) {
-  bool end = false;
   switch (state_) {
     case State::NEW:
       // Cannot receive a frame before sending the initial request.
       CHECK(false);
```
yschimke: Are these absolutely only internal invariants? e.g. data corruption on the wire could end up here?

lehecka: yes, if end up failing in the CHECK, we broke the invariant and didn't handle some cases. The frames sent over the wire should never cause the check fail.

PR #229, opened 2017-01-10
removing ExecutorBase::start

lehecka:
The ExecutorBase::start was used to be able to queue up method calls and start executing them later. The reason why we needed it was because stream automatons were implemented without SubscriberBase/SubscriptionBase and the methods were synchronous. We needed to set several pointers first before we allowed processing the stream frames.
Now all stream automatons are implemented with SubscriberBase/SubscriptionBase and so their methods are queued on executor automatically. We don't need to delay processing of the method calls anymore.


yschimke:
looks like a dodgy CI build with timeout, kick off again?

PR #230, opened 2017-01-11
Deliver onConnected after output has been requested

lexs:
This allows us to only deliver the onConnected after we know we have a valid connection, eg after HTTP 200.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -205,7 +205,8 @@ add_executable(
   test/SmartPointersTest.cpp
   test/AllowanceSemaphoreTest.cpp
   test/ResumeIdentificationTokenTest.cpp
-  test/PayloadTest.cpp)
+  test/PayloadTest.cpp
+  test/FrameTransportTest.cpp)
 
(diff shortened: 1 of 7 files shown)


lexs:
`TcpDuplexConnection` doesn't support this behavior as it takes an already connected `AsyncSocket` (so we can't install any `ConnectCallbacks`). This means it'll send the onConnected signal immediately too.

'Fixing' this requires changing the API of `TcpDuplexConnection` to require the connection to happen after it's taken ownership of the socket. This also doesn't seem great. Perhaps it's best to just ignore the `onConnected` signal and implement this on a higher level (basically don't give ReactiveSocket the `DuplexConnection` before we know it has connected).

Thoughts?

lexs:
We don't want this for now. I've implemented connectivity guarantees at the DuplexConnection level instead (guaranteeing it has basic connectivity before handing it to ReactiveSocket).

However, we probably wan't to revisit this at some point as RS needs to know about connectivity if we want to support fast-start.

PR #235, opened 2017-01-16
Add StreamCompletionSignal to Stats::socketClosed() call

martinsix:
We have very little insight into why sockets are being closed. This
diff adds the StreamCompletionSignal to the Stats interface, so that we
can distinguish the different reasons


lehecka:
LGTM. Just need to fix tests and make travis happy.

PR #236, opened 2017-01-17
Codemod folly::make_unique to std::make_unique

Orvid:
Folly will soon be dropping support for GCC 4.8, which means that `folly::make_unique` will be going away, so codemod away as much as possible before then.

### DIFF
--- a/tck-test/client.cpp
+++ b/tck-test/client.cpp
@@ -74,7 +74,7 @@ int main(int argc, char* argv[]) {
 
   evbt.getEventBase()->runInEventBaseThreadAndWait([&]() {
     socket.reset(new folly::AsyncSocket(evbt.getEventBase()));
-    callback = folly::make_unique<SocketConnectCallback>();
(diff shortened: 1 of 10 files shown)


yschimke:
I think this will be blocked on this PR working and landing

https://github.com/ReactiveSocket/reactivesocket-cpp/pull/187

This should work fine locally, but can't land until we get travis building with 4.9

Orvid:
Btw, the plan is to kill support for GCC 4.8 in Folly around mid February, which is when the last release of HHVM that needs to support GCC 4.8 will be released.

yschimke:
cc @lexs @lehecka

lehecka:
please rebase and check in

Orvid:
Rebased.

Orvid:
Rebased again.

yschimke:
follow up https://github.com/ReactiveSocket/reactivesocket-cpp/pull/277

PR #238, opened 2017-01-20
Callback signature

martinsix:
Instead of taking a StandardReactiveSocket&, these should just take a
ReactiveSocket&, so we can mock out reactivesocket for tests


martinsix:
It looks like the tests failed. Should we revert?

yschimke:
No I think those are existing failures from a merge back into github.  Can't really revert that, @lehecka is working on fixing those.

PR #239, opened 2017-01-20
Separating executor for ReactiveSocket and DuplexConnection

lehecka:
We need to be able to run ReactiveSocket and DuplexConnection each run on a different Executor.
This change introduces dispatching frames (FrameProcessor interface) to ConnectionAutomaton via ExecutorBase. Also Executor to DuplexConnection ctor is now explicit.

### DIFF
--- a/src/ConnectionAutomaton.cpp
+++ b/src/ConnectionAutomaton.cpp
@@ -393,7 +393,8 @@ void ConnectionAutomaton::onConnectionFrame(
         //          }
         //        }
       } else {
-        closeWithError(Frame_ERROR::connectionError("can not resume"));
+        closeWithError(
+            Frame_ERROR::connectionError("RS not resumable. Can not resume"));
       }
       return;
     }
@@ -490,7 +491,8 @@ bool ConnectionAutomaton::resumeFromPositionOrClose(
     resumeFromPosition(position);
     return true;
   } else {
-    closeWithError(Frame_ERROR::connectionError("can not resume"));
+    closeWithError(
+        Frame_ERROR::connectionError("Position not available. Can not resume"));
     return false;
   }
 }
--- a/src/ConnectionAutomaton.cpp
+++ b/src/ConnectionAutomaton.cpp
@@ -3,6 +3,7 @@
 #include "src/ConnectionAutomaton.h"
 
 #include <folly/ExceptionWrapper.h>
+#include <folly/MoveWrapper.h>
 #include <folly/String.h>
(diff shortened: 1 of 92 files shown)


src/ConnectionAutomaton.cpp
```diff
@@ -260,6 +277,15 @@ void ConnectionAutomaton::closeStreams(StreamCompletionSignal signal) {
 }
 
 void ConnectionAutomaton::processFrame(std::unique_ptr<folly::IOBuf> frame) {
+  auto thisPtr = this->shared_from_this();
+  auto movedFrame = folly::makeMoveWrapper(frame);
```
yschimke: this is making work for us to revert, I'll try to see if I can fix travis for 4.9

yschimke: Don't let this block you, its a current problem, not a new problem

src/StandardReactiveSocket.cpp
```diff
@@ -195,6 +211,23 @@ void StandardReactiveSocket::createResponder(
     ConnectionAutomaton& connection,
     StreamId streamId,
     std::unique_ptr<folly::IOBuf> serializedFrame) {
+  debugCheckCorrectExecutor();
+
+  if (streamId != 0) {
+    if (nextStreamId_ % 2 == streamId % 2) {
```
yschimke: TIOLI: this block inside "streamId != 0" seems involved enough to pull out.

yschimke:
LGTM

src/StandardReactiveSocket.cpp
lehecka: this chunk of code will be moved to a separate class. I already started working on it.

src/ConnectionAutomaton.cpp
```diff
@@ -345,11 +343,11 @@ void ConnectionAutomaton::processFrameImpl(
...
     folly::exception_wrapper ex,
     StreamCompletionSignal signal) {
-  auto thisPtr = this->shared_from_this();
-  auto movedEx = folly::makeMoveWrapper(ex);
-  runInExecutor([thisPtr, movedEx, signal]() mutable {
-    thisPtr->onTerminalImpl(movedEx.move(), signal);
-  });
+  runInExecutor([
+    thisPtr = this->shared_from_this(),
+    movedEx = std::move(ex),
```
yschimke: nit: do we need the moved prefix here?  What are our conventions for names with lambdas?

yschimke:
Nice!

PR #241, opened 2017-01-23
Make resumption token truly optional for SETUP

lexs:
Makes the SETUP frame not send/expect the resumption token if the RESUME_ENABLE flag isn't set.

This makes it possible for us to talk to Java for the time being. This isn't complying with any specific version of the spec, but makes us support the current version here together with the Java impl of 0.1 .

### DIFF
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -430,7 +430,11 @@ std::unique_ptr<folly::IOBuf> Frame_SETUP::serializeOut() {
   appender.writeBE(static_cast<uint32_t>(version_));
   appender.writeBE(static_cast<uint32_t>(keepaliveTime_));
   appender.writeBE(static_cast<uint32_t>(maxLifetime_));
-  appender.push((const uint8_t*)token_.data().data(), token_.data().size());
+
+  // TODO: Remove hack: https://github.com/ReactiveSocket/reactivesocket-cpp/issues/243
+  if (header_.flags_ & FrameFlags_RESUME_ENABLE) {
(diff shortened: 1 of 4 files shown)


test/SubscriberBaseTest.cpp
```diff
@@ -111,8 +111,8 @@ TEST(SubscriberBaseTest, CancelStopsOnNext) {
   EXPECT_CALL(*subscriber, onNextImpl_(_))
       .WillOnce(Invoke([&](int) {}))
       .WillOnce(Invoke([&](int) {
-        done = true;
         outSubscription->cancel();
+        done = true;
```
lexs: this test starting racing for me after my change (I know right) so I fixed it.

src/Frame.cpp
```diff
@@ -430,7 +430,11 @@ std::unique_ptr<folly::IOBuf> Frame_SETUP::serializeOut() {
   appender.writeBE(static_cast<uint32_t>(version_));
   appender.writeBE(static_cast<uint32_t>(keepaliveTime_));
   appender.writeBE(static_cast<uint32_t>(maxLifetime_));
-  appender.push((const uint8_t*)token_.data().data(), token_.data().size());
```
lehecka: please fix StandardReactiveSocket::clientConnect method to set the flag FrameFlags_RESUME_ENABLE if ConnectionSetupPayload.resumable == true. Otherwise we never send the Token in the setup frame.

lexs: fixed

PR #242, opened 2017-01-23
Swap odd/even for server/client streams to be spec compliant

lexs:
This makes C++ compliant with both the spec and the Java version. As long as only one side initiates requests this doesn't break old clients.

Depends on https://github.com/ReactiveSocket/reactivesocket-cpp/pull/241

### DIFF
--- a/src/StandardReactiveSocket.cpp
+++ b/src/StandardReactiveSocket.cpp
@@ -62,7 +62,7 @@ StandardReactiveSocket::StandardReactiveSocket(
           executeListenersFunc(onConnectListeners_),
           executeListenersFunc(onDisconnectListeners_),
           executeListenersFunc(onCloseListeners_))),
-      nextStreamId_(isServer ? 1 : 2),
+      nextStreamId_(!isServer ? 1 : 2),
       executor_(executor) {
   debugCheckCorrectExecutor();
   stats.socketCreated();
@@ -213,6 +213,8 @@ void StandardReactiveSocket::createResponder(
   debugCheckCorrectExecutor();
 
   if (streamId != 0) {
+// TODO: Remove this: https://github.com/ReactiveSocket/reactivesocket-cpp/issues/243
+#ifndef DONT_VERIFY_STREAM_IDS
     if (nextStreamId_ % 2 == streamId % 2) {
       // if this is an unknown stream to the socket and this socket is
       // generating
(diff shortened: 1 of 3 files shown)


yschimke:
This makes us compliant with the 1.0 spec, but not the earlier one we implemented.

LGTM - I'd like Ondrej to agree it's safe

lehecka:
We need to make this transition such that our old mobile clients will not break.
This change is a breaking change so I would suggest you introduce a FrameParser interface as part of this change and bump up version number to 0.1 (currently we are setting version = 0).

lexs:
I don't believe clients will break (I was going to test all of these changes after they land). reactivesocket-cpp doesn't care about the how the stream ids are assigned, it only fails if the clash.

This will only happen if we have bidirectional requests communication, something that doesn't exist today.

I rather make this change now as it unblocks the java<-->c++ compatibility for now.

I'd like to do the FrameParser interface eventually but that is a much larger refactoring.

lexs:
Can we merge this @lehecka?

lehecka:
Travis is unhappy. Please fix the build and make Mr. Travis green again.

lehecka:
Also one more comment on this. I believe there is some new code in StandardReactiveSocket::createResponder which is checking whether the instance is receiving frames with the same parity (even vs. odd) and throws them away...
This change should be tested at least manually between TcpServer and TcpClient where both are on the same even/odd stream numbers.

lexs:
@lehecka Yeah, saw that. Will fix and test.

lexs:
I fixed a test and also made the check for compliant stream ids be the default, we'll disable it internally to handle old clients.

src/StandardReactiveSocket.cpp
```diff
@@ -213,13 +213,16 @@ void StandardReactiveSocket::createResponder(
   debugCheckCorrectExecutor();
 
   if (streamId != 0) {
+// TODO: Remove this: https://github.com/ReactiveSocket/reactivesocket-cpp/issues/243
+#ifndef DONT_VERIFY_STREAM_IDS
```
yschimke: Is this really safe given the bump to lastPeerStreamId_ could now be a stream that we created?  Or we live for this assuming a one directional use case?

lehecka: As I understand it. We will allow server and client to generate the same Ids. In the lithium project we only create server streams and no client streams so they won't collide, but I hope we won't keep this hack in for a long time...

lehecka:
LGTM. Can you please rebase and land this change?

lehecka:
closing in favor of https://github.com/ReactiveSocket/reactivesocket-cpp/pull/301

PR #244, opened 2017-01-23
Make KEEPALIVE respect if resume is enabled

lexs:
This means we can sent keepalive to Java and properly handle the frames it sends back.

Depends on https://github.com/ReactiveSocket/reactivesocket-cpp/pull/242

### DIFF
--- a/src/ConnectionAutomaton.cpp
+++ b/src/ConnectionAutomaton.cpp
@@ -65,7 +65,7 @@ void ConnectionAutomaton::setResumable(bool resumable) {
   debugCheckCorrectExecutor();
   DCHECK(isDisconnectedOrClosed()); // we allow to set this flag before we are
   // connected
-  isResumable_ = resumable;
+  remoteResumeable_ = isResumable_ = resumable;
 }
 
 void ConnectionAutomaton::connect(
@@ -369,13 +369,13 @@ void ConnectionAutomaton::onConnectionFrame(
   switch (type) {
     case FrameType::KEEPALIVE: {
       Frame_KEEPALIVE frame;
-      if (!deserializeFrameOrError(frame, std::move(payload))) {
+      if (!deserializeFrameOrError(remoteResumeable_, frame, std::move(payload))) {
         return;
       }
       if (isServer_) {
         if (frame.header_.flags_ & FrameFlags_KEEPALIVE_RESPOND) {
           frame.header_.flags_ &= ~(FrameFlags_KEEPALIVE_RESPOND);
-          outputFrameOrEnqueue(frame.serializeOut());
+          outputFrameOrEnqueue(frame.serializeOut(remoteResumeable_));
         } else {
           closeWithError(
               Frame_ERROR::connectionError("keepalive without flag"));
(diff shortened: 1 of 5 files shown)


src/ConnectionAutomaton.cpp
```diff
@@ -320,13 +320,13 @@ void ConnectionAutomaton::onConnectionFrame(
   switch (type) {
     case FrameType::KEEPALIVE: {
       Frame_KEEPALIVE frame;
-      if (!deserializeFrameOrError(frame, std::move(payload))) {
+      if (!deserializeFrameOrError(isResumable_, frame, std::move(payload))) {
```
lehecka: I understand why the hack is necessary, but I think the implementation in deserialize method is not the desired one.
Assume this. Mobile clients are always sending RESUMABLE == true. And on Server we have a switch to turn on/off the resumability as a feature. So if client is ON and server is OFF the desired behavior is that we are still talking to each other, but without resumability. With you hack we will not agree on the format of the KEEPALIVE frame.
I would suggest a hack which would check the length of the IOBuf in the deserialize method and if it is sizeof(frame-in-spec)+sizeof(position), then we are getting the resumable position.

src/Frame.cpp
```diff
@@ -387,23 +387,31 @@ std::ostream& operator<<(std::ostream& os, const Frame_ERROR& frame) {
...
 /// @{
-std::unique_ptr<folly::IOBuf> Frame_KEEPALIVE::serializeOut() {
+std::unique_ptr<folly::IOBuf> Frame_KEEPALIVE::serializeOut(bool resumeable) {
   auto queue = createBufferQueue(FrameHeader::kSize + sizeof(ResumePosition));
   folly::io::QueueAppender appender(&queue, /* do not grow */ 0);
   header_.serializeInto(appender);
-  appender.writeBE(position_);
+  // TODO: Remove hack: https://github.com/ReactiveSocket/reactivesocket-cpp/issues/243
+  if (resumeable) {
+    appender.writeBE(position_);
```
lehecka: this is what we want.

src/ConnectionAutomaton.cpp
lexs: Ah, I assumed that isResumable_ was a fixed value. Didn't realize it could change or differ from the other side. I'll fix.

lexs: Reading the code is seems like isResumable_ is indeed fixed, there is even a TODO to add a check that client and server agrees:
> // TODO(tmont): check for ENABLE_RESUME and make sure isResumable_ is true

Spec also mentions this:
> NOTE: A server that receives a SETUP frame that has resume Enabled set, but does not support resuming operation, must reject the SETUP with an ERROR.

Seems like the behavior in this PR is correct, we should however support reconnecting without resumption (this isn't implemented).

lexs: Note I started implementing looking at frame size but it gets tricky fast as the KEEPALIVE frame has variable data.

lehecka: true, the spec is correct, however our gateways don't respect that. Mainly because the resumability was turned on on mobile before it was turned on on server.

lexs: Introspecting the frame doesn't even work right? We can't actually know if we're looking at the data size or if we're looking at the ResumePosition?

Given that all gateways should now run with resumability isn't this change fine to make?

lexs: How about this approach, store if the client cares about resumption or not (based on if resumeable if set in the SETUP frame?). I'll implement this now to show what I mean.

lexs:
I've updated this to infer if the client supports resumability based on if it sets the flag or not.

lehecka:
travis failed. Can you fix it/resubmit?

lexs:
Tests pass locally but master seems broken?

PR #246, opened 2017-01-23
Pause/resume subscription and subscriber

lehecka:
RequestHandler has 4 new callbacks:
onSubscriptionPaused / onSubscriptionResumed
onSubscriberPaused / onSubscriberResumed

the parameters are subscription / subscriber objects which the application code passed into RequestHandler::handleRequestXXX(...subscriber::onSubscribe(subscription);...) (on the server) / ReactiveSocket::request_XXX (on the client)

This allows the application code (both client and server) to process the pause/resume signal for each stream individually.


lehecka:
merging to unblock next diffs. 
Any feedback would be appreciated and I will incorporate any changes later.

PR #249, opened 2017-01-24
Ignore unexpected frames instead of dying

lexs:
Following the spec we should ignore things we don't expect/understand instead of failing hard. This commit also makes the logging better in these cases.

This is needed as reactivesocket-java in fact sends a RESPONSE(complete) after each subscription frame, see https://github.com/ReactiveSocket/reactivesocket-java/issues/226.

### DIFF
--- a/src/automata/StreamAutomatonBase.cpp
+++ b/src/automata/StreamAutomatonBase.cpp
@@ -6,49 +6,53 @@
 
 namespace reactivesocket {
 
+namespace {
+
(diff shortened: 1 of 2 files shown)


lexs:
Note that this follows the Java impl and is allowed by the spec:
> An implementation MAY send an ERROR frame (with CONNECTION_ERROR error code) and close the underlying transport connection on reception of a frame that it does not understand with this bit not set.

PR #251, opened 2017-01-25
fixing pedantic warnings

lehecka:


yschimke:
How does this include actually work? including either one includes the other?

jspahrsummers:
Since one of these is conceptually at the end of the other (i.e., `ConsumerMixin-inl.h` and `ConsumerMixin.h`) and the `#pragma once`s ensure that neither gets included twice, this should stitch them together properly regardless of which one is included.

But it's a bit unintuitive, so maybe it should be changed. ☺️

yschimke:
OK, so the point is that one includes at the end and so the order is always enforced.

Is this common? I found this ref in Google style guide. 

> Prefer placing the definitions for template and inline functions in the same file as their declarations. The definitions of these constructs must be included into every .cc file that uses them, or the program may fail to link in some build configurations. If declarations and definitions are in different files, including the former should transitively include the latter. Do not move these definitions to separately included header files (-inl.h); this practice was common in the past, but is no longer allowed.

jspahrsummers:
I don't know whether it's common or not, and I don't think Google's style guide is a reference we follow closely.

The advantage is that it makes reading the header easier, as the implementation details are somewhere else. But personally, I'm partial to doing this with an IDE or documentation generator, and keeping everything inline where it belongs.

PR #253, opened 2017-01-25
Travis upgrade (GCC 4.9, 5, 6, CLANG 3.6 3.7)

lehecka:
After long battles with Travis I managed to upgrade the toolset to GCC 4.9, 5, 6, CLANG 3.6 3.7 (all major releases compatible with c++14 and higher).

This should unblock some of the PR waiting for the upgrade.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,10 +40,6 @@ matrix:
           sources:
             - ubuntu-toolchain-r-test
 
-    - env: GCC_VERSION=5 BUILD_TYPE=Debug CPP=11 ASAN=Off
-      os: linux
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -1,21 +1,83 @@
...
-os:
-  - linux
-  
-compiler:
-  - gcc
-  - clang
+# Test matrix:
+matrix:
+   include:
+    - env: CLANG_VERSION=3.6 BUILD_TYPE=Debug CPP=11 ASAN=On
```
yschimke: Why keep this CPP=11 build?

PR #259, opened 2017-01-26
Make `Stats::frameRead` and `Stats::frameWritten` accept `FrameType` …

th0br0:
…instead of `std::string`

@lexs

### DIFF
--- a/src/ConnectionAutomaton.cpp
+++ b/src/ConnectionAutomaton.cpp
@@ -273,11 +273,7 @@ void ConnectionAutomaton::resumeStreams() {
 
 void ConnectionAutomaton::processFrame(std::unique_ptr<folly::IOBuf> frame) {
   auto frameType = FrameHeader::peekType(*frame);
-
-  std::stringstream ss;
-  ss << frameType;
-
-  stats_.frameRead(ss.str());
+  stats_.frameRead(frameType);
 
   // TODO(tmont): If a frame is invalid, it will still be tracked. However, we
   // actually want that. We want to keep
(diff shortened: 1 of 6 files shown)


src/Stats.h
```diff
@@ -27,8 +27,8 @@ class Stats {
 
   virtual void bytesWritten(size_t bytes) = 0;
   virtual void bytesRead(size_t bytes) = 0;
-  virtual void frameWritten(const std::string& frameType) = 0;
-  virtual void frameRead(const std::string& frameType) = 0;
+  virtual void frameWritten(const FrameType frameType) = 0;
```
lexs: you can kill const here as it's a value type now

src/ConnectionAutomaton.cpp
```diff
@@ -274,10 +274,7 @@ void ConnectionAutomaton::resumeStreams() {
 void ConnectionAutomaton::processFrame(std::unique_ptr<folly::IOBuf> frame) {
   auto frameType = FrameHeader::peekType(*frame);
```
lexs: very nit: i'd kill this whitespace

lexs:
LGTM, letting @lehecka merge

lehecka:
I am not sure if FrameType enum is part of the public API. Can you verify that?

th0br0:
How do you mean `public api`?
It's part of the spec as of https://github.com/ReactiveSocket/reactivesocket/blob/master/Protocol.md#frame-types (although,  although looking at the protocol definition, there's a discrepancy between our definitions and the official protocol spec - `REQUEST_SUB` is missing from the latter)

lehecka:
public API of the cpp library, not the spec.
I am not sure if the application code has access to the internal enum type.

th0br0:
`Frame.h` is pulled in via `reactivesocket-plugins`.

PR #260, opened 2017-01-26
Connect callback

martinsix:
We were calling the `onConnected` callback before finishing setting up
the connection.

### DIFF
--- a/src/ConnectionAutomaton.cpp
+++ b/src/ConnectionAutomaton.cpp
@@ -64,7 +64,6 @@ void ConnectionAutomaton::connect(
   CHECK(!frameTransport->isClosed());
 
   frameTransport_ = std::move(frameTransport);
-  onConnected_();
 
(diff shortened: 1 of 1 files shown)


src/ConnectionAutomaton.cpp
```diff
@@ -87,6 +86,7 @@ void ConnectionAutomaton::connect(
       keepaliveTimer_->start(shared_from_this());
     }
   }
+  onConnected_();
```
lehecka: by changing this order, we might get into the situation that sending frames on line 82 may terminate socket and the client code would get onClose before onConnected.

lehecka:
this needs a different solution. Will rest this change for now.

lexs:
I have some ideas to "fix" onConnected once and for all but that requires delegating the behavior to DuplexConnection, which is a larger change (and I won't get that done in the near future).

lehecka:
closing as obsolete.

PR #267, opened 2017-01-27
Add hack ifdef to assume assumption

lexs:
This is needed for now to handle old clients as they incorrectly didn't set the header but assumed resumption.

### DIFF
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -465,6 +465,12 @@ bool Frame_SETUP::deserializeFrom(std::unique_ptr<folly::IOBuf> in) {
 
     // TODO: Remove hack:
     // https://github.com/ReactiveSocket/reactivesocket-cpp/issues/243
+#ifdef HACK_ASSUME_RESUMPTION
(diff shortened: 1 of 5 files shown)


src/Frame.cpp
```diff
@@ -465,6 +465,9 @@ bool Frame_SETUP::deserializeFrom(std::unique_ptr<folly::IOBuf> in) {
 
     // TODO: Remove hack:
     // https://github.com/ReactiveSocket/reactivesocket-cpp/issues/243
+#ifdef HACK_ASSUME_RESUMPTION
+    header_.flags_ |= FrameFlags_RESUME_ENABLEö
```
yschimke: WAT?!
`
FrameFlags_RESUME_ENABLEö`

lexs:
Updated, I switched the strategy to bump the version, this means we can assume resumption for older clients with version == 0 while everything else is running at version == 1.

lexs:
Reopened to trigger travis.

lehecka:
closing as this is not handled differently and the PR is obsolete.

PR #272, opened 2017-01-30
Fix connection automaton client param

yschimke:


src/ConnectionAutomaton.h
```diff
@@ -78,7 +78,7 @@ class ConnectionAutomaton
       ResumeListener resumeListener,
       Stats& stats,
       std::unique_ptr<KeepaliveTimer> keepaliveTimer_,
-      bool client,
+      bool isServer,
```
yschimke: now matches both

https://github.com/ReactiveSocket/reactivesocket-cpp/blob/master/src/ConnectionAutomaton.cpp#L27

and

https://github.com/ReactiveSocket/reactivesocket-cpp/blob/master/src/ConnectionAutomaton.h#L225

jspahrsummers:
Please consider using an `enum` instead, or separate classes.

PR #273, opened 2017-01-30
moving initialRequestN to ctor of response automatons

lehecka:
and removing some dead code.


yschimke:
Looks like a great cleanup, less boilerplate, less logging.

PR #274, opened 2017-01-30
Introducing StreamsFactory class for creating stream requesters and responders.

lehecka:


yschimke:
Could you clean first and rebase on the cleanup version, seems mostly formatting here.

PR #276, opened 2017-01-31
Disconnect instead of close

yschimke:


src/folly/FollyKeepaliveTimer.cpp
```diff
@@ -34,8 +34,7 @@ void FollyKeepaliveTimer::schedule() {
 void FollyKeepaliveTimer::sendKeepalive() {
   if (pending_) {
     stop();
-    connection_->closeWithError(
-        Frame_ERROR::connectionError("no response to keepalive"));
+    connection_->disconnect();
```
lehecka: I would like to keep the error message for the close. It would help with debugging.

src/ConnectionAutomaton.cpp
```diff
@@ -119,10 +119,14 @@ void ConnectionAutomaton::disconnect() {
     return;
   }
 
-  closeFrameTransport(folly::exception_wrapper());
-  pauseStreams();
-  stats_.socketDisconnected();
-  onDisconnected_();
+  if (isResumable_) {
```
lehecka: The ConnectionAutomaton::disconnect should not change behavior and close the socket. This method can be called from application code thru ReactiveSocket::disconnect() and the expectation is that it will just disconnect.

I would suggest creating one of the following new methods:
a) disconnectOrCloseWithError(Frame_ERROR)  (more generic approach)
b) keepaliveIntervalFailed(...)    (concrete purpose)

yschimke: what does disconnect for a non resumable socket mean?

lehecka:
Closing in the favor of https://github.com/ReactiveSocket/reactivesocket-cpp/pull/284.

PR #279, opened 2017-02-01
excluding connection level ERROR frame from resume cache

lehecka:


src/ResumeTracker.cpp
```diff
@@ -0,0 +1,49 @@
...
+    case FrameType::REQUEST_STREAM:
+    case FrameType::REQUEST_SUB:
+    case FrameType::REQUEST_RESPONSE:
+    case FrameType::REQUEST_FNF:
+    case FrameType::REQUEST_N:
+    case FrameType::CANCEL:
+    case FrameType::RESPONSE:
+      return true;
+
+    case FrameType::ERROR: {
```
yschimke: why are connection errors excluded?  So close with an error is resumable?

lehecka: I will close this PR as I am not confident this is an issue.

PR #283, opened 2017-02-01
one keepalive loop

yschimke:


src/folly/FollyKeepaliveTimer.cpp
```diff
@@ -20,10 +20,13 @@ std::chrono::milliseconds FollyKeepaliveTimer::keepaliveTime() {
...
-  auto running = running_;
+  auto scheduledState = state_.get();
+  auto currentState = state_;
   eventBase_.runAfterDelay(
-      [this, running]() {
-        if (*running) {
+      [this, currentState, scheduledState]() {
+        LOG(INFO) << "State " << *currentState << " " << scheduledState;
+
+        if (currentState.get() == scheduledState) {
```
lehecka: you are comparing addresses here, not the value on stored in the shared_ptr

src/folly/FollyKeepaliveTimer.cpp
```diff
@@ -8,7 +8,7 @@ FollyKeepaliveTimer::FollyKeepaliveTimer(
     folly::EventBase& eventBase,
     std::chrono::milliseconds period)
     : eventBase_(eventBase),
-      running_(std::make_shared<bool>(false)),
+      state_(std::make_shared<uint32_t>(0)),
```
lehecka: do you want to rename state_ to generation_ and initialize the member variable on line 33?

PR #284, opened 2017-02-03
Disconnect or Close on keepalive fail

lehecka:
I also had to clean up the termination signals and the mechanism when we close frameTransport with onComplete vs. onError.

### DIFF
--- a/src/ConnectionAutomaton.cpp
+++ b/src/ConnectionAutomaton.cpp
@@ -24,28 +24,19 @@ ConnectionAutomaton::ConnectionAutomaton(
     ResumeListener resumeListener,
     Stats& stats,
     std::unique_ptr<KeepaliveTimer> keepaliveTimer,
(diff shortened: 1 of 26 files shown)


src/ConnectionAutomaton.cpp
```diff
@@ -191,10 +207,13 @@ void ConnectionAutomaton::closeWithError(Frame_ERROR&& error) {
     case ErrorCode::REJECTED_SETUP:
       signal = StreamCompletionSignal::REJECTED_SETUP;
       break;
-    case ErrorCode::CONNECTION_ERROR:
-      signal = StreamCompletionSignal::CONNECTION_ERROR;
-      break;
 
+    case ErrorCode::
+        CONNECTION_ERROR:
```
yschimke: two lines?

yschimke:
Looks good, just one case with formatting screwed

PR #291, opened 2017-02-03
Throw instead of asserting on invalid metadata sizes

alexmalyshev:
Also fix a bug in Payload's string ctor where it will always make a metadata
IOBuf, regardless if the metadata string is empty or not.

### DIFF
--- a/src/Payload.cpp
+++ b/src/Payload.cpp
@@ -1,12 +1,13 @@
 // Copyright 2004-present Facebook. All Rights Reserved.
 
 #include "Payload.h"
+
 #include <folly/String.h>
 #include <folly/io/Cursor.h>
 
 namespace reactivesocket {
 
-constexpr static auto kMaxMetadataLength = std::numeric_limits<int32_t>::max();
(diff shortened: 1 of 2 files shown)


src/Payload.cpp
```diff
@@ -45,20 +53,28 @@ void Payload::serializeInto(folly::io::QueueAppender& appender) {
...
 
-    const auto metadataPayloadLength = length - sizeof(uint32_t);
+  if (length >= kMaxMetadataLength) {
+    throw std::runtime_error("Metadata is too big to deserialize");
+  }
 
-    if (metadataPayloadLength > 0) {
-      cur.clone(metadata, metadataPayloadLength);
-    }
+  if (length <= sizeof(uint32_t)) {
```
yschimke: What is intended here?

yschimke: I think we had a bug here

yschimke: Ignore, it has changed in version 1. so I think we are ok until we adopt the new framing format.

https://github.com/ReactiveSocket/reactivesocket/blob/0.2/Protocol.md

PR #293, opened 2017-02-07
Fix crash in FollyKeepaliveTimer

Cristinag:
Only re-schedule a keepalive if got a reply back from the previous one.


yschimke:
ReactiveSocketConcurrencyTest asan failure in one of the builds, retrying.

yschimke:
cc @lehecka

PR #294, opened 2017-02-07
ServerConnectionAcceptor for simplifying setup vs. resume scenario

lehecka:


CMakeLists.txt
```diff
@@ -180,7 +180,7 @@ add_library(
   src/FrameTransport.cpp
   src/FrameTransport.h
   src/FrameProcessor.h
-  src/ReactiveSocket.h src/StreamsFactory.cpp src/StreamsFactory.h)
+  src/ReactiveSocket.h src/StreamsFactory.cpp src/StreamsFactory.h src/ServerConnectionAcceptor.cpp src/ServerConnectionAcceptor.h)
```
yschimke: can you reformat these, or combine the rest onto fewer lines?

src/ConnectionSetupPayload.cpp
```diff
@@ -9,6 +9,8 @@ std::ostream& operator<<(
     const ConnectionSetupPayload& setupPayload) {
   return os << "[metadataMimeType: " << setupPayload.metadataMimeType
             << " dataMimeType: " << setupPayload.dataMimeType
-            << " payload: " << setupPayload.payload << "]";
+            << " payload: " << setupPayload.payload << "]"
```
yschimke: why end the bracket here?

src/ConnectionSetupPayload.cpp
```diff
@@ -9,6 +9,8 @@ std::ostream& operator<<(
...
+            << " token: " << setupPayload.token.toString()
```
yschimke: should token support operator<<?

test/simple/StatsPrinter.h
```diff
@@ -24,5 +24,8 @@ class StatsPrinter : public Stats {
   void frameWritten(const std::string& frameType) override;
   void frameRead(const std::string& frameType) override;
   void resumeBufferChanged(int framesCountDelta, int dataSizeDelta) override;
+
+  void connectionClosedInServerConnectionAcceptor(
```
yschimke: nit: this seems awfully specific.  Stats is really only for reporting on errors, could we use a more generic name.  badIncomingConnection?

test/resume/TcpResumeServer.cpp
```diff
@@ -175,35 +142,23 @@ class ServerRequestHandler : public DefaultRequestHandler {
...
-        eventBase_(eventBase),
-        stats_(stats){};
+  MyServerConnectionAcceptor(EventBase& eventBase, Stats& stats)
+      : ServerConnectionAcceptor(stats), eventBase_(eventBase) {}
 
-  virtual ~Callback() = default;
+  void setupNewSocket(
+      std::shared_ptr<FrameTransport> frameTransport,
+      ConnectionSetupPayload setupPayload) override {
+    CHECK(setupPayload.resumable);
```
yschimke: it seems useful that this server is written as an example that can accept requests from either the resume client or standard client and just fallback to no resumption.  e.g. CHECK seems extreme.

test/resume/TcpResumeServer.cpp
```diff
@@ -223,26 +178,57 @@ class Callback : public AsyncServerSocket::AcceptCallback {
...
-                    ResumeIdentificationToken>& kv) {
-            return kv.first.get() == &socket;
-          }));
-    }
+  void resumeSocket(
+      std::shared_ptr<FrameTransport> frameTransport,
+      ResumeIdentificationToken token,
+      ResumePosition position) override {
+    LOG(INFO) << "MyServerConnectionAcceptor::resumeSocket resume token "
+              << token.toString();
```
yschimke: what are we, java developers?

yschimke:
LGTM - I'd open it up for wider discussion though

src/ConnectionSetupPayload.cpp
lehecka: I will remove them completely..

src/ConnectionSetupPayload.cpp
lehecka: we couldn't agree on the format so we didn't make it printable thru <<

test/simple/StatsPrinter.h
lehecka: I think this should be very specific because it is used for one specific case. I believe we should have way more specific counters all over the library. Essentially the idea is that for debugging reasons you want to know if you library is hitting any edge cases in production, if you are refusing connections / closing / rejecting frames and such. Would you agree with that statement?

test/resume/TcpResumeServer.cpp
lehecka: since this is a test code, I dont care. Will remove the check

test/resume/TcpResumeServer.cpp
lehecka: :) 
I guess not.

PR #296, opened 2017-02-07
Fix crash on empty payload in FramedReader

vitaut:
This PR fixes handling of empty payload in `FramedReader` by allocating an empty `IOBuf` object and passing a `unique_ptr` to it to `onNext`. Previously a null `unique_ptr<IOBuf>` was passed causing a crash when it was dereferenced in `FrameHeader::peekType()`.


lehecka:
looks great. This code change exposed a "use-after-free" bug shown in travis. I will try to address that before merging this in.

PR #298, opened 2017-02-08
Remove Frame template param from ConsumerMixin

somasun:
This is part of an effort to get rid of mixins/ConsumerMixin and
mixins/ProducerMixin and replace them with regular base classes instead.

### DIFF
--- a/src/mixins/ConsumerMixin.h
+++ b/src/mixins/ConsumerMixin.h
@@ -7,10 +7,12 @@
 #include <iostream>
 #include "src/AllowanceSemaphore.h"
 #include "src/Common.h"
+#include "src/ConnectionAutomaton.h"
 #include "src/NullRequestHandler.h"
 #include "src/Payload.h"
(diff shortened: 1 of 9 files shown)


src/mixins/ConsumerMixin.h
```diff
@@ -7,18 +7,19 @@
...
 #include "src/ReactiveStreamsCompat.h"
 #include "src/SmartPointers.h"
+#include "src/SubscriptionBase.h"
 #include "src/automata/StreamAutomatonBase.h"
 
 namespace reactivesocket {
 
 enum class StreamCompletionSignal;
 
 /// A mixin that represents a flow-control-aware consumer of data.
```
lehecka: Please update the comment.
Also can you move all of the method bodies from the .h file to the .cpp file?

lehecka:
Looks great! Merging.
Please rename ConsumerMixin to ConsumerAutomatonBase

PR #301, opened 2017-02-09
introducing FrameSerializer

lehecka:
As we are moving toward upgrading to protocol v1.0 and supporting old versions we need to separate serialization/deserialization from the frames. This PR introduces the interface and also fixes compatibility bug with Java v0 version.


lehecka:
reviewed in fbcode, closing

PR #306, opened 2017-02-14
Add missing files to CMakeLists.txt that were causing test failures

jspahrsummers:
Not sure if my setup of auto-syncing broke this, or if it was already broken, but in any case this looks like the correct fix.


lexs:
LGTM

PR #310, opened 2017-02-17
RSocket API and Examples

benjchristensen:
I'm working on examples and experimentation with the API. I'd like to get some feedback before I continue, so submitting this for discussion, but don't expect it to be merged right now.

1) I intend on creating various usage examples, such as:

- simple stream (the one in this PR)
- stream cancellation, request(n), etc
- fast stream with slow consumer and different flow control strategies
- bidi channel
- request/response
- symmetric communication with server calling client

Are those useful to have examples of? Are there others to do?

2) I want to demonstrate these examples with simpler code and more natural APIs than currently exist in https://github.com/ReactiveSocket/reactivesocket-cpp/tree/master/test/tcp

I'm playing with the `RSocket.h` API. I'd appreciate knowing if the direction I'm taking is considered worth pursuing. 

It should allow clients like this:

```cpp
  ScopedEventBaseThread eventBaseThread;
  auto rsf = RSocket::createClientFactory(
      ConnectionFactory::tcpClient(FLAGS_host, FLAGS_port));
  rsf->connect(eventBaseThread)
      .then([](std::shared_ptr<StandardReactiveSocket> rs) {
        rs->requestStream(
            Payload("args-here"), std::make_shared<ExampleSubscriber>(5, 6));
      });
```

or this:

```cpp
  ScopedEventBaseThread eventBaseThread;
  auto rsf = RSocket::createClientFactory(
      ConnectionFactory::tcpClient(FLAGS_host, FLAGS_port));
    auto rs = rsf->connect(eventBaseThread).get();
    rs->requestStream(
        Payload("args-here"), std::make_shared<ExampleSubscriber>(5, 6));
```

3) Once I have client and server with simple APIs like this, I would then like to tackle the `rs->requestStream` part which currently takes a `Subscriber` and instead allow returning a `Flowable` for fluent composition, similar to the Folly Future. But that's a whole different pursuit.


benjchristensen:
And note to everyone that I'm a C++ novice, so by all means tell me where I'm doing it completely wrong! There's a reason I'm doing examples and experimentation, rather than touching the core code :-)

fiorix:
This is definitely very useful. I also think it'd be important to have small code snippets right in the README for each repo so people can have a basic idea of how to use it even before they open any source file.

examples/util/RSocket.h
```diff
@@ -0,0 +1,36 @@
...
+namespace rsocket {
+
+class RSocket {
+ public:
+  static std::unique_ptr<RSocket> createClientFactory(
+      std::unique_ptr<ConnectionFactory> connection);
+  RSocket(std::unique_ptr<ConnectionFactory> connection);
+  ~RSocket();
+  /*
+   * Connect asynchronously and return a Future
```
lexs: Making connect in itself async has some downsides like not allowing fast start as I can't issue requests before/after the connect call happens (and before it returns).

This is also something we currently depend on for clients, where we'll kick of the connect() call on the first subscription from user code but still allow all of of them to transparently succeed.

lexs:
I like the idea of constructing a RSocket that knows how to actually establish the connection itself, we have something very similar internally.

As mentioned in the code comment I still think it would be very useful if this new API allows to construct an unconnected socket.

> I intend on creating various usage examples, such as

I would love if these can also pass as small integration tests.

examples/util/RSocket.h
benjchristensen: Good point regarding fast start. I wonder if the API should be explicit about opting into that, such as connectFast() which communicates that fast start will be done, which means all requests may fail, rather than waiting for a successful connection?

benjchristensen: Reason being that there are many times when fast start isn't needed/wanted, such as when building up a connection pool. In those times I want connect() to call me back with the connection when it's ready.

benjchristensen:
>  I also think it'd be important to have small code snippets right in the README for each repo so people can have a basic idea of how to use it even before they open any source file.

Agreed. As code evolves let's definitely do that.

benjchristensen:
> I would love if these can also pass as small integration tests.

Agreed. I think once the examples are all in place we can incorporate them as a dependency of the tests.

benjchristensen:
Added the most basic type of server API as a starting point in my experimentation:

```c++
  auto rs = RSocket::createServer(
      ServerConnectionAcceptor::tcpServer(FLAGS_port),
      []() { return std::make_unique<ServerRequestHandler>(); });
  rs->start();
```

Doing this brought up some questions for me:

- Why can't we share a single `RequestHandler` across many connections? 
- Shouldn't we be able to receive the `SETUP` frame before deciding what `RequestHandler` to return? (The Java impl follows this model)

benjchristensen:
Current status is we have a server:

```c++
  auto rs = RSocket::createServer(
      ServerConnectionAcceptor::tcpServer(FLAGS_port),
      []() { return std::make_unique<ServerRequestHandler>(); });
  rs->start();
```

and client:

```c++
  ScopedEventBaseThread eventBaseThread;
  auto rsf = RSocket::createClientFactory(
      ClientConnectionFactory::tcpClient(FLAGS_host, FLAGS_port));
  rsf->connect(eventBaseThread)
      .then([](std::shared_ptr<StandardReactiveSocket> rs) {
        rs->requestStream(
            Payload("args-here"), std::make_shared<ExampleSubscriber>(5, 6));
      });
```

... with many aspects of functionality not exposed yet in these APIs.

benjchristensen:
Server looking like this:

```cpp
  // RSocket server accepting on TCP
  auto rs = RSocket::createServer(TcpConnectionAcceptor::create(FLAGS_port));
  // global request handler
  auto requestHandler = std::make_shared<ServerRequestHandler>();
  // start accepting connections
  rs->start([requestHandler](auto connectionRequest) {
    if (!connectionRequest->isResumptionAttempt()) {
      return requestHandler;
    } else {
      throw std::runtime_error("Unable to handle these SETUP params");
    }
  });
```

This now supports a global request handler, and conditional handling of SETUP and RESUME to then error or return a different request handler (not all implemented yet, but flow is in place).

These changes required touching the actual production code in a few places.

PR #319, opened 2017-03-08
[reactivsocket-cpp/build] Remove unnecessary boost dependency

somasun:
boost is not used used anywhere in the code


lehecka:
we need to fix travis first

PR #320, opened 2017-03-09
Build folly from source on Travis

lexs:
We now build folly from a release instead of having a prepackaged build. Currently using a specific Github release but we might switch to just building master.


scripts/build_folly.sh
```diff
@@ -0,0 +1,71 @@
...
+      libsnappy-dev \
+      make \
+      zlib1g-dev \
+      binutils-dev \
+      libjemalloc-dev \
+      libssl-dev \
+      libiberty-dev
+else
+  export CPPFLAGS="-I/usr/local/opt/openssl/include"
+  export LDFLAGS="-L/usr/local/opt/openssl/lib"
```
somasun: curious.  why do we need openssl

lexs: Folly needs this, not sure exactly for what.

PR #321, opened 2017-03-09
Fluent API for Flowable and Subscriber/Subscription

vjn:
Main goal is to specify user functionality as function objects (lambdas), and move/copy them into the flowable/subscriber. This is an example, and is far from complete: particularly, there is one operator example, map, but its code needs to be factored properly and used in writing several more.

### DIFF
--- a/experimental/vjn/yarpl/include/reactivestreams/subscriber.h
+++ b/experimental/vjn/yarpl/include/reactivestreams/subscriber.h
@@ -7,104 +7,86 @@
 
 namespace reactivestreams {
 
-template<typename T>
+template <typename T>
 class Subscription;
 
-template<typename T>
+template <typename T>
 class Subscriber {
-public:
+ public:
   virtual ~Subscriber() = default;
 
-  virtual void on_next(const T& value) {}
+  virtual void on_next(const T&) {}
   virtual void on_next(T&& value) { on_next(value); }
   virtual void on_error(const std::exception& error) { throw error; }
   virtual void on_complete() {}
-  virtual void on_subscribe(Subscription<T>* subscription) {}
+  virtual void on_subscribe(Subscription<T>*) {}
 
-  template<
-      typename Next,
-      typename = typename std::enable_if<
-          std::is_callable<Next(const T&), void>::value
-          >::type>
+  template <typename Next,
+            typename = typename std::enable_if<
+                std::is_callable<Next(const T&), void>::value>::type>
   static std::unique_ptr<Subscriber<T>> create(Next&& next) {
     return std::unique_ptr<Subscriber<T>>(
         new Derived1<Next>(std::forward<Next>(next)));
   }
 
-  template<
-      typename Next,
-      typename Error,
+  template <
+      typename Next, typename Error,
       typename = typename std::enable_if<
           std::is_callable<Next(const T&), void>::value &&
-          std::is_callable<Error(const std::exception&), void>::value
-          >::type>
+          std::is_callable<Error(const std::exception&), void>::value>::type>
   static std::unique_ptr<Subscriber<T>> create(Next&& next, Error&& error) {
-    return std::unique_ptr<Subscriber<T>>(
-        new Derived2<Next, Error>(
-            std::forward<Next>(next),
-            std::forward<Error>(error)));
+    return std::unique_ptr<Subscriber<T>>(new Derived2<Next, Error>(
+        std::forward<Next>(next), std::forward<Error>(error)));
   }
 
-  template<
-      typename Next,
-      typename Error,
-      typename Complete,
-      typename = typename std::enable_if<
-          std::is_callable<Next(const T&), void>::value &&
-          std::is_callable<Error(const std::exception&), void>::value &&
-          std::is_callable<Complete(), void>::value
-          >::type>
-  static std::unique_ptr<Subscriber<T>> create(
-      Next&& next, Error&& error, Complete&& complete) {
-    return std::unique_ptr<Subscriber<T>>(
-        new Derived3<Next, Error, Complete>(
-            std::forward<Next>(next),
-            std::forward<Error>(error),
-            std::forward<Complete>(complete)));
+  template <typename Next, typename Error, typename Complete,
+            typename = typename std::enable_if<
+                std::is_callable<Next(const T&), void>::value &&
+                std::is_callable<Error(const std::exception&), void>::value &&
+                std::is_callable<Complete(), void>::value>::type>
+  static std::unique_ptr<Subscriber<T>> create(Next&& next, Error&& error,
+                                               Complete&& complete) {
+    return std::unique_ptr<Subscriber<T>>(new Derived3<Next, Error, Complete>(
+        std::forward<Next>(next), std::forward<Error>(error),
+        std::forward<Complete>(complete)));
   }
 
-private:
-  template<typename Next>
+ private:
+  template <typename Next>
   class Derived1 : public Subscriber {
-  public:
+   public:
     Derived1(Next&& next) : next_(std::forward<Next>(next)) {}
 
-    virtual void on_next(const T& value) {
-      next_(value);
-    }
+    virtual void on_next(const T& value) { next_(value); }
 
-  private:
+   private:
     Next next_;
   };
 
-  template<typename Next, typename Error>
+  template <typename Next, typename Error>
   class Derived2 : public Derived1<Next> {
-  public:
+   public:
     Derived2(Next&& next, Error&& error)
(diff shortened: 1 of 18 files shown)


benjchristensen:
Thank you Vijayan.

benjchristensen:
I will build on top of this commit to merge into the rest of the repo.

experimental/vjn/yarpl/include/reactivestreams/subscriber.h
```diff
@@ -0,0 +1,94 @@
...
+
+template <typename T>
+class Subscription;
+
+template <typename T>
+class Subscriber {
+ public:
+  virtual ~Subscriber() = default;
+
+  virtual void on_next(const T&) {}
```
lehecka: nit: can't say that I like the naming conventions. It looks like naming conventions from the standard library, but the class name starts with a capital letter, so this feels like a hybrid ;)

experimental/vjn/yarpl/include/reactivestreams/subscriber.h
```diff
@@ -0,0 +1,94 @@
...
+  virtual void on_next(T&& value) { on_next(value); }
```
lehecka: this overload seems unnecessary as rvalue reference will automatically bind to const lvalue reference...

experimental/vjn/yarpl/include/reactivestreams/subscriber.h
lehecka: did you want to provide default body to the virtual methods or was this supposed to be the interface only?

lehecka: I would provide only T&& parameter version of the method. Otherwise if T is a big type, then the application code will take the hit by copying a big type... These things should be explicit in C++ in my opinion.

experimental/vjn/yarpl/include/reactivestreams/subscriber.h
```diff
@@ -0,0 +1,94 @@
...
+  virtual void on_error(const std::exception_ptr error) { throw error; }
+  virtual void on_complete() {}
+  virtual void on_subscribe(Subscription<T>*) {}
+
+  template <typename Next,
+            typename = typename std::enable_if<
+                std::is_callable<Next(const T&), void>::value>::type>
+  static std::unique_ptr<Subscriber<T>> create(Next&& next) {
+    return std::unique_ptr<Subscriber<T>>(
```
lehecka: nit: use std::make_unique

experimental/vjn/yarpl/include/reactivestreams/subscriber.h
```diff
@@ -0,0 +1,94 @@
...
+        std::forward<Complete>(complete)));
+  }
+
+ private:
+  template <typename Next>
+  class Derived1 : public Subscriber {
+   public:
+    Derived1(Next&& next) : next_(std::forward<Next>(next)) {}
+
+    virtual void on_next(const T& value) { next_(value); }
```
lehecka: nit: override keyword

experimental/vjn/yarpl/include/reactivestreams/subscriber.h
```diff
@@ -0,0 +1,94 @@
...
+    return std::unique_ptr<Subscriber<T>>(new Derived3<Next, Error, Complete>(
+        std::forward<Next>(next), std::forward<Error>(error),
+        std::forward<Complete>(complete)));
+  }
+
+ private:
+  template <typename Next>
+  class Derived1 : public Subscriber {
+   public:
+    Derived1(Next&& next) : next_(std::forward<Next>(next)) {}
```
lehecka: std::move is appropriate, not std::forward

lehecka: this ctor should be also marked as explicit

experimental/vjn/yarpl/include/reactivestreams/subscription.h
```diff
@@ -0,0 +1,48 @@
...
+  virtual void request(long n) = 0;
+  virtual void cancel() = 0;
+};
+
+template <typename T, typename Next>
+class DeletingSubscription : public Subscription<T> {
+ public:
+  DeletingSubscription(std::shared_ptr<Next> function,
+                       std::unique_ptr<Subscriber<T>> subscriber)
+      : function_(function), subscriber_(std::move(subscriber)) {}
```
lehecka: you can move the function into function_

experimental/vjn/yarpl/include/reactivestreams/subscription.h
```diff
@@ -0,0 +1,48 @@
...
+    subscriber_->on_subscribe(this);
+
+    try {
+      (*function_)(*subscriber_);
+      subscriber_->on_complete();
+    } catch (...) {
+      subscriber_->on_error(std::current_exception());
+    }
+
+    delete this;
```
lehecka: with this memory management model, however with good intentions, we will end up in a SEGFAULT hell. 
We should let shared_ptrs and unique_ptrs do the memory management job. There should be no new or delete in the code (and never "delete this").

experimental/vjn/yarpl/include/yarpl/flowable.h
```diff
@@ -0,0 +1,85 @@
...
+class Flowable : public reactivestreams::Publisher<T>,
+                 public std::enable_shared_from_this<Flowable<T>> {
+  using Subscriber = reactivestreams::Subscriber<T>;
+  using Subscription = reactivestreams::Subscription<T>;
+
+ public:
+  template <typename F,
+            typename = typename std::enable_if<
+                std::is_callable<F(Subscriber&), void>::value>::type>
+  static std::shared_ptr<Flowable> create(F&& function) {
```
lehecka: to implement proper perfect forwarding, the parameter should be of type F and then std::forward is appropriate.

experimental/vjn/yarpl/include/yarpl/operators/map.h
```diff
@@ -0,0 +1,69 @@
...
+class Mapper : public Flowable<T>, public ::reactivestreams::Subscriber<T> {
+  using Flowable = ::yarpl::Flowable<T>;
+  using Subscriber = ::reactivestreams::Subscriber<T>;
+  using Subscription = ::reactivestreams::Subscription<T>;
+
+ public:
+  Mapper(Transform&& transform, std::shared_ptr<Flowable> upstream)
+      : forwarder_(
+            std::make_unique<Forwarder>(std::forward<Transform>(transform))),
+        upstream_(upstream) {}
```
lehecka: nit: upstream_(std::move(upstream))

experimental/vjn/yarpl/main/main.cc
```diff
@@ -0,0 +1,80 @@
...
+
+using reactivestreams::Subscriber;
+
+auto make_subscriber(int state) {
+  return Subscriber<int>::create(
+      [state](int i) mutable {
+        std::cout << "  state: " << ++state << ", i = " << i << std::endl;
+      },
+      [](const std::exception_ptr eptr) {
+        try {
```
lehecka: to avoid expensive try ... catch .. rethrow .. observe type, folly provides folly::exception_wrapper which does the same thing, but much cheaper.

experimental/vjn/yarpl/main/main.cc
```diff
@@ -0,0 +1,80 @@
...
+        } catch (const std::exception& exception) {
+          std::cerr << "  exception: " << exception.what() << std::endl;
+        } catch (...) {
+          std::cerr << "  !unknown exception!" << std::endl;
+        }
+      },
+      []() { std::cout << "  complete." << std::endl; });
+}
+
+class Scheduler : public yarpl::Scheduler {
```
lehecka: can we name the Scheduler such that it represents that for each task it spins a new thread?

experimental/vjn/yarpl/main/main.cc
```diff
@@ -0,0 +1,80 @@
...
+ public:
+  virtual ~Scheduler() = default;
```
lehecka: unnecessary, it derives such thing from the base class

experimental/vjn/yarpl/main/main.cc
```diff
@@ -0,0 +1,80 @@
...
+  virtual void schedule(std::function<void()>&& action) {
```
lehecka: nit: no virtual, add override

benjchristensen:
I've got a PR working on top of this with unit tests etc. I'll try and apply the feedback.

vjn:
Ben's working on (an expansion of) these changes in #323

PR #323, opened 2017-03-14
Yarpl (for review only)

benjchristensen:
Do not merge. Just want to share and get feedback.


experimental/yarpl/examples/FlowableExamples.cpp
```diff
@@ -1,70 +1,61 @@
...
-    void cancel() {}
-    void request(uint64_t n) {
-      auto sp = s.lock();
-      if (sp) {
-        // do stuff
-        sp->onNext(1);
+
+   private:
+    Subscription* s_;
+    int requested{0};
```
vjn: requested_

experimental/yarpl/examples/FlowableExamples.cpp
```diff
@@ -1,70 +1,61 @@
 // Copyright 2004-present Facebook. All Rights Reserved.
 
-#include "examples/FlowableExamples.h"
```
lehecka: this might bring some issues in fbcode. You dont like "examples/..." path?

experimental/yarpl/examples/FlowableExamples.cpp
```diff
@@ -1,70 +1,61 @@
...
+#include "FlowableExamples.h"
+#include <thread>
+#include "reactivestreams/ReactiveStreams.h"
+#include "yarpl/Flowable.h"
+
+using namespace reactivestreams_yarpl;
+using namespace yarpl::flowable;
+
+using namespace reactivestreams_yarpl;
```
lehecka: duplicate

experimental/yarpl/examples/FlowableExamples.cpp
```diff
@@ -1,70 +1,61 @@
...
-    };
-    void cancel() {}
-    void request(uint64_t n) {
-      auto sp = s.lock();
-      if (sp) {
-        // do stuff
-        sp->onNext(1);
+
+   private:
+    Subscription* s_;
```
lehecka: nit: the member variables typically belongs to the bottom under all private methods

experimental/yarpl/examples/FlowableExamples.cpp
```diff
@@ -1,70 +1,61 @@
...
+      requested = 10;
+      s_->request(10);
     }
-    ~MySubscriber() {
-      std::cout << "MySubscriber DESTROYED" << std::endl;
-    }
-    void onNext(const int& value) {
-      std::cout << "  onNext received " << value << std::endl;
+
+    void onNext(const long& t) {
```
lehecka: override keyword for all overrides

experimental/yarpl/include/reactivestreams/ReactiveStreams.h
```diff
@@ -0,0 +1,39 @@
...
+  virtual void request(int64_t n) = 0;
+  virtual void cancel() = 0;
+};
+
+template <typename T>
+class Subscriber {
+ public:
+  virtual ~Subscriber() = default;
+  virtual void onSubscribe(Subscription*) = 0;
+  virtual void onNext(const T&) = 0;
```
lehecka: I don't think having 2 overrides of onNext is a lucky choice.
We should just have one override onNext(T value). It will allow the implementation to use std::move to avoid copying large T and also will be simple to use.

experimental/yarpl/include/reactivestreams/ReactiveStreams.h
```diff
@@ -0,0 +1,39 @@
...
+  virtual void onNext(T&& value) {
+    onNext(value);
+  }
+  virtual void onComplete() = 0;
+  virtual void onError(const std::exception_ptr error) {
```
lehecka: with std::exception_ptr we are introducing performance hit for the users right away. Throwing exceptions is expensive in C++ and at the same you can't get exception.what() or exception type without throwing.
I suggest using folly::exception_wrapper which is designed exactly to solve this problem.

experimental/yarpl/include/reactivestreams/ReactiveStreams.h
```diff
@@ -0,0 +1,39 @@
...
+  virtual ~Subscription() = default;
+  virtual void request(int64_t n) = 0;
+  virtual void cancel() = 0;
+};
+
+template <typename T>
+class Subscriber {
+ public:
+  virtual ~Subscriber() = default;
+  virtual void onSubscribe(Subscription*) = 0;
```
lehecka: Passing Subscription as raw pointer feels to me that we will have some memory issues sooner or later. It is hard to manage lifetime with raw pointers. I would suggest using unique_ptr<Subscription>. It will have some consequences to the application code, like it will be impossible to implement a class which implements for Subscriber and Subscription interfaces, but those might be just "healthy" consequences..

experimental/yarpl/include/yarpl/Flowable.h
```diff
@@ -2,48 +2,87 @@
...
+  using Subscription = reactivestreams_yarpl::Subscription;
+
  public:
-  explicit Flowable(
-      std::function<void(std::unique_ptr<Subscriber<T>>)> onSubscribe);
+  template <
+      typename F,
+      typename = typename std::enable_if<
+          std::is_callable<F(std::unique_ptr<Subscriber>), void>::value>::type>
+  static std::unique_ptr<Flowable> create(F&& onSubscribeFunc) {
```
lehecka: repeating same mistakes like in vjn's diff
You are implementing perfect forwarding here incorrectly: http://stackoverflow.com/questions/6829241/perfect-forwarding-whats-it-all-about

experimental/yarpl/include/yarpl/Flowable.h
```diff
@@ -2,48 +2,87 @@
...
+      typename F,
+      typename = typename std::enable_if<
+          std::is_callable<F(T), typename std::result_of<F(T)>::type>::value>::
+          type>
+  auto map(F&& function);
 
-  void subscribe(std::unique_ptr<Subscriber<T>>);
+  static std::unique_ptr<Flowable<long>> range(long start, long count) {
+    return create([start, count](auto subscriber) {
+      auto s_ = new yarpl::flowable::sources::RangeSubscription(
```
lehecka: nit: no underscores for local variables

experimental/yarpl/include/yarpl/Flowable.h
```diff
@@ -2,48 +2,87 @@
...
+  Flowable& operator=(Flowable&&) = delete;
+  Flowable& operator=(const Flowable&) = delete;
 
  private:
-  std::function<void(std::unique_ptr<Subscriber<T>>)> onSubscribe_;
+  template <typename Function>
+  class FlowableOnSubscribeFunc : public Flowable {
+   public:
+    FlowableOnSubscribeFunc(Function&& function)
+        : function_(Function(std::forward<Function>(function))) {}
```
lehecka: wrong use of std::forward, std::move should be used

experimental/yarpl/include/yarpl/Flowable_TestSubscriber.h
```diff
@@ -0,0 +1,297 @@
...
+  /**
+   * Create a TestSubscriber that will delegate all on* method calls
+   * to the provided Subscriber.
+   *
+   * This will store all values it receives to allow assertions.
+   * @return
+   */
+  static std::shared_ptr<TestSubscriber<T>> create(std::unique_ptr<Subscriber>);
+
+  TestSubscriber(std::unique_ptr<Subscriber> delegate);
```
lehecka: explicit keyword

experimental/yarpl/include/yarpl/Flowable_TestSubscriber.h
```diff
@@ -0,0 +1,297 @@
...
+  TestSubscriber(long initialRequestN);
```
lehecka: explicit keyword

experimental/yarpl/src/yarpl/flowable/operators/Flowable_Map.h
```diff
@@ -0,0 +1,64 @@
...
+class Mapper : public flowable::Flowable<T>,
+               public reactivestreams_yarpl::Subscriber<T> {
+  using Flowable = flowable::Flowable<T>;
+  using Subscriber = reactivestreams_yarpl::Subscriber<T>;
+  using Subscription = reactivestreams_yarpl::Subscription;
+
+ public:
+  Mapper(Transform&& transform, std::shared_ptr<Flowable> upstream)
+      : forwarder_(
+            std::make_unique<Forwarder>(std::forward<Transform>(transform))),
```
lehecka: std::forward -> std::move

experimental/yarpl/src/yarpl/flowable/sources/Flowable_RangeSubscription.cpp
```diff
@@ -0,0 +1,102 @@
...
+      tryDelete();
+      return;
+    }
+  }
+}
+
+/** do the actual work of emission */
+int64_t RangeSubscription::tryEmit() {
+  int64_t consumed{0};
+  int64_t r{0};
```
lehecka: nit: you can declare r in the for( auto r = reuqested_.load() ; ... ) init section

experimental/yarpl/src/yarpl/flowable/utils/SubscriptionHelper.cpp
```diff
@@ -0,0 +1,82 @@
...
+  for (;;) {
+    int64_t r = current->load();
+    if (r == INT64_MAX) {
+      return INT64_MAX;
+    }
+    if (n <= 0) {
+      // do nothing, return existing unmodified value
+      return r;
+    }
+    long u = r + n;
```
lehecka: long type will be 32 bit number on 32 bit architecture. You probably want to be more specific and use int64_t or "long long"

experimental/yarpl/src/yarpl/flowable/utils/SubscriptionHelper.cpp
```diff
@@ -0,0 +1,82 @@
...
+int64_t SubscriptionHelper::consumeCredits(
+    std::atomic<std::int64_t>* current,
+    int64_t n) {
+  for (;;) {
+    int64_t r = current->load();
+    if (n <= 0) {
+      // do nothing, return existing unmodified value
+      return r;
+    }
+    long u = r - n;
```
lehecka: same here

experimental/yarpl/src/yarpl/flowable/utils/SubscriptionHelper.cpp
```diff
@@ -0,0 +1,82 @@
...
+    if (u < 0) {
+      // bad usage somewhere ... be resilient, just set to 0
```
lehecka: should we add LOG(ERROR) << "..." here? or DCHECK?

experimental/yarpl/include/reactivestreams/ReactiveStreams.h
lehecka: I guess you are trying to avoid dependency to folly.
Is there an equivalent in boost?

vjn: Ben and I discussed the pros/cons of exception_ptr, exception_wrapper, etc. There is no equivalent in boost:: that does what exception_wrapper does: hold an exception without it ever having been thrown.

experimental/yarpl/include/reactivestreams/ReactiveStreams.h
vjn: Should we assume that value is always movable, and will always be moved from? (We'll have a chain of onNext() calls down a pipeline.)

experimental/yarpl/src/yarpl/flowable/utils/SubscriptionHelper.cpp
vjn: Perhaps just an assert(), to avoid glog dependency.

experimental/yarpl/include/reactivestreams/ReactiveStreams.h
vjn: In the general case, Subscription owns Subscriber, and will delete it when it finishes. (This is a consequence of cancel() not providing a notification to the subscriber, per the core spec. Thus, only the subscription knows when to clean up, never the subscriber.) If a class that has both interfaces is defined, it should manage the lifetime the same way: the subscription manages the cleanup; if there is no separate subscriber object, there's nothing to delete.

experimental/yarpl/examples/FlowableExamples.cpp
benjchristensen: Is this better?

> #include "experimental/yarpl/examples/FlowableExamples.h"

experimental/yarpl/include/reactivestreams/ReactiveStreams.h
benjchristensen: Documented as:

```
  /**
   * Receive the Subscription that controls the lifecycle of this
   * Subscriber/Subscription pair.
   *
   * This Subscription* will exist until onComplete, onError, or cancellation.
   *
   * Rules:
   * - Do NOT use Subscription* after onComplete();
   * - Do NOT use Subscription* after onError();
   * - Do NOT use Subscription* after calling Subscription*->cancel();
   */
  virtual void onSubscribe(Subscription*) = 0;
```

Having been working with this for a few days, it actually is pretty nice working with this knowing that only the `Subscription` concerns itself with lifecycle.

experimental/yarpl/include/reactivestreams/ReactiveStreams.h
benjchristensen: Here's how it looks implementing an operator right now:

```cpp
  void onNext(const T& t) {
    downstream_->onNext((*f_)(t));
  }

  void onNext(T&& t) {
    downstream_->onNext((*f_)(std::move(t)));
  }
```

This works and correctly calls either the lvalue or rvalue version depending on what I pass in from the top.

experimental/yarpl/examples/FlowableExamples.cpp
benjchristensen: Done.

experimental/yarpl/examples/FlowableExamples.cpp
benjchristensen: Done.

experimental/yarpl/examples/FlowableExamples.cpp
benjchristensen: Removed.

experimental/yarpl/include/yarpl/Flowable.h
benjchristensen: I apologize, I'm kind of slow ... read through that, and re-read Item 25 about perfect forwarding in "Effective Modern C++", but am not understanding what I'm doing wrong here. 

What needs to change?

Here is the current method:

```cpp
template <
    typename T,
    typename F,
    typename = typename std::enable_if<std::is_callable<
        F(std::unique_ptr<reactivestreams_yarpl::Subscriber<T>>),
        void>::value>::type>
static auto unsafeCreateUniqueFlowable(F&& onSubscribeFunc) {
  return std::make_unique<UniqueFlowable<T, F>>(
      std::forward<F>(onSubscribeFunc));
}
```

experimental/yarpl/include/yarpl/Flowable.h
benjchristensen: Fixed

experimental/yarpl/include/yarpl/Flowable.h
benjchristensen: If I understand correctly, because this is not a universal reference, despite looking like one, because the template is defined at the class level?

experimental/yarpl/include/yarpl/Flowable_TestSubscriber.h
benjchristensen: Does this have to be done whenever there is a single argument constructor?

experimental/yarpl/include/yarpl/Flowable_TestSubscriber.h
benjchristensen: fixed.

experimental/yarpl/src/yarpl/flowable/sources/Flowable_RangeSubscription.cpp
benjchristensen: Done.

experimental/yarpl/src/yarpl/flowable/utils/SubscriptionHelper.cpp
benjchristensen: Good point. Fixed.

experimental/yarpl/src/yarpl/flowable/utils/SubscriptionHelper.cpp
benjchristensen: Fixed.

experimental/yarpl/src/yarpl/flowable/utils/SubscriptionHelper.cpp
benjchristensen: Doesn't assert() cause the program to terminate? That would defeat my goal of being resilient and continuing to work. 

DCHECK fails the app only in debug mode, right? I'd be okay with that.

benjchristensen:
Code like this can now be written:

```cpp
  Flowable::range(1, 100)
      ->map([](auto i) { return "hello->" + std::to_string(i); })
      ->take(10)
      ->subscribe(createSubscriber<std::string>(
          [](auto t) { std::cout << "Value received: " << t << std::endl; }));
```

A key change in this code since yesterday occurred as a result of discussions with AlexM. Flowable is now split into `Flowable` purely with factory methods, and then `UniqueFlowable` which contains `lift` on it, and works without shared pointers. When chaining it steals the function from the predecessor by moving it from one to the next. 

This seems to work well, and should be efficient for the majority of cases. 

It does have one problem left that I haven't yet solved ... it's template type is <T, Function> and I have no idea how to expose this in public APIs, such as:

```cpp
std::unique_ptr<UniqueFlowable<T, ???>> getData()
```

I would like a cover type:

```cpp
std::unique_ptr<Flowable<T>> getData()
or
std::unique_ptr<UFlowable<T>> getData()
```

but haven't yet figured out how to get something to work.

benjchristensen:
A problem I see .. code that shouldn't work but is, which means I don't understand something:

```cpp
  auto a = Flowable::range(1, 100);
  auto b = a->take(10);
  auto c = b->map([](auto i) { return "hello->" + std::to_string(i); });

  c->subscribe(createSubscriber<std::string>(
      [](auto t) { std::cout << "Value received: " << t << std::endl; }));

  // this should not work, but it does
  auto c2 = b->map([](auto i) { return "should break->" + std::to_string(i); });
  c2->subscribe(createSubscriber<std::string>(
      [](auto t) { std::cout << "Value received: " << t << std::endl; }));
```

The `Flowable.lift` process std::moves the `function_` from one Flowable to the next, so it should *not* work to lift from the same `Flowable` twice as the code above does.

benjchristensen:
> it should not work to lift from the same Flowable twice as the code above does.

This means to me that the function is being copied from Flowable to Flowable.

benjchristensen:
> This means to me that the function is being copied from Flowable to Flowable.

Yes, confirmed this by deleting the copy constructors on an operator and it stops working. 

I'll start playing with this more, probably no point in you reviewing until I dig deeper, unless you can help suggest how to change `UniqueFlowable` and `lift`.

benjchristensen:
Oh, and I would like to end up with a `SharedFlowable` as well, similar to how there is `std::future` and `std::shared_future` so if a single `Flowable` is to be branched in multiple composition chains, it would be converted so someone opts-in to the use of `shared_ptr` or whatever.

```cpp
aUniqueFlowable.make_shared()
```

lexs:
I haven't dug very deep into how it's implemented so making some very broad assumptions.

You can specify different overloads for lift depending on if it it is called on an lvalue or rvalue:
lvalue:
```
auto lift(F&& onSubscribeLift);
```
rvalue:
```
auto lift(F&& onSubscribeLift) &&;
```

This way I suspect you could automatically convert internal state if you need to copy it.

benjchristensen:
> You can specify different overloads for lift depending on if it it is called on an lvalue or rvalue:
lvalue:

Yes I'm interested in exploring that, and have added a TODO with comments now based on you bringing this up. 

I don't yet know what I'd do differently inside lift because of these reasons:

- Should we require all OnSubscribeFunctions to support both copy and move? copying could be bad, for example when a Flowable represents network behavior. Or in that case would it just blow up?
- If copying wasn't used, then we could try and use a shared_ptr ... but going from `UniqueFlowable` to a shared_ptr is far from clear from the inside of that object, as we specifically don't want "enabled_shared_from_this" as that requires use of shared_ptr in all cases (if I understand it correctly).

I'll play with this more to see how it works when some operators support move and copy, and others only move, and if that would enforce itself elegantly or not.

benjchristensen:
Also, it would require us returning a `UniqueFlowable`, not `std::unique_ptr<UniqueFlowable>` if I understand correctly.

lexs:
What the reason for having them be `unique_ptr`s? Seem like keeping them as rvalues would work as well? You're right that rvalue qualifiers won't work on pointers.

Perhaps you should even disallow calling lift on lvalues and require an explicit call to some "copy()" method that moves the internal state into a shared_ptr.

lexs:
This copy would also probably encode the Function into folly::Function or std::function so it's no longer part of the type (and can easily be passed around).

benjchristensen:
> Seem like keeping them as rvalues would work as well?

I think I agree. I will play with that shortly.

> This copy would also probably encode the Function into folly::Function or std::function so it's no longer part of the type (and can easily be passed around).

I don't understand this comment. What do you mean? The code doesn't use std::function and folly isn't used anywhere in this.

benjchristensen:
Just pushed more code that allows this with `subscribeOn`:

```cpp
  ThreadScheduler scheduler;

  Flowable::range(1, 10)
      ->subscribeOn(scheduler) // put on background thread
      ->map([](auto i) { return "Value received: " + std::to_string(i); })
      ->take(6)
      ->subscribe(createSubscriber<std::string>([](auto t) {
        std::cout << t << " on thread: " << std::this_thread::get_id()
                  << std::endl;
      }));
```

This outputs:

```
Main Thread ID 0x7fffa05203c0
Value received: 1 on thread: 0x70000b6e0000
Value received: 2 on thread: 0x70000b6e0000
Value received: 3 on thread: 0x70000b6e0000
Value received: 4 on thread: 0x70000b6e0000
Value received: 5 on thread: 0x70000b6e0000
Value received: 6 on thread: 0x70000b6e0000
```

benjchristensen:
@lexs I've been messing with the use of ref-qualifiers and can't figure out how to get it working. My logs show it always chooses the lvalue version, whereas on the following code I expect the first to use lvalue, the second to use rvalue:

```cpp
    auto a = Flowable::range(1, 10);
    auto b = a.take(3);
    auto c = b.map([](auto i) { return "hello->" + std::to_string(i); });
    c.subscribe(createSubscriber<std::string>(
            [](auto t) { std::cout << "Value received: " << t << std::endl; }));

  Flowable::range(1, 10)
      .map([](auto i) { return "hello->" + std::to_string(i); })
      .take(3)
      .subscribe(createSubscriber<std::string>(
          [](auto t) { std::cout << "Value received: " << t << std::endl; }));
```

I'm trying ref-qualifers like this:

```cpp
  auto take(int64_t toTake) && {
    return lift<T>(yarpl::operators::FlowableTakeOperator<T>(toTake));
  };

  auto take(int64_t toTake) & {
    return lift<T>(yarpl::operators::FlowableTakeOperator<T>(toTake));
  };
```
... and similar for `lift` (which is too gnarly to paste here) and that's where my logging was.

Example of what I've been reading: http://www.jeremyong.com/blog/2014/02/15/using-ref-qualifiers/

Giving up on this one for a bit :-/

benjchristensen:
Removing the return type of `unique_ptr` ends up with code like this:

```cpp
  ThreadScheduler scheduler;

  Flowable::range(1, 10)
      .subscribeOn(scheduler) // put on background thread
      .map([](auto i) { return "Value received: " + std::to_string(i); })
      .take(6)
      .subscribe(createSubscriber<std::string>([](auto t) {
        std::cout << t << " on thread: " << std::this_thread::get_id()
                  << std::endl;
      }));
```

and

```cpp
  auto a = Flowable::range(1, 10);
  auto b = a.take(3);
  auto c = b.map([](auto i) { return "hello->" + std::to_string(i); });
  c.subscribe(createSubscriber<std::string>(
      [](auto t) { std::cout << "Value received: " << t << std::endl; }));
```

benjchristensen:
Question, how do I get a reasonable return type on method signatures when we're using template functions as we are?

Our type signature is currently templated as `UniqueFlowable<T, Function>`, but we don't know `Function`, nor should a method like `getData()` care.

So we end up like this:

```cpp
auto getDataAsFlowable() {
  return Flowable::range(1, 4);
}
```

This is *very* unlikable for public APIs.

An alternative is we return a `Publisher` type, wrapped in a unique_ptr since it's an abstract class and can't be returned directly:

```cpp
std::unique_ptr<reactivestreams_yarpl::Publisher<long>> getDataAsPublisher() {
    return Flowable::range(1, 4).as_unique_ptr();
}
```

but then we must convert from `Publisher` to `Flowable` in order to chain off of it, like this:

```cpp
  Flowable::fromPublisher(getDataAsPublisher())
      .take(3)
      .subscribe(subscriber);
```

That's ugly, and inefficient.

And I can't create a parent class with virtual methods containing map, take, etc ... because templates.

Ideas?

lexs:
Ignoring the rvalue stuff, I'll look into that later.

People able to return Flowables is obviously very important but we also don't want to go ahead and use std::function everywhere because of perf. A good compromise seems to make `UniqueFlowable<T, Function>` implicitly convertible to `Flowable<T>` that instead stores the `Function` in an `std::function` internally.

benjchristensen:
> A good compromise seems to make UniqueFlowable<T, Function> implicitly convertible to Flowable<T> that instead stores the Function in an std::function internally.

But doesn't that mean that every time we have an API (such as RSocket) we pay the cost of std::function? That would mean we have std::function involved most of the time.

lexs:
> But doesn't that mean that every time we have an API (such as RSocket) we pay the cost of std::function?

Indeed it does. Not sure how you can design around that. cc @jspahrsummers

lexs:
Returning a `unique_ptr` has the same cost as using `std::function` btw. `std::function` might be cheaper as it can inline store some lambdas.

benjchristensen:
I'd like to get this code merged in, as then I can do jemalloc experiments on the proper hardware, and can get real numbers on the memory usage of different options.

benjchristensen:
How about I rebase/squash this massive thing, go through the merge process, and then memory test different options with fbsource tooling?

lexs:
That sounds fine to me, I'd like to play with this too.

benjchristensen:
Replacing this with https://github.com/ReactiveSocket/reactivesocket-cpp/pull/326

PR #324, opened 2017-03-15
Add tck tests to travis

somasun:


somasun:
Currently we run only cpp-cpp and java-java tests.  Once v1 is made default, we should be able to run cpp-java combinations.  It will involve just adding another line in .travis.yml

lehecka:
this should be an awesome validation step for all our implementations. Good work!

PR #326, opened 2017-03-17
YARPL - Flowable (experimental)

benjchristensen:
Squashed and rebased version of code from https://github.com/ReactiveSocket/reactivesocket-cpp/pull/323

Intended for merging for broader collaboration.

Still significant open questions as shown in #323.


lexs:
I just played around with this with the goal of being able to define a Flowable return type without having to use auto. Came up with this which is pretty simple, based it on `FlowableA`:

```c++
template <typename T>
using SimpleFlowable =
  FlowableA<T, std::function<void(std::unique_ptr<reactivestreams_yarpl::Subscriber<T>>)>>;

template <typename T>
class Flowable : public SimpleFlowable<T> {
public:
  template <typename Function>
  Flowable(FlowableA<T, Function>&& flowable)
    : SimpleFlowable<T>(std::move(flowable.function_)) {
  }
};
```

This makes it possible to have code like this that inside the function is 100% allocation free but is implicitly converted to a std::function when being returned:
```c++
Flowable<std::string> helloWorld(long n) {
  return Flowables::range(1, n) // This is still FlowableA<T, Function>
      .map([](auto i) { return "hello->" + std::to_string(i); });
}
```

benjchristensen:
> is implicitly converted to a std::function when being returned

Wouldn't that cause use of std::function in the majority of `Flowable` instances in a system? For example, the RSocket API would have a method like this:

```
Flowable<Payload> requestStream(Payload request)
```

In that case, every `requestStream` usage would return a `SimpleFlowable` that uses std::function, and our goal so far has been to not use that.

lexs:
It would cause `std::function` in public APIs yes, but I don't see any way around that if you want it to be an interface and not be templated.

benjchristensen:
Latest commit is from discussions @vjn and I are having. Currently there are 4 variants:

- Flowable.h that uses smart_ptr
- FlowableA.h that is stack allocated, and uses a templated function which gives an awkward public API, but can be converted to a unique_ptr for returning from a method with a reasonable API
- FlowableB.h that is stack allocated, and uses std::function which forces an OnSubscribe to be copyable, not just moveable
- FlowableC.h that uses only unique_ptr, and isn't completely usable since it can destroy an OnSubscribe across method boundaries

We have also been looking at use of the Boost intrusive_ptr.

benjchristensen:
Flowable.h allows code like this:

```cpp
std::shared_ptr<Flowable<std::string>> getFlowable() {
  return Flowables::range(1, 5)->map(
      [](auto i) { return "Data=>" + std::to_string(i); });
}

  getFlowable()->take(2)->subscribe(createSubscriber<std::string>(
      [](auto t) { std::cout << "Value received: " << t << std::endl; }));

  Flowables::range(1, 10)
      ->subscribeOn(scheduler) // put on background thread
      ->map([](auto i) { return "Value received: " + std::to_string(i); })
      ->take(6)
      ->subscribe(createSubscriber<std::string>([](auto t) {
        std::cout << t << " on thread: " << std::this_thread::get_id()
                  << std::endl;
      }));

```

benjchristensen:
FlowableA.h allows code like this:

```cpp
auto getFlowableA() {
  return FlowablesA::range(1, 5)
      .map([](auto i) { return "Data=>" + std::to_string(i); });
}

  getFlowableA().take(2).subscribe(createSubscriber<std::string>(
      [](auto t) { std::cout << "Value received: " << t << std::endl; }));


  FlowablesA::range(1, 10)
      .subscribeOn(scheduler) // put on background thread
      .map([](auto i) { return "Value received: " + std::to_string(i); })
      .take(6)
      .subscribe(createSubscriber<std::string>([](auto t) {
        std::cout << t << " on thread: " << std::this_thread::get_id()
                  << std::endl;
      }));
```

benjchristensen:
FlowableB.h allows code like this: 

```cpp
FlowableB<std::string> getFlowableB() {
    return FlowablesB::range(1, 5)
            .map([](auto i) { return "Data=>" + std::to_string(i); });
}

    getFlowableB().take(2).subscribe(createSubscriber<std::string>(
            [](auto t) { std::cout << "Value received: " << t << std::endl; }));


    FlowablesB::range(1, 10)
            .subscribeOn(scheduler) // put on background thread
            .map([](auto i) { return "Value received: " + std::to_string(i); })
            .take(6)
            .subscribe(createSubscriber<std::string>([](auto t) {
                std::cout << t << " on thread: " << std::this_thread::get_id()
                          << std::endl;
            }));
```

PR #327, opened 2017-03-21
Removing frame serialization from stream automata

lehecka:
This is a half of the effort to remove dependency on *_Frame types from the stream automata.
The automatons will call virtual methods to write on the wire. Also in the next change they will not parse frames. The framing will not be exposed outside of the ConnectionAutomaton.

Tested with unit tests.


src/StreamsHandler.h
```diff
@@ -0,0 +1,35 @@
...
+
+class StreamsWriter {
+ public:
+  virtual ~StreamsWriter() = default;
+  virtual void writeNewStream(
+      StreamId streamId,
+      StreamType streamType,
+      uint32_t initialRequestN,
+      Payload payload,
+      bool TEMP_completed) = 0;
```
somasun: curious. any reason for following different convention in naming ?

src/ConnectionAutomaton.h
```diff
@@ -246,6 +248,21 @@ class ConnectionAutomaton final
   void pauseStreams();
   void resumeStreams();
 
+  void writeNewStream(
+      StreamId streamId,
+      StreamType streamType,
+      uint32_t initialRequestN,
+      Payload payload,
+      bool TEMP_completed) override;
```
somasun: ditto

src/ConnectionAutomaton.cpp
```diff
@@ -764,4 +765,99 @@ FrameSerializer& ConnectionAutomaton::frameSerializer() const {
   return *frameSerializer_;
 }
 
+void ConnectionAutomaton::writeNewStream(
+    StreamId streamId,
+    StreamType streamType,
+    uint32_t initialRequestN,
+    Payload payload,
+    bool TEMP_completed) {
```
somasun: ditto

src/Frame.cpp
```diff
@@ -154,18 +154,17 @@ Frame_ERROR Frame_ERROR::connectionError(const std::string& message) {
   return Frame_ERROR(0, ErrorCode::CONNECTION_ERROR, Payload(message));
 }
 
-Frame_ERROR Frame_ERROR::invalid(
-    StreamId streamId,
-    const std::string& message) {
+Frame_ERROR Frame_ERROR::error(StreamId streamId, Payload&& payload) {
   DCHECK(streamId) << "streamId MUST be non-0";
-  return Frame_ERROR(streamId, ErrorCode::INVALID, Payload(message));
+  return Frame_ERROR(streamId, ErrorCode::INVALID, std::move(payload));
```
somasun: curious. Do we want all onError wire-writes to be of ErrorCode::INVALID ?

src/StreamsHandler.h
lehecka: this is just because the parameter is temporary and will be remove in the next 2 diffs.
I wanted to make it visible.

src/Frame.cpp
lehecka: INVALID is the general error code that something was wrong.
For application errors we are using APPLICATION_ERROR

PR #332, opened 2017-03-29
Benchmarks

tmontgomery:
Beginnings of some benchmarks.

Using the experimental API. Building can be disabled via CMake option (BUILD_BENCHMARKS) which defaults to ON. But can be set to OFF.

Comments welcome.


benchmarks/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
+
+##########################################################
+# Google benchmark
+
+ExternalProject_Add(
+    google_benchmark
```
lexs: How is google_benchmark different from follys, any advantages? Would be nice to not add another dep if possible.

https://github.com/facebook/folly/blob/master/folly/docs/Benchmark.md

benchmarks/RequestResponseLatency.cpp
```diff
@@ -0,0 +1,228 @@
...
+class BM_RequestHandler : public DefaultRequestHandler
+{
+public:
+    void handleRequestResponse(
+        Payload request, StreamId streamId, const std::shared_ptr<Subscriber<Payload>> &response) noexcept override
+    {
+        LOG(INFO) << "BM_RequestHandler.handleRequestResponse " << request;
+
+        const char* p = reinterpret_cast<const char*>(request.data->data());
+        auto requestString = std::string(p, request.data->length());
```
lexs: you could use `request.data->moveToFbString().toStdString()` here instead

benchmarks/RequestResponseLatency.cpp
```diff
@@ -0,0 +1,228 @@
...
+{
+public:
+    BM_RsFixture() :
+        host_(FLAGS_host),
+        port_(FLAGS_port),
+        serverRs_(RSocket::createServer(TcpConnectionAcceptor::create(port_))),
+        handler_(std::make_shared<BM_RequestHandler>())
+    {
+        FLAGS_v = 0;
+        FLAGS_minloglevel = 6;
```
lexs: Does this make the LOG(INFO)s not log? They should probably just be VLOGs to begin with which would let us pass -v on the commandline instead.

benchmarks/RequestResponseLatency.cpp
```diff
@@ -0,0 +1,228 @@
...
+        serverRs_->start([this](auto r) { return handler_; });
+    }
+
+    virtual ~BM_RsFixture()
+    {
+    }
+
+    virtual void SetUp(benchmark::State &state)
```
lexs: -`virtual` +`override`?
same below

benchmarks/CMakeLists.txt
tmontgomery: I would agree on not needing another dep. Folly is already a big dep. However, the use of GB is conditional on the benchmarks being built and will only be brought in when needed. And is self contained in the build via the snapshot.

- Cosmetically, GB and Folly are close. But the setup/teardown are not the same. Googles' is much more mature, IMO.
- Measurement technique is similar. However, GB is a closer fit to getting a more relevant benchmark with its sampling technique and run to min time approach.
- Custom counters are relied upon. Not seeing how those are integrated with folly.
- harness should be self contained and not rely on external components. Especially ones that are in use by the unit under test.

benchmarks/RequestResponseLatency.cpp
tmontgomery: Could yes. This is the critical path for that benchmark, though. So, it should be simple and quick. Not sure which call is going to add noise to the measurement. And which one could be optimized best by the compiler. So, I went with one I know can normally be very efficient.

benchmarks/RequestResponseLatency.cpp
tmontgomery: Benchmarks should not log. Thus setting them by hand here to emphasize that.

benchmarks/RequestResponseLatency.cpp
tmontgomery: Better to make them go away as I believe they are null... they are only left for those new to GB and perhaps wanting to add state. They are out of the measurement path, so they don't hurt. If stylistically, `override` is preferred, go for it.

tmontgomery:
Duplication is intentional. These are self contained benchmarks as much as possible. One can be changed without, hopefully, changing the others.

In terms of style, yes, can run clang-format.

tmontgomery:
A better way to manage the clang-format is to format the entire repo upon a merge. So, I will let you all do that once merging. Or if that is not OK, let me know and I can update once it is ready to merge.

benchmarks/CMakeLists.txt
lexs: (y)

benchmarks/RequestResponseLatency.cpp
lexs: Sure. I just realized you don't actually use `name_` anywhere. Why parse it at all then?

benchmarks/RequestResponseLatency.cpp
lexs: Agreed, just saying that not allowing turning logging on via command-line makes it more effort to debug the benchmarks when you need to.

benchmarks/RequestResponseLatency.cpp
lexs: Ok, yeah `override` is preferred, I believe we warn on it internally.

lexs:
I think it's better to run a clang-format before merging so the impending repo-wide clang-format changes as little as possible.

Is this ready to merge otherwise? Could you post the output of the benchmarks, just curious how they look. I'd like @lehecka to take a look too.

benchmarks/RequestResponseLatency.cpp
tmontgomery: Was trying to keep things as close as possible to typical usage from the rsocket experimental code at the time. For this benchmark, it may not be relevant, though.

tmontgomery:
I don't have changes planned. If ready to merge, I can format it.

benchmarks/RequestResponseLatency.cpp
lexs: It might also be optimized out (it probably doesn't happen). The raw tcp benchmark doesn't construct an std::string, it only does one copy which is already done here. IMO they should do the same amount of work in the "user code".

lehecka:
pulling this in. We will reformat it in another iteration.

tmontgomery:
Just a capture of current benchmark output. I removed the request name and removed the 64 request run for reqrespthroughput. And moved to override instead of virtual. The rest is as is.

Results are on Mac OS.

Baseline
```
Run on (8 X 2900 MHz CPU s)
2017-04-13 12:45:13
Benchmark                                   Time           CPU Iterations
-------------------------------------------------------------------------
BM_Baseline_TCP_Throughput/40/1024       8762 ns       3357 ns     216761   63.5628MB/s   1.58907M items/s
BM_Baseline_TCP_Throughput/40/4k         9478 ns       3529 ns     214601   64.3907MB/s   1.60977M items/s
BM_Baseline_TCP_Throughput/80/4k         9184 ns       3630 ns     199456   122.022MB/s   1.52527M items/s
BM_Baseline_TCP_Throughput/4k/4k         2267 ns       1741 ns     404330   2.19156GB/s    561.04k items/s
BM_Baseline_TCP_Latency/32              19351 ns       6524 ns     106706   4.67801MB/s   149.696k items/s
BM_Baseline_TCP_Latency/128             20120 ns       6724 ns     103871   18.1535MB/s   145.228k items/s
BM_Baseline_TCP_Latency/4k              20053 ns       6748 ns     100815   578.845MB/s   144.711k items/s
```
All the rsocket benchmarks have an abort at the end when the server goes out of scope. The rsocket server isn't cleaning up appropriately. But it does not interfere with the benchmarks. So, I will omit the output accept to illustrate it below.

```
Assertion failed: (eventBase_ == nullptr || eventBase_->isInEventBaseThread()), function stopAccepting, file io/async/AsyncServerSocket.cpp, line 186.
Abort trap: 6
```

Stream Throughput
```
Run on (8 X 2900 MHz CPU s)
2017-04-13 12:45:44
Benchmark                                      Time           CPU Iterations
----------------------------------------------------------------------------
BM_RsFixture/BM_Stream_Throughput/8          170 ns        147 ns    4613215   100.001k items/s Message Length: 32
BM_RsFixture/BM_Stream_Throughput/32         150 ns        137 ns    4763168   196.577k items/s Message Length: 32
BM_RsFixture/BM_Stream_Throughput/128        148 ns        134 ns    4920291   262.582k items/s Message Length: 32
```

Request/Response Throughput (I removed 64 in my branch because it is not any different)
```
Run on (8 X 2900 MHz CPU s)
2017-04-13 12:45:55
Benchmark                                              Time           CPU Iterations
------------------------------------------------------------------------------------
BM_RsFixture/BM_RequestResponse_Throughput/1       78519 ns      70740 ns       9385    13.805k items/s Max Requests: 1, Message Length: 32
BM_RsFixture/BM_RequestResponse_Throughput/2       43978 ns      38873 ns      17375   25.1218k items/s Max Requests: 2, Message Length: 32
BM_RsFixture/BM_RequestResponse_Throughput/8       26621 ns      23948 ns      28586   40.7789k items/s Max Requests: 8, Message Length: 32
BM_RsFixture/BM_RequestResponse_Throughput/16      23054 ns      21403 ns      27297    45.628k items/s Max Requests: 16, Message Length: 32
BM_RsFixture/BM_RequestResponse_Throughput/32      23261 ns      20891 ns      32884   46.7463k items/s Max Requests: 32, Message Length: 32
```

Request/Response Latency
```
Run on (8 X 2900 MHz CPU s)
2017-04-13 12:46:08
Benchmark                                        Time           CPU Iterations
------------------------------------------------------------------------------
BM_RsFixture/BM_RequestResponse_Latency      78720 ns      70628 ns       9636   13.8268k items/s Message Length: 32
```

lehecka:
Todd, I am unable to win the fight with the FB open source tooling / CLA to be able to land this PR to fbcode under your account. Would you be ok if I duplicate the PR under my name and land this?

lehecka:
awesome work btw. Its very helpful for us to have the benchmark.

tmontgomery:
@lehecka feel free to land it however is easiest.

lehecka:
closing in favor of https://github.com/ReactiveSocket/reactivesocket-cpp/pull/340

PR #338, opened 2017-04-11
Fix travis build

lehecka:


lehecka:
need to unblock everyone so landing now. 
I would appreciate any feedback anyways.

PR #348, opened 2017-04-22
RSocket Test Bootstrap

benjchristensen:
Getting started with unit tests for the new RSocket APIs.

The first test just starts a server and lets it shut down while leaving the contextual scope.

### DIFF
--- a/experimental/rsocket-test/RSocketClientServerTest.cpp
+++ b/experimental/rsocket-test/RSocketClientServerTest.cpp
@@ -1,0 +1,33 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#include <utility>
+
+#include <folly/io/IOBuf.h>
+#include <gmock/gmock.h>
+
+#include "handlers/HelloStreamRequestHandler.h"
+#include "rsocket/RSocket.h"
+#include "rsocket/RSocket.h"
+#include "rsocket/transports/TcpConnectionAcceptor.h"
+
+using namespace rsocket;
+using namespace rsocket::tests;
+
+TEST(RSocketClientServer, StartAndShutdown) {
+  {
+    TcpConnectionAcceptor::Options opts;
+    opts.port = 9889;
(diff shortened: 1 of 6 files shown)


experimental/rsocket-test/RSocketClientServerTest.cpp
```diff
@@ -0,0 +1,33 @@
...
+    auto rs = RSocket::createServer(
+        std::make_unique<TcpConnectionAcceptor>(std::move(opts)));
+
+    // global request handler
+    auto handler = std::make_shared<HelloStreamRequestHandler>();
+    // start server
+    rs->start([handler](auto r) { return handler; });
+  }
+  // everything should cleanly shut down here
+  // TODO anything to assert here?
```
benjchristensen: Is there anything I can assert here that doesn't require invasive use of destructors and counters to ensure everything was disposed correctly?

benjchristensen:
@lehecka @vjn A proposed start to unit testing the new RSocket APIs. If this is a reasonable setup of code path, CMake, etc ... I can proceed down this path. 

Also a question on how to assert shutdown/cleanup.

experimental/rsocket-test/RSocketClientServerTest.cpp
vjn: Within Yarpl, everything's (subscribers, flowables, subscriptions) a Refcounted<T>. There is a poor-man's counter of outstanding objects: static method Refcounted::objects(). I use this to assert that everything's cleaned up after a pipeline is created and run.

benjchristensen:
Rebased.

experimental/rsocket-test/RSocketClientServerTest.cpp
benjchristensen: Is that something you're doing just while validating everything, or do you plan on leaving that in the code?

vjn:
+1, let's merge this in github, and work out the fbcode sync later.

PR #350, opened 2017-04-22
RSocketRequester with Flowable (DO NOT MERGE)

benjchristensen:
Only seeking feedback, not ready to merge yet.


examples/stream-hello-world/StreamHelloWorld_Client.cpp
```diff
@@ -26,20 +28,22 @@ int main(int argc, char* argv[]) {
...
     LOG(INFO) << "------------------ Run in future.then";
-    auto s = std::make_shared<ExampleSubscriber>(5, 6);
+    auto s = yarpl::Reference<yarpl::Subscriber<Payload>>(
+        new ExampleSubscriber(5, 6));
     rsf->connect().then([s](std::shared_ptr<RSocketRequester> rs) {
-      rs->requestStream(Payload("Bob"), s);
+      rs->requestStream(Payload("Bob"))->subscribe(s);
     });
-    s->awaitTerminalEvent();
+    //    s->awaitTerminalEvent();
```
benjchristensen: I can't make `Reference<ExampleSubscriber>` and still pass in to the `subscribe(s)` call. 

What do I need to do so I can still have it be `ExampleSubscriber` and call methods such as `awaitTerminalEvent`?

My expectation is that `Reference` allows refCounting similar to `shared_ptr`. Am I misunderstanding? Does it actually behave like `unique_ptr` so I need to move it into `subscribe`?

experimental/rsocket-src/RSocketRequester.cpp
```diff
@@ -41,31 +43,29 @@ std::shared_ptr<Subscriber<Payload>> RSocketRequester::requestChannel(
...
-    std::shared_ptr<Subscriber<Payload>> responseSink) {
-  eventBase_.runInEventBaseThread(
-      [ this, request = std::move(request), responseSink ]() mutable {
-        standardReactiveSocket_->requestStream(
-            std::move(request), std::move(responseSink));
-      });
-}
+yarpl::Reference<yarpl::Flowable<Payload>> RSocketRequester::requestStream(
+    Payload request) {
+  auto flowable = yarpl::Flowable<Payload>::create([total = 0](
```
benjchristensen: I'm just returning a placeholder `Flowable` right now.

experimental/rsocket-src/RSocketRequester.cpp
```diff
@@ -41,31 +43,29 @@ std::shared_ptr<Subscriber<Payload>> RSocketRequester::requestChannel(
...
+    subscriber.onNext(Payload());
+    return std::make_tuple(int64_t{1}, false);
+  });
+  return flowable;
 
-std::shared_ptr<yarpl::flowable::Flowable<reactivesocket::Payload>>
-RSocketRequester::requestStream(reactivesocket::Payload request) {
-  auto& eb = eventBase_;
-  auto srs = standardReactiveSocket_;
-  return yarpl::flowable::Flowable<reactivesocket::Payload>::create(
```
benjchristensen: This is a place where we may need `Flowable::unsafeCreate` that exposes the raw behavior.

experimental/rsocket/OldNewBridge.h
```diff
@@ -6,15 +6,19 @@
...
+#include "yarpl/v/Subscription.h"
+
 #include "src/Payload.h"
 #include "src/ReactiveStreamsCompat.h"
 
 namespace rsocket {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-class NewToOldSubscription : public reactivestreams_yarpl::Subscription {
```
benjchristensen: Right now "v Flowable" has concrete types, no "Reactive Streams" types. Those will need to be extracted so RSocket can implement those types.

experimental/rsocket/RSocketRequester.h
```diff
@@ -41,18 +42,8 @@ class RSocketRequester {
...
-          responseSink);
-
-  /**
-   * TODO: This is a temporary hack to bridge instances of the old
-   * reactivesocket::Subscri{ber,ption} to
-   * reactivestreams_yarpl::Subscri{ber,ption}.
-   */
-  std::shared_ptr<yarpl::flowable::Flowable<reactivesocket::Payload>>
-  requestStream(reactivesocket::Payload request);
+  yarpl::Reference<yarpl::Flowable<reactivesocket::Payload>> requestStream(
```
benjchristensen: Does this look like the right signature now using `Flowable`?

examples/stream-hello-world/StreamHelloWorld_Client.cpp
vjn: No, your understanding is correct: In this case, you have "s", which you can hang onto, and invoke methods on. It is an additional reference.

experimental/rsocket/RSocketRequester.h
vjn: Yes, it does.

examples/stream-hello-world/StreamHelloWorld_Client.cpp
benjchristensen: Then what do I need to do for it to be Reference<ExampleSubscriber>? I couldn't get that to compile.

experimental/rsocket/OldNewBridge.h
vjn: Yes. There's some pain there since RS streams subscribe call has the wrong signature: subscribe(Subscription*). There are no decent ways to handle this case: every subclass ends up needing to implement this, or incur a warning for hidden superclass methods. (Which we can disable, but that seemed bad to me.)

experimental/rsocket-src/RSocketRequester.cpp
vjn: Could you spell this out: what are we unable to do, what is the "raw" behavior?

benjchristensen: Here since the underlying RSocket code just takes a `Subscriber` and does the `Subscriber`/`Subscription` behavior itself, I just need to pass the `Subscriber` into it ... after awkwardly scheduling the subscribe onto an eventBase. 

This code effectively:

```cpp
    auto& eb = eventBase_;
    auto srs = standardReactiveSocket_;
    return yarpl::Flowable<Payload>::create(
        [&eb, request = std::move(request), srs = std::move(srs) ](
            auto uptr_subscriber) mutable {
          auto os =
              std::make_shared<OldToNewSubscriber>(std::move(uptr_subscriber));
          eb.runInEventBaseThread([
            request = std::move(request),
            os = std::move(os),
            srs = std::move(srs)
          ]() mutable { srs->requestStream(std::move(request), std::move(os));
          });
        });
```

How do you recommend doing this with the new `Flowable`?

In short, I don't think the `create(Emitter)` can do this one as that assumes synchronous execution, whereas this is an async use case.

experimental/rsocket/OldNewBridge.h
benjchristensen: I was thinking we were going to change that to whatever we come up with here. That's why there is a new yarpl version of ReactiveStreams.

examples/stream-hello-world/StreamHelloWorld_Client.cpp
benjchristensen: Specifically, I have to do this:

```cpp
auto s = yarpl::Reference<yarpl::Subscriber<Payload>>(new ExampleSubscriber(5, 6));
```

but I want this:

```cpp
auto s = yarpl::Reference<ExampleSubscriber>(new ExampleSubscriber(5, 6));
```

so that I can call methods of `ExampleSubscriber`.

But that can't be passed into `subscribe`:

> error: no viable conversion from 'Reference<rsocket_example::ExampleSubscriber>' to 'Reference<Subscriber<reactivesocket::Payload>>'

vjn:
```
  LOG(INFO) << "------------------ Hello Bob!";
  auto s1 = yarpl::Reference<ExampleSubscriber>(new ExampleSubscriber(5, 6));
  rs->requestStream(Payload("Bob"))
      ->subscribe(yarpl::Reference<Subscriber>(s1.get()));
  s1->awaitTerminalEvent();

  LOG(INFO) << "------------------ Hello Jane!";
  auto s2 = yarpl::Reference<ExampleSubscriber>(new ExampleSubscriber(5, 6));
  rs->requestStream(Payload("Jane"))
      ->subscribe(yarpl::Reference<Subscriber>(s2.get()));
  s2->awaitTerminalEvent();
```

benjchristensen: Thanks, I've committed the fix that uses this approach.

PR #353, opened 2017-04-25
Upgrade clang version to 4.0

lehecka:


lehecka:
this has landed in fbcode

PR #354, opened 2017-04-25
Attempt at Flowables::fromPublisher, and an example

vjn:
The example has an "onSubscribe" function, and a fromPublisher operator. I think this is what @benjchristensen had in mind.


lehecka:
closing as it landed with https://github.com/ReactiveSocket/reactivesocket-cpp/pull/356

PR #355, opened 2017-04-25
Flowable delegating everything to onSubscribeFunc

vjn:
Simpler version, with ::yarpl:: types, and just stashing away an onSubscribeFunc for use during later subscribe() calls.


lehecka:
closing as it landed with https://github.com/ReactiveSocket/reactivesocket-cpp/pull/356

PR #357, opened 2017-04-26
fixing travis build

lehecka:


lehecka:
Vjn will handle this..

PR #358, opened 2017-04-26
disabling flaky tests

lehecka:


lehecka:
modifying comments only

PR #362, opened 2017-04-28
Fix missing gflags includes

alexmalyshev:
Needed to build on Arch Linux.


lehecka:
I think because of mobile we have to use folly/Portability/Flags.h or something like this...

alexmalyshev:
Ah yeah, I can switch to folly/portability/GFlags.h.

There's at least one test file where we include gflags.h directly though (without my diff), I can fix that too.

alexmalyshev:
Ugh I typed a wrong command and I'm not happy with what I did to the history.  Let me abandon this  and put up a new PR where I do things right.

PR #364, opened 2017-04-28
Fix missing override warnings in ServerConnectionAcceptorTest.cpp

alexmalyshev:
MOCK_METHOD*() was being used to mock virtual methods.  There's nothing wrong
with this, it's just gcc/clang will warn about missing an override qualifier.


test/ServerConnectionAcceptorTest.cpp
```diff
@@ -23,29 +23,45 @@ class MockConnectionHandler : public ConnectionHandler {
   void setupNewSocket(
       std::shared_ptr<FrameTransport> frameTransport,
       ConnectionSetupPayload setupPayload) override {
-    doSetupNewSocket(std::move(frameTransport), setupPayload);
+    setupNewSocket_(std::move(frameTransport), setupPayload);
+  }
+
+  bool resumeSocket(
```
lehecka: override keyword?

test/ServerConnectionAcceptorTest.cpp
```diff
@@ -23,29 +23,45 @@ class MockConnectionHandler : public ConnectionHandler {
...
+      std::shared_ptr<FrameTransport> transport,
+      ResumeParameters params) override {
+    return resumeSocket_(std::move(transport), params);
+  }
+
+  void connectionError(
```
lehecka: ditto

lexs:
This is unfortunate because gtest hasn't been updated in a long time and doesn't support the override qualifier (or move-only types etc). MOCK_METHOD is really for mocking virtual methods.

Instead of adding more boilerplate could we either disable the warning in `test/` or add a macro that does the boilerplate for you?

test/ServerConnectionAcceptorTest.cpp
alexmalyshev: It's on the `resumeSocket()` method already.

test/ServerConnectionAcceptorTest.cpp
```diff
@@ -23,29 +23,45 @@ class MockConnectionHandler : public ConnectionHandler {
   void setupNewSocket(
       std::shared_ptr<FrameTransport> frameTransport,
       ConnectionSetupPayload setupPayload) override {
-    doSetupNewSocket(std::move(frameTransport), setupPayload);
+    setupNewSocket_(std::move(frameTransport), setupPayload);
+  }
+
+  bool resumeSocket(
+      std::shared_ptr<FrameTransport> transport,
+      ResumeParameters params) override {
```
alexmalyshev: Right here

alexmalyshev:
> Instead of adding more boilerplate...

I'll look into disabling the warning for `test/`.  I don't feel like adding a macro that'll need to be #included by every test file that uses mocks...

lexs:
Thank you!

PR #365, opened 2017-04-28
Removed extra versions of Flowable, leaving just one

vjn:
Deleted examples and tests that are covered in existing tests; also
commented out a few, which will need to be redone.  Those are still
present in the code, as a reminder of what needs to be ported.


benjchristensen:
Changes look good reading/scrolling through them all. Except for the build break.

benjchristensen:
Not obvious to me why the tck test would fail. I don't see any changes that should affect that.

somasun:
Seems to be a dependency issue - make[2]: *** No rule to make target `gmock-prefix/src/gmock-build/./googlemock/libgmock.a', needed by `experimental/yarpl/yarpl-tests'.  Stop.

somasun:
Seems to be a dependency issue
make[2]: *** No rule to make target `gmock-prefix/src/gmock-build/./googlemock/libgmock.a', needed by `experimental/yarpl/yarpl-tests'.  Stop.

alexmalyshev:
Yeah, it seems like we need

  `add_dependencies(yarpl-tests gmock)`

or for anything that uses `gmock` really...  Just having `yarpl-tests` link in `$GMOCK_LIBS` doesn't express a dependency between the two.

vjn:
Thanks, will add that, and if Travis OK's, will merge it.

PR #366, opened 2017-05-01
Adds socketOn(Connected|Disconnected|Close) to RequestHandler.h

ragansa:
Bringing D4947387 over to github


lehecka:
Please add also the (modified) unit test(s).

src/RequestHandler.h
```diff
@@ -74,5 +74,10 @@ class RequestHandler {
       const std::shared_ptr<Subscriber<Payload>>& subscriber) noexcept = 0;
   virtual void onSubscriberResumed(
       const std::shared_ptr<Subscriber<Payload>>& subscriber) noexcept = 0;
+
+  // TODO (T17774014): Move to separate interface
```
lexs: Is there anything depending on this diff that makes us having to delay this? It seems like splitting these is a <1h job.

ragansa: Hey Alexander!  I talked with Ondrej about this.  He wanted to hold off building out the separate interface until we were ready to actually start using these calls.  For now, this PR is to demonstrate that the calls are being made when they should be.

test/ReactiveSocketResumabilityTest.cpp
```diff
@@ -34,10 +35,15 @@ TEST(ReactiveSocketResumabilityTest, Disconnect) {
...
 
+  auto requestHandler = std::make_unique<StrictMock<MockRequestHandler>>();
+  EXPECT_CALL(*requestHandler, socketOnConnected()).Times(1);
+  EXPECT_CALL(*requestHandler, socketOnDisconnected(_)).Times(1);
+  EXPECT_CALL(*requestHandler, socketOnClosed(_)).Times(1);
+
   auto socket = ReactiveSocket::fromClientConnection(
       defaultExecutor(),
       std::move(socketConnection),
-      std::make_unique<DefaultRequestHandler>(),
```
ragansa: I changed this from the NullRequestHandler (alias: DefaultRequestHandler) to the mock in order to add the EXPECT_CALL.  Is this okay?  The tests still passed, and it doesn't look to me that the NullRequestHandler was doing a lot that the Mock wouldn't.  Are my assumptions correct?

lehecka: yep

PR #367, opened 2017-05-01
Refactor RSocketConnectionHandler out of RSocketServer

mwinstanley:
Alternative use cases may want to use similar logic to RSocketServer in connection handling behavior but manage and maintain their own RSockets. The connection handling behavior is pulled into its own file and adds two virtual function that can be implemented for customization of management behavior.

This has no functionality change.


experimental/rsocket-src/RSocketServer.cpp
```diff
@@ -1,77 +1,44 @@
...
-      const std::shared_ptr<Subscriber<Payload>>&
-          subscriber) noexcept override {
-    auto flowable =
-        handler_->handleRequestStream(std::move(request), std::move(streamId));
-    // bridge from the existing eager RequestHandler and old Subscriber type
-    // to the lazy Flowable and new Subscriber type
-    flowable->subscribe(yarpl::Reference<yarpl::flowable::Subscriber<Payload>>(
-        new NewToOldSubscriber(std::move(subscriber))));
+  void manageSocket(
+      std::shared_ptr<ConnectionSetupRequest> request,
```
lehecka: request parameter is not used. Where do you see this being used? It leads to question why shared_ptr and not unique_ptr of ConnectionSetupRequest.

lehecka: got the answer:
" I added it because I need the context in order to associate the socket with the right Session
otherwise I have a socket but no way to know what Session it goes with
Yeah it's not used in the sample code"

makes sense to me.

PR #368, opened 2017-05-01
Fix missing gflags includes

alexmalyshev:
Needed to build on Arch Linux


alexmalyshev:
Second attempt, I did one too many rebase commands in the last one due to being used to hg...

PR #369, opened 2017-05-01
Get rid of -Weverything for clang

alexmalyshev:
Weverything tends to have a lot of low-signal, high-firing warnings.  Right now,
we're dealing with these by blacklisting them with -Wno-foo as we hit them, but
this is a losing battle as every new clang version will bring more of them.

Stick with -Wall and -Wextra for now.  Those should have the most valuable
warnings, and we can enable more warnings if we find specific ones we think are
valuable.


lehecka:
agreed. good move.

PR #370, opened 2017-05-02
Observable Refactor

benjchristensen:
Refactor `Observable` to match the design of `Flowable`.


alexmalyshev:
From my reading,

`observable/Observer` is exactly the same as `flowable/Subscriber` and

`observable/Subscription` is the same as `flowable/Subscription` minus `Subscription::request(n)`.

Is there a reason why most of the stuff in `observable/` is just copy-pasted from `flowable/`?  It seems to me like we can implement all the stuff in `observable/` using the classes in `flowable/`

(I'm not too familiar with these APIs, so forgive me if I'm talking about stuff that's already been discussed)

benjchristensen:
The only difference is lack of request(n) on Subscription. But that is a fundamental difference that changes how the whole thing behaves, how an Observable is created, how it runs, and the performance characteristics. So it ends up meaning everything gets reimplemented with the simpler pure-push model without request(n) flow control.

benjchristensen:
The key difference is here: https://github.com/benjchristensen/rsocket-cpp/blob/803c64f7a2ea05457c2d152ee2731a87ea8eca3b/experimental/yarpl/include/yarpl/observable/Subscription.h

benjchristensen:
I think this is ready to merge and then followup with other work on top of it.

experimental/yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -55,18 +55,15 @@ class ObservableOperator : public Observable<D> {
     virtual void onComplete() override {
       subscriber_->onComplete();
       upstream_.reset();
-      release();
```
lehecka: good job figuring out the issue

PR #372, opened 2017-05-02
Code cleanup: moved (some) headers from src/ to include/

vjn:
Also deleted src/Tuple.cpp, no longer being used.


lexs:
What's the reason for having `include/` at all? Can we just move them all to `src`? This is how the rest of the code looks and what I personally prefer (makes navigation easier etc), but interested in hearing others opinion on the matter.

lehecka:
I understand the separation of public API (include/ folder) vs. internal implementation (src/ folder).
I also don't like to search for corresponding cpp file for a given h file. Doesn't CLion have some support for this?

vjn:
Yes, it'd be a pity to lose the include/ vs. src/ distinction, though with templates, the world is moving to header-only libraries. I think most every IDE now has automatic header/src switch: specifically I know about Emacs/cmake-ide, QT Creator (which I use) and CLion.

PR #374, opened 2017-05-02
Add a simple test for connecting an RSocketClient to an RSocketServer

alexmalyshev:


experimental/rsocket-test/RSocketClientServerTest.cpp
```diff
@@ -1,32 +1,61 @@
...
 using namespace rsocket;
 using namespace rsocket::tests;
 
+namespace {
+
+std::random_device device;
+std::uniform_int_distribution<uint16_t> dis(9000, 10000);
+
+// Helps prevent against port collisions.
+uint16_t randPort() {
```
lexs: Instead of this you can use port 0 to have the kernel assign a port (and read it back).

alexmalyshev: Neat, first time I've heard of it.  I tried to change the start-shutdown test to do that (since I wasn't sending anything across the wire anyway), but it caused the test to crash.  I'm guessing we're depending on some `if (socket)` check somewhere...

I'll merge this as is, and look into the crashes separately.

PR #376, opened 2017-05-02
Add Payload::Clone and fix NewToOldSubscriber which was cloning unsafely

nickgg:
Added unit test for the new method - I have tested NewOldBridge manually.

### DIFF
--- a/experimental/rsocket/OldNewBridge.h
+++ b/experimental/rsocket/OldNewBridge.h
@@ -121,8 +121,7 @@ class NewToOldSubscriber : public yarpl::flowable::Subscriber<reactivesocket::Pa
   void onNext(const reactivesocket::Payload& payload) override {
     // Cloning IOBufs just shares their internal buffers, so this isn't the end
     // of the world.
-    auto clone = payload.clone();
-    inner_->onNext(std::move(clone));
+    inner_->onNext(payload.clone());
   }
 
   void onComplete() override {
--- a/experimental/rsocket/OldNewBridge.h
+++ b/experimental/rsocket/OldNewBridge.h
@@ -121,8 +121,7 @@ class NewToOldSubscriber : public yarpl::flowable::Subscriber<reactivesocket::Pa
   void onNext(const reactivesocket::Payload& payload) override {
     // Cloning IOBufs just shares their internal buffers, so this isn't the end
     // of the world.
-    reactivesocket::Payload clone(
-        payload.data->clone(), payload.metadata->clone());
+    reactivesocket::Payload clone = payload.clone();
     inner_->onNext(std::move(clone));
   }
 
--- a/experimental/rsocket/OldNewBridge.h
+++ b/experimental/rsocket/OldNewBridge.h
@@ -121,7 +121,7 @@ class NewToOldSubscriber : public yarpl::flowable::Subscriber<reactivesocket::Pa
   void onNext(const reactivesocket::Payload& payload) override {
     // Cloning IOBufs just shares their internal buffers, so this isn't the end
     // of the world.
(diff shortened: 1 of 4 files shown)


src/Payload.cpp
```diff
@@ -55,6 +55,18 @@ void Payload::clear() {
   metadata.reset();
 }
 
+Payload Payload::clone() const {
+	Payload out;
```
alexmalyshev: We use 2-space indents, I think these are tabs?

src/Payload.h
```diff
@@ -34,6 +34,9 @@ struct Payload {
   std::string cloneDataToString() const;
   void clear();
 
+  // Clone this object onto the stack
```
alexmalyshev: Not sure the `onto the stack` part is helpful.  A Payload is a tuple of 2 heap pointers, might be confusing.  No good ideas on what to say instead though...

test/PayloadTest.cpp
```diff
@@ -34,3 +34,47 @@ TEST(PayloadTest, GiantMetadata) {
       FrameSerializerV0_1::deserializeMetadataFrom(cur, FrameFlags::METADATA),
       std::runtime_error);
 }
+
+TEST(PayloadTest, Clone) {
```
alexmalyshev: (y)

experimental/rsocket/OldNewBridge.h
```diff
@@ -121,8 +121,7 @@ class NewToOldSubscriber : public yarpl::flowable::Subscriber<reactivesocket::Pa
   void onNext(const reactivesocket::Payload& payload) override {
     // Cloning IOBufs just shares their internal buffers, so this isn't the end
     // of the world.
-    reactivesocket::Payload clone(
-        payload.data->clone(), payload.metadata->clone());
+    reactivesocket::Payload clone = payload.clone();
```
alexmalyshev: `auto clone = payload.clone();`

src/Payload.cpp
```diff
@@ -55,6 +55,18 @@ void Payload::clear() {
...
+	if (data) {
+		out.data = data->clone();
```
lexs: tabs? 👎

src/Payload.cpp
lexs:
```cpp
return Payload(
  data ? data->clone() : nullptr,
  metadata ? metadata->clone() : nullptr)
```
is a more concise and clearer version

src/Payload.cpp
nickgg: Ah, I copied this out of phabricator - not sure why it came out like this.

src/Payload.cpp
nickgg: I don't agree with clearer, you've packed two ternary operators into a variable intialisation.

lexs: To each his own I suppose. I felt it was clearer as we're no longer mutating the Payload. I don't feel strongly about it however and it'll compile to exactly the same code so ¯\_(ツ)_/¯

experimental/rsocket/OldNewBridge.h
```diff
@@ -121,8 +121,7 @@ class NewToOldSubscriber : public yarpl::flowable::Subscriber<reactivesocket::Pa
   void onNext(const reactivesocket::Payload& payload) override {
     // Cloning IOBufs just shares their internal buffers, so this isn't the end
     // of the world.
-    reactivesocket::Payload clone(
-        payload.data->clone(), payload.metadata->clone());
+    auto clone = payload.clone();
     inner_->onNext(std::move(clone));
```
lexs: nit: `inner_->onNext(payload.clone());`

PR #377, opened 2017-05-02
Open frame conn automaton

ragansa:
Encapsulates frame management/serialization/de-serialization inside ConnectionAutomaton


src/ConnectionAutomaton.h
```diff
@@ -186,7 +186,7 @@ class ConnectionAutomaton final
     return streamsFactory_;
   }
 
-  FrameSerializer& frameSerializer() const override;
+  FrameSerializer& frameSerializer() const ;//override;
```
lexs: what is this?

src/automata/ChannelRequester.cpp
```diff
@@ -130,15 +130,15 @@ void ChannelRequester::endStream(StreamCompletionSignal signal) {
   ConsumerBase::endStream(signal);
 }
 
-void ChannelRequester::onNextFrame(Frame_PAYLOAD&& frame) {
+void ChannelRequester::handlePayload(Payload&& payload, bool complete, bool flagsNext) {
```
lexs: bool arguments is kinda horrible, why exactly are we no longer passing the frame? if we really don't want to pass the frame could we create some sort of Flags type that allows easier querying? `flags.has(FrameFlags::COMPLETE)`?

src/automata/ChannelRequester.cpp
```diff
@@ -147,14 +147,14 @@ void ChannelRequester::onNextFrame(Frame_PAYLOAD&& frame) {
...
-  processPayload(std::move(frame.payload_), frame.header_.flagsNext());
+  processPayload(std::move(payload), flagsNext);
 
   if (end) {
     closeStream(StreamCompletionSignal::COMPLETE);
   }
 }
 
-void ChannelRequester::onNextFrame(Frame_ERROR&& frame) {
+void ChannelRequester::handleError(std::string errorPayload) {
```
lexs: this should pass an `folly::exception_wrapper`

src/automata/ChannelResponder.cpp
```diff
@@ -91,18 +91,20 @@ void ChannelResponder::processInitialFrame(Frame_REQUEST_CHANNEL&& frame) {
       true);
 }
 
+/*
```
lexs: what's going on here? why not remove this if we no longer need it?

src/automata/RequestResponseRequester.cpp
```diff
@@ -71,7 +71,7 @@ void RequestResponseRequester::endStream(StreamCompletionSignal signal) {
   }
 }
 
-void RequestResponseRequester::onNextFrame(Frame_ERROR&& frame) {
+void RequestResponseRequester::handleError(std::string errorPayload) {
```
lexs: same here

src/ConnectionAutomaton.cpp
```diff
@@ -568,8 +561,75 @@ void ConnectionAutomaton::onConnectionFrame(
...
+                               frame_payload.header_.flagsNext());
+      break;
+    }
+    case FrameType::ERROR: {
+      Frame_ERROR frame_error;
+      if (!deserializeFrameOrError(frame_error,
+                                   std::move(serializedFrame))) {
+        return;
+      }
+      automaton->handleError(frame_error.payload_.moveDataToString());
```
ragansa: @lexs 
I've updated the later methods to use folly::exception_wrapper, but I'm not clear on how to construct one that will capture the error frame's payload.

lexs: You'll need to give it an actual exception, `std::runtime_error("the_string")` is probably fine for now.

ragansa: That did it!  Thanks!

lehecka:
The summary for this PR is rather short and not descriptive.
Scott, can you update why this change and what is the intent for the code changes?

src/ConnectionAutomaton.h
lehecka: I believe we can remove the method now as the intent for this PR is to not use Frames and FrameSerializer outside of ConnectionAutomaton.

ragansa:
I've made a huge mess of this, I'm going to try to get a clean PR together

PR #379, opened 2017-05-02
Refactor TcpConnectionFactory

alexmalyshev:
* s|SocketConnectorAndCallback|ConnectCallback|g

* Construct the ConnectCallback object entirely on the worker thread.  Fixes
  TSAN warnings about the object being accessed from both the main thread and
  the worker thread.

* Use a unique_ptr deleter in both connectSuccess() and connectErr().

* Switch all the LOG(INFO)s to VLOGs.  We can opt into the terminal spam.

Tested the stream-hello-world client and server.  ASAN is happy.  TSAN is
generating a lot of noise in Flowable code, but the warnings from
TcpConnectionFactory appear to be gone.

### DIFF
--- a/examples/conditional-request-handling/ConditionalRequestHandling_Client.cpp
+++ b/examples/conditional-request-handling/ConditionalRequestHandling_Client.cpp
@@ -25,22 +25,27 @@ int main(int argc, char* argv[]) {
   FLAGS_minloglevel = 0;
   folly::init(&argc, &argv);
 
+  folly::SocketAddress address;
+  address.setFromHostPort(FLAGS_host, FLAGS_port);
+
   auto rsf = RSocket::createClient(
-      TcpConnectionFactory::create(FLAGS_host, FLAGS_port));
+      std::make_unique<TcpConnectionFactory>(std::move(address)));
   auto rs = rsf->connect().get();
 
   LOG(INFO) << "------------------ Hello Bob!";
   auto s1 = yarpl::Reference<ExampleSubscriber>(new ExampleSubscriber(5, 6));
   rs->requestStream(Payload("Bob"))
       ->take(5)
-      ->subscribe(yarpl::Reference<yarpl::flowable::Subscriber<Payload>>(s1.get()));
+      ->subscribe(
+          yarpl::Reference<yarpl::flowable::Subscriber<Payload>>(s1.get()));
   s1->awaitTerminalEvent();
 
   LOG(INFO) << "------------------ Hello Jane!";
   auto s2 = yarpl::Reference<ExampleSubscriber>(new ExampleSubscriber(5, 6));
   rs->requestStream(Payload("Jane"))
       ->take(3)
-      ->subscribe(yarpl::Reference<yarpl::flowable::Subscriber<Payload>>(s2.get()));
+      ->subscribe(
+          yarpl::Reference<yarpl::flowable::Subscriber<Payload>>(s2.get()));
   s2->awaitTerminalEvent();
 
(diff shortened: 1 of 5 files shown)


examples/conditional-request-handling/ConditionalRequestHandling_Client.cpp
```diff
@@ -25,22 +25,27 @@ int main(int argc, char* argv[]) {
...
+      std::make_unique<TcpConnectionFactory>(std::move(address)));
   auto rs = rsf->connect().get();
 
   LOG(INFO) << "------------------ Hello Bob!";
   auto s1 = yarpl::Reference<ExampleSubscriber>(new ExampleSubscriber(5, 6));
   rs->requestStream(Payload("Bob"))
       ->take(5)
-      ->subscribe(yarpl::Reference<yarpl::flowable::Subscriber<Payload>>(s1.get()));
+      ->subscribe(
+          yarpl::Reference<yarpl::flowable::Subscriber<Payload>>(s1.get()));
```
lexs: isn't s1 copyable? Why do you need to explicitly create a new Reference?

examples/conditional-request-handling/ConditionalRequestHandling_Client.cpp
```diff
@@ -25,22 +25,27 @@ int main(int argc, char* argv[]) {
...
   s1->awaitTerminalEvent();
 
   LOG(INFO) << "------------------ Hello Jane!";
   auto s2 = yarpl::Reference<ExampleSubscriber>(new ExampleSubscriber(5, 6));
   rs->requestStream(Payload("Jane"))
       ->take(3)
-      ->subscribe(yarpl::Reference<yarpl::flowable::Subscriber<Payload>>(s2.get()));
+      ->subscribe(
+          yarpl::Reference<yarpl::flowable::Subscriber<Payload>>(s2.get()));
```
lexs: same here

experimental/rsocket-src/transports/TcpConnectionFactory.cpp
```diff
@@ -1,99 +1,84 @@
...
-  folly::AsyncSocket::UniquePtr socket_;
-  OnConnect onConnect_;
-  EventBase* eventBase_;
+  ~ConnectCallback() {
+    VLOG(2) << "Destroying ConnectCallback";
+  }
 
   void connectSuccess() noexcept {
-    LOG(INFO) << "ConnectionFactory => socketCallback => Success";
+    std::unique_ptr<ConnectCallback> deleter(this);
```
lexs: I realize this was already here, but this is all kinds of horrible

examples/conditional-request-handling/ConditionalRequestHandling_Client.cpp
alexmalyshev: (This was a clang-format change)

Unfortunately, Reference<Derived> can't be upcast to Reference<Base>, you'll just get a compiler error.  That's why most Subscriber code gets the raw Subscriber pointer and makes a new Reference out of it.  I have a diff in the works to fix that, just need to fix some outstanding test issues first.

experimental/rsocket-src/transports/TcpConnectionFactory.cpp
alexmalyshev: I'm not sure it is, doing `delete this;` in folly socket callbacks is a common thing.  Most of the time I see people doing it with `SCOPE_EXIT { delete this; };`, but the `unique_ptr` way also works.

PR #385, opened 2017-05-03
Fix object leak when chaining operators that do not complete

nickgg:
FlowableOperator relied on a complete to release the upstream subscription, meaning that it would leak if the source flowable was unfinished when it was cancelled. The new test covers a few cases that would cause the leak.


nickgg:
We use valgrind for the heap checking?

lexs:
> We use valgrind for the heap checking?

Would ASAN find it? We do build asan on travis so should be possible to locally too.

alexmalyshev:
ASAN wouldn't find it, but LSAN would/should.  I should note that LSAN is only supported by Linux x64 which is what I think we're running on Travis, but local macOS laptop development wouldn't catch it.

nickgg:
I've tried a few times to reproduce this and haven't been able to, but thinking about it I wonder if it might be related to the initializer list safety issue in #389?

alexmalyshev:
I don't think it's the initializer_list, as the Travis errors are about accessing a `std::string` that's already been destructed.  The initializer list contains `const char*` values with process lifetime.

I think an operator is calling `Subscriber::onNext(const T&)` with a temporary, and then it goes out of scope and we die someplace later.  Hence why I want to make them take `T` or `T&&` instead (https://github.com/rsocket/rsocket-cpp/pull/387)

nickgg:
I don't understand initializer lists as deeply as maybe I should, but that doesn't sound correct. The just<std::string> constructor requires an initializer_list<std::string> so the type conversion needs to be done before the constructor. No reason that it should ever exist as a const char*: that feels like the kind of optimization a compiler would do.

Not to say i'm convinced it's the issue here though.

alexmalyshev:
> juststd::string constructor requires an initializer_liststd::string so the type conversion needs to be done before the constructor

Ah you're right, I read it wrong.  Then it does seem like #389, as the `Flowables::just()` is grabbing an `initializer_list<std::string>` full of `std::string` temporaries.

nickgg:
Uh ok, I'm going to do a fresh PR.

PR #387, opened 2017-05-04
Switch {flowable,observable}::Subscriber::onNext() to take a T

alexmalyshev:
Instead of a `const T&`.  `const T&` is too limited in that it will not work
with move-only types.  We tried to have separate overloads for `const T&` and
`T&&`, but ran into issues with some compilers incorrectly treating `T&&` as a
universal reference.

### DIFF
--- a/examples/util/ExampleSubscriber.cpp
+++ b/examples/util/ExampleSubscriber.cpp
@@ -29,9 +29,9 @@ void ExampleSubscriber::onSubscribe(
   subscription_->request(initialRequest_);
 }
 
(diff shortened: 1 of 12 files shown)


benjchristensen:
Is the plan that once we have compiler support we'll move back to `const T&` and add the `T&&` overload?

alexmalyshev:
I'm not sure if that'll be necessary.  Using `T` gives us access to both copying and moving of data.  I'm not sure if it'll be valuable to split the logic for lvalues and rvalues.

PR #388, opened 2017-05-04
Get rid of unnecessary Subscriber upcasts

alexmalyshev:
Now that the new overloads for Reference are available


benjchristensen:
Fixed merge conflicts, waiting on build to see if I fixed them correctly.

PR #389, opened 2017-05-04
Rename Flowables::just(initialize_list<>) to ::justN(), make it safe

alexmalyshev:
Copying an initialize_list is unsafe, as it doesn't actually own the memory it
points to.  To make it safe we'll just save all the values into a vector and
keep that around.

Rename the overload to justN() to make the usage clearer.

### DIFF
--- a/experimental/yarpl/include/yarpl/Flowables.h
+++ b/experimental/yarpl/include/yarpl/Flowables.h
@@ -47,8 +47,7 @@ class Flowables {
 
   template <typename T>
   static Reference<Flowable<T>> justN(std::initializer_list<T> list) {
-    std::vector<T> vec;
-    std::move(list.begin(), list.end(), std::back_inserter(vec));
+    std::vector<T> vec(list);
 
     auto lambda = [ v = std::move(vec), i = size_t{0} ](
         Subscriber<T> & subscriber, int64_t requested) mutable {
--- a/experimental/yarpl/include/yarpl/Flowables.h
+++ b/experimental/yarpl/include/yarpl/Flowables.h
@@ -1,6 +1,8 @@
 #pragma once
 
+#include <algorithm>
 #include <limits>
+#include <vector>
 
 #include "Flowable.h"
 
@@ -44,23 +46,28 @@ class Flowables {
   }
 
   template <typename T>
-  static Reference<Flowable<T>> just(std::initializer_list<T> list) {
-    auto lambda = [ list, it = list.begin() ](
+  static Reference<Flowable<T>> justN(std::initializer_list<T> list) {
+    std::vector<T> vec;
+    std::move(list.begin(), list.end(), std::back_inserter(vec));
+
+    size_t i = 0;
+
+    auto lambda = [ v = std::move(vec), i ](
         Subscriber<T> & subscriber, int64_t requested) mutable {
       int64_t emitted = 0;
       bool done = false;
 
-      while (it != list.end() && emitted < requested) {
-        subscriber.onNext(*it++);
+      while (i < v.size() && emitted < requested) {
+        subscriber.onNext(v[i++]);
         ++emitted;
       }
 
-      if (it == list.end()) {
+      if (i == v.size()) {
         subscriber.onComplete();
         done = true;
       }
 
-      return std::make_tuple(static_cast<int64_t>(emitted), done);
+      return std::make_tuple(emitted, done);
     };
 
     return Flowable<T>::create(std::move(lambda));
--- a/experimental/yarpl/include/yarpl/Flowables.h
+++ b/experimental/yarpl/include/yarpl/Flowables.h
@@ -50,9 +50,7 @@ class Flowables {
     std::vector<T> vec;
     std::move(list.begin(), list.end(), std::back_inserter(vec));
 
-    size_t i = 0;
-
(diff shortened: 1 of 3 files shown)


experimental/yarpl/include/yarpl/Flowables.h
```diff
@@ -44,23 +46,28 @@ class Flowables {
   }
 
   template <typename T>
-  static Reference<Flowable<T>> just(std::initializer_list<T> list) {
-    auto lambda = [ list, it = list.begin() ](
+  static Reference<Flowable<T>> justN(std::initializer_list<T> list) {
+    std::vector<T> vec;
+    std::move(list.begin(), list.end(), std::back_inserter(vec));
+
+    size_t i = 0;
```
nickgg: This is only used inside the lambda, could move it in there rather than capture.

experimental/yarpl/include/yarpl/Flowables.h
```diff
@@ -44,23 +46,28 @@ class Flowables {
...
+
+    auto lambda = [ v = std::move(vec), i ](
         Subscriber<T> & subscriber, int64_t requested) mutable {
       int64_t emitted = 0;
       bool done = false;
 
-      while (it != list.end() && emitted < requested) {
-        subscriber.onNext(*it++);
+      while (i < v.size() && emitted < requested) {
```
nickgg: Why the change from an iterator to access by index?

experimental/yarpl/include/yarpl/Flowables.h
```diff
@@ -44,23 +46,28 @@ class Flowables {
   }
 
   template <typename T>
-  static Reference<Flowable<T>> just(std::initializer_list<T> list) {
-    auto lambda = [ list, it = list.begin() ](
+  static Reference<Flowable<T>> justN(std::initializer_list<T> list) {
```
nickgg: What is the motivation here, i.e. what usage is clearer? I guess I can't envisage a situation where it matters which implementation of the just flowable you are using.

experimental/yarpl/include/yarpl/Flowables.h
alexmalyshev: I initially had it as `[..., i = 0]` in the lambda's capture list, but then got a signed vs unsigned warning when comparing `i` with `v.size()`, so I declared it up here as a `size_t`.

(I don't think I can do [..., size_t i = 0]` in a capture list, but I haven't tried....

experimental/yarpl/include/yarpl/Flowables.h
alexmalyshev: `just(T)` should send a single `T`.  It seemed weird to me to have
  `just(T) (T=initializer_list)` not send a single `initializer_list` (despite that being a bad idea and all).  I dunno, I like that when I read `just()` I can easily tell that it's only one value to send.

experimental/yarpl/include/yarpl/Flowables.h
alexmalyshev: I wasn't sure if `[v = std::move(vec), it = v.begin()]` would be safe, i.e. I'm not sure if the capture list expressions are guaranteed to be executed LTR by the C++ standard.

experimental/yarpl/include/yarpl/Flowables.h
nickgg: Does it have to be in the capture list? Why not just define it inside the lambda func - like emitted & done?

experimental/yarpl/include/yarpl/Flowables.h
nickgg: I guess this is the same thing as the above, if its not used outside the lambda then you don't need to define it outside the lambda and capture it - you can just have it in the lambda's local scope.

experimental/yarpl/include/yarpl/Flowables.h
lexs: @nickgg Because it's mutated and needs to be persisted between calls to the lambda.

nickgg: Ah okay, it's implicit state. That makes sense, but it's non-obvious from a reading of the code. 

If the simple flowables are intended partially to teach people how to write their own flowables then perhaps a comment?

experimental/yarpl/include/yarpl/Flowables.h
```diff
@@ -44,23 +46,28 @@ class Flowables {
...
+    std::vector<T> vec;
```
alexmalyshev: I checked the source for vector, and it seems like that ctor copies each element from the list, whereas I want to move them all in.

lexs: This suggests that moving is essentially a no-op (due to const) anyway: http://stackoverflow.com/a/8193157/253583

alexmalyshev: Thanks for the link.  Yeah, that's unfortunate, but it makes sense though.  initializer_list<T> doesn't own the `T` values, so you shouldn't be able to move from them (and most users of it will make copies).

I'll switch it to using the copying ctor in vector.  I also considered having justN() be a variadic template, but it gets weird because the values in a parameter pack aren't required to be all the same type.

PR #391, opened 2017-05-04
Add Flowables::repeat(T)

alexmalyshev:
Should be useful in testing out the movability/copyability of our `T` values.
So far all our operators and `Flowables` helpers only require `T` to be movable,
repeat() will be the first use case that requires `T` to be copyable.

### DIFF
--- a/experimental/yarpl/include/yarpl/Flowables.h
+++ b/experimental/yarpl/include/yarpl/Flowables.h
@@ -104,6 +104,18 @@ class Flowables {
     return Flowable<T>::create(std::move(lambda));
   }
 
+  template <typename T>
+  static Reference<Flowable<T>> repeat(T value) {
+    auto lambda = [v = std::move(value)](
+        Subscriber<T> & subscriber, int64_t requested) mutable {
+      for (int64_t i = 0; i < requested; ++i) {
+        subscriber.onNext(v);
+      }
+      return std::make_tuple(requested, false);
+    };
+    return Flowable<T>::create(std::move(lambda));
+  }
+
  private:
   Flowables() = delete;
 };
--- a/experimental/yarpl/test/FlowableTest.cpp
+++ b/experimental/yarpl/test/FlowableTest.cpp
@@ -164,5 +164,11 @@ TEST(FlowableTest, FlowableEmpty) {
   EXPECT_EQ(collector->error(), false);
 }
 
+TEST(FlowableTest, FlowableRepeat) {
+  auto flowable = Flowables::repeat<int>(5)->take(10);
+  EXPECT_EQ(run(std::move(flowable)), std::vector<int>(10, 5));
+  EXPECT_EQ(std::size_t{0}, Refcounted::objects());
+}
+
 } // flowable
 } // yarpl


nickgg:
Oh rats I was just working on this: https://github.com/nickgg/rsocket-cpp/tree/cycle-flowable - can we compare the two?

Differences: 
* I called it cycle rather than repeat (thats what haskell & python call it), but no strong feelings.
* I added the initialiser list version as well.
* Tests for my version are a bit more comprehensive (i found the bug from #385, for instance)

alexmalyshev:
Hah, great minds think alike.

* I called it `repeat()` because I believe that's what the other rx/reactive libraries call it.
* The initializer list overload is interesting.  I still find it a bit odd to have `repeat(ONE_VALUE)` and `repeat(INITIALIZER_LIST)` have the same name, as they don't really do the same thing.  I'd actually say that the `initializer_list` overload would be better named as `repeatCycle()` (assuming the normal one was called `repeat()`).
* Yeah I wasn't looking to test it thoroughly, mostly to add an example that compiled.

nickgg:
* definitely looks like repeat is preferred in the reactive world: http://reactivex.io/documentation/operators/repeat.html
* I guess it depends on the way you look at it, I was thinking of them both as overloads of repeat(SOURCE). If you take a look at rxjava it does this by making the input be a publisher an then resubscribing to it on complete: https://github.com/reactor/reactive-streams-commons/blob/master/src/main/java/rsc/publisher/PublisherRepeat.java. 

Changing name for each input type can blow out if you have a bunch of input sources, e.g. now we have just, justN and fromPublisher (which is basically the just(PUBLISHER) overload) - for repeat we need to add repeat, repeatN, repeatFromPublisher? For every new kind of flowable we add 3 more api names, and for every new type of source we add X (for each applicable flowable kind).

I don't think the user cares about the distinction here, and we can save them from some effort. I envisage wanting to write stuff like just(SOMETHING)->map(FUNC)->take(X) and the what those things represent is important to me, not how they're implmented.

I don't feel that strongly about this though, does anyone else have an opinion?
* If not now, then when? :)

alexmalyshev:
> I was thinking of them both as overloads of repeat(SOURCE)...

Yeah this isn't how I read it.  To me, `repeat(T)` means a Flowable that always sends the given `T` value.  `repeat(initializer_list)` on the other hand is a Flowable that sends a value from a list, cycling through the list infinitely.  In my mind these behaviors are different enough that they don't deserve the same name.  To be honest I would name the latter as `cycle(initializer_list)`.

> e.g. now we have just, justN and fromPublisher (which is basically the just(PUBLISHER) overload)

I don't agree with the last part of that.  When I see `just(T)` I think of a Flowable that sends a single T value and then completes.  `fromPublisher` can do whatever it wants with the subscriber, there's no more notion of that "single value and then complete".

> Changing name for each input type can blow out if you have a bunch of input sources...

But this only applies to "creation" operators as listed on http://reactivex.io/documentation/operators.html#creating, of which there aren't that many.  Everything else (e.g. map, filter, take, first, etc.) just takes a single input at a time.

> If not now, then when? :)

:)  Yeah I should have added more.

benjchristensen:
Repeat is an operator I can apply to any Flowable. For example: just(T).repeat() or fromIterable(iter).repeat(). 

Regardless of the source I can cause it to repeat indefinitely until cancelled.

alexmalyshev:
> Regardless of the source I can cause it to repeat indefinitely until cancelled.

Does that mean that the `repeat()` operator just re-subscribes to whatever flowable it was applied to?  Or does it buffer all `onNext()` signals it gets until `onComplete()`, after which it cycles through its buffer?

From your description it seems like you mean the former, but I really think it should be the latter.  If I have `F = Flowables::just(V)`, then I want `F` to move `V` to its subscriber via `Subscriber::onNext(std::move(V))`.  If I then compose a re-subscribing operator on that, I won't get `V` anymore, I'll get an empty object as `F` no longer owns `V`.

benjchristensen:
It just resubscribes. Buffering by default is a bad thing with very large streams. The standard approach to that in these libraries is compose replay().repeat(). 

The replay() operator with no arguments caches everything it receives and replays it each time it is subscribed to.

alexmalyshev:
Ah okay, so the resubscribing logic and the buffer logic are two different operators, that makes more sense.

With the way I see things, that means `Flowables::just<T>(T(5)).repeat()` will stream `T(5), T(), T(), ...`, because `just` will relinquish ownership of the object after the first `onNext`, and then afterwards can only send a default object.

`Flowables::just<T>(T(5)).replay().repeat()` will stream `T(5), T(5), T(5), ...`, and the replay operator will statically assert that `T` is a copyable object.

benjchristensen:
That's a problem if it relinquishes the object because as long as someone has a reference to the Flowable anyone can subscribe to it and it needs to work correctly.

benjchristensen:
In other words, just(T) needs to copy T, or a reference to T and be able to emit it multiple times. Otherwise that Flowable doesn't really work correctly.

experimental/yarpl/include/yarpl/Flowables.h
```diff
@@ -104,6 +104,18 @@ class Flowables {
     return Flowable<T>::create(std::move(lambda));
   }
 
+  template <typename T>
+  static Reference<Flowable<T>> repeat(T value) {
+    auto lambda = [v = std::move(value)](
+        Subscriber<T> & subscriber, int64_t requested) mutable {
```
phoad: Can we give a negative value into this function? Should we have any check for negative?

alexmalyshev: The rsocket spec forbids request(n) where n <= 0.  That said, we still use `int64_t` because it's a signed int in the spec for compat with Java.

I don't think we should check for `requested <= 0` here because that would suggest a bug in `Flowable::create()`.

alexmalyshev:
I'm now onboard with @benjchristensen's explanation of how `just()` should keep a copy of the value and that `repeat()` should just re-subscribe.  That said, not enough time to work on this right now.

PR #392, opened 2017-05-04
Add `Reference<T> make_ref<T, Args...>(Args&&...)`

alexmalyshev:
Akin to `std::make_shared`.  Makes code more concise.

### DIFF
--- a/experimental/yarpl/examples/FlowableExamples.cpp
+++ b/experimental/yarpl/examples/FlowableExamples.cpp
@@ -47,7 +47,7 @@ void fromPublisherExample() {
       }
     };
 
-    Reference<::yarpl::flowable::Subscription> subscription(new Subscription);
+    auto subscription = make_ref<Subscription>();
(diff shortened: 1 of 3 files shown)


lexs:
I'm asking here because I assume there is a reason we have `Reference` but I've never seen it explained anywhere. What does it bring over `shared_ptr`? `make_shared` is also mostly invasive (inlines the control block with the object).

benjchristensen:
@vjn can explain this far better than I. But to start the discussion:

- less memory overhead 
- working towards support of something we can do both stack and heap allocation with, and paging in/out of disk so we can have 100s of millions of these things 
- it can release itself from inside a subscriber/subscription safely in response to cancel/complete/error

vjn:
@lexs Reference<> is basically boost::intrusive_ptr<>, and has the same advantages. The only difference is that argument-dependent lookup is not needed, and we don't handle classes not inheriting from a common base. (This is mentioned in Refcounted.h.)

PR #393, opened 2017-05-05
[rsocket-cpp] Basic cold resumption.

somasun:
This is a first stab at cold resumption.

API Changes
- It tries to reuse warm-resumption mechanics as much as possible.  The state
is preserved in ResumeCache.  ResumeCache is an optional parameter while
constructing a client ReactiveSocket.
- Streams are identified by a unique StreamName (called it StreamName to
distinguish it from StreamId).  Apps are responsible for maintaining a unique
streamName for all the streams they create (could be a simple hash of the
initialPayload).
- Introduced ReactiveSocket::requestStream() which takes an addition streamName
parameter.   A streamName must be passed by clients to ensure the streams are
resumable.  This returns error if a stream already exists with the same name.
- Introduced RequestHandler::handleResumeStream(streamName).  This method gets
called when streams are resumed after cold start.  Apps should implement this
method and return a Subscriber for this streamName.
- All other API changes should be internal to RSocket.

Tests
- I have written two integration tests - BasicColdResumtion and
2StreamColdResumption

Open to discussion
- If a stream is created without a name and a cold-resumption is attempted,
then that stream is resumed with its streamId being used as the streamName.
An alternative is to enforce all streams are created with a name (might be
too harsh).
- everything else.

TODO / Future Work

1. More sanity checks.  Eg: check RSocket::mode is same after cold-resumption.
2. Right now we FATAL if handleResumeStream() is called but not implemented.  Find a better alternative.
3. ResumeCache::streamMap_ is not accurate.  It has entries only for streams which currently have frames buffered in the cache. Also streamNameMap_ is never flushed.  We need to remove entries as streams get closed.
4. Allowance in StreamRequester needs to be persisted.
5. Change ResumeCache to an interface.  Have two sample implementations. 
 InMemResumeCache and DiskResumeCache.  InMemResumeCache will be the default and use the existing implementation of ResumeCache.  DiskResumeCache will write to disk.

### DIFF
--- a/src/temporary_home/ReactiveSocket.cpp
+++ b/src/temporary_home/ReactiveSocket.cpp
@@ -26,6 +26,7 @@ ReactiveSocket::~ReactiveSocket() {
 ReactiveSocket::ReactiveSocket(
     ReactiveSocketMode mode,
     std::shared_ptr<RequestHandler> handler,
+    std::shared_ptr<ResumeCache> resumeCache,
     std::shared_ptr<Stats> stats,
     std::unique_ptr<KeepaliveTimer> keepaliveTimer,
     folly::Executor& executor)
@@ -35,7 +36,8 @@ ReactiveSocket::ReactiveSocket(
           std::move(handler),
           std::move(stats),
           std::move(keepaliveTimer),
-          mode)),
+          mode,
+          std::move(resumeCache))),
       executor_(executor) {
   debugCheckCorrectExecutor();
   connection_->stats().socketCreated();
@@ -52,6 +54,7 @@ ReactiveSocket::fromClientConnection(
   auto socket = disconnectedClient(
       executor,
       std::move(handler),
+      std::make_shared<ResumeCache>(stats),
       std::move(stats),
       std::move(keepaliveTimer),
       setupPayload.protocolVersion);
@@ -65,12 +68,14 @@ std::unique_ptr<ReactiveSocket>
 ReactiveSocket::disconnectedClient(
     folly::Executor& executor,
     std::unique_ptr<RequestHandler> handler,
+    std::shared_ptr<ResumeCache> resumeCache,
     std::shared_ptr<Stats> stats,
     std::unique_ptr<KeepaliveTimer> keepaliveTimer,
     ProtocolVersion protocolVersion) {
   std::unique_ptr<ReactiveSocket> socket(new ReactiveSocket(
       ReactiveSocketMode::CLIENT,
       std::move(handler),
+      std::move(resumeCache),
       std::move(stats),
       std::move(keepaliveTimer),
       executor));
@@ -111,6 +116,7 @@ ReactiveSocket::disconnectedServer(
   std::unique_ptr<ReactiveSocket> socket(new ReactiveSocket(
       ReactiveSocketMode::SERVER,
       std::move(handler),
+      std::make_shared<ResumeCache>(stats),
       std::move(stats),
       nullptr,
       executor));
@@ -136,6 +142,25 @@ void ReactiveSocket::requestStream(
   checkNotClosed();
   connection_->streamsFactory().createStreamRequester(
       std::move(request), std::move(responseSink));
+}
+
+bool ReactiveSocket::requestStream(
+    Payload request,
+    yarpl::Reference<yarpl::flowable::Subscriber<Payload>> responseSink,
+    std::string streamName) {
+  debugCheckCorrectExecutor();
+  checkNotClosed();
+  auto streamId = connection_->getStreamId(streamName);
+  if (streamId == 0) {
+    // new stream
+    streamId = connection_->streamsFactory().createStreamRequester(
+        std::move(request), std::move(responseSink), executor_);
+    connection_->setStreamName(streamId, streamName);
+    return true;
+  } else {
+    // a stream is already present with the same name 
+    return false;
+  }
 }
 
 void ReactiveSocket::requestResponse(
--- a/src/temporary_home/ReactiveSocket.cpp
+++ b/src/temporary_home/ReactiveSocket.cpp
@@ -40,6 +40,7 @@ ReactiveSocket::ReactiveSocket(
           std::move(resumeCache))),
       executor_(executor) {
   debugCheckCorrectExecutor();
+  connection_->resumeStreamsFromCache();
   connection_->stats().socketCreated();
 }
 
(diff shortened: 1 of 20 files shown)


src/ReactiveSocket.cpp
```diff
@@ -129,13 +135,14 @@ std::shared_ptr<Subscriber<Payload>> ReactiveSocket::requestChannel(
       std::move(responseSink), executor_);
 }
 
-void ReactiveSocket::requestStream(
+StreamId ReactiveSocket::requestStream(
     Payload request,
```
lexs: Why do I need to provide this if we know we are resuming?

test/integration/ColdResumptionTest.cpp
```diff
@@ -0,0 +1,192 @@
...
+  std::condition_variable cv;
+  std::unique_lock<std::mutex> lk(cvM);
+
+  // Get a few subscriptions (happens right after connecting)
+  EXPECT_CALL(*mySub, onSubscribe_()).WillOnce(Invoke([&]() {
+    mySub->request(3);
+  }));
+  EXPECT_CALL(*mySub, onNext_("1"));
+  EXPECT_CALL(*mySub, onNext_("2"));
+  EXPECT_CALL(*mySub, onNext_("3")).WillOnce(Invoke([&](std::string) {
```
lehecka: should we expect call to onComplete or onError since we close the RS instance with this subscriber?

somasun:
I think the latest update to the PR has addressed most of the concerns.  I will address TODO(3) after https://github.com/rsocket/rsocket-cpp/pull/423 gets in.  I will send separate PRs for TODO(4) and TODO(5).  TODO(1) and TODO(2) are not needed right away.

somasun:
Rebased on top of yarpl changes.  ResumeCache::getResumableStreams will change once #423 is in.

test/integration/WarmResumptionTest.cpp
```diff
@@ -66,7 +66,7 @@ TEST_F(ServerFixture, DISABLED_BasicWarmResumption) {
   clientEvb->runInEventBaseThreadAndWait([&]() { rsocket->disconnect(); });
 
   // This request should be buffered
-  mySub->request(2);
+  clientEvb->runInEventBaseThreadAndWait([&]() { mySub->request(2); });
```
lehecka: Thanks for fixing this. I will have a class which helps with the threading model and will test it here.

lehecka:
the changes conflict with Ben's PR. Can you please rebase this PR on top of his changes when he lands it and pull it? Currently there is a known problem with TCK tests (threading model is broken now)

src/temporary_home/ReactiveSocket.h
```diff
@@ -80,6 +82,11 @@ class ReactiveSocket {
       Payload payload,
       yarpl::Reference<yarpl::flowable::Subscriber<Payload>> responseSink);
 
+  bool requestStream(
+      Payload payload,
+      yarpl::Reference<yarpl::flowable::Subscriber<Payload>> responseSink,
+      std::string streamName);
```
lexs: I dislike the term `streamName`, I liked the name `streamToken` but maybe there's something more descriptive.

src/temporary_home/RequestHandler.h
```diff
@@ -29,6 +29,10 @@ class RequestHandler {
       StreamId streamId,
       const yarpl::Reference<yarpl::flowable::Subscriber<Payload>>& response) noexcept = 0;
 
+  /// Handles a Stream resumed after cold restart
+  virtual yarpl::Reference<yarpl::flowable::Subscriber<Payload>> handleResumeStream(
```
lexs: I would prefer passing this callback in the resume() call instead of having it here, but we can fix that up later

src/temporary_home/ResumeCache.cpp
```diff
@@ -132,6 +132,33 @@ bool ResumeCache::isPositionAvailable(
...
+  /// This is a little inefficient.  But is called very infrequently
+  for (const auto& it : streamNameMap_) {
+    if (it.second == streamName) {
+      return it.first;
+    }
+  }
+  return 0;
+}
+
+std::unordered_map<StreamId, std::string /* streamName */>
```
lexs: This can probably be `std::vector<Stream>` or something as order doesn't seem important. Gets rid of the `std::pair` name problem too.

src/temporary_home/ReactiveSocket.cpp
```diff
@@ -138,6 +145,25 @@ void ReactiveSocket::requestStream(
       std::move(request), std::move(responseSink));
 }
 
+bool ReactiveSocket::requestStream(
+    Payload request,
+    yarpl::Reference<yarpl::flowable::Subscriber<Payload>> responseSink,
+    std::string streamName) {
+  debugCheckCorrectExecutor();
+  checkNotClosed();
+  auto streamId = connection_->getStreamId(streamName);
```
lexs: Should this just be a CHECK()? Seems like programmer error, could also rename to `hasStreamId()`.

src/temporary_home/ResumeCache.cpp
lexs: Or a `using StreamInfo = std::pair<...>` instead of concrete class.

src/temporary_home/ResumeCache.h
```diff
@@ -73,6 +73,15 @@ class ResumeCache {
     return clientPosition <= impliedPosition_;
   }
 
+  void setStreamName(StreamId streamId, const std::string& streamName) {
```
lexs: take by copy, emplace

src/temporary_home/ResumeCache.h
```diff
@@ -95,6 +104,8 @@ class ResumeCache {
 
   std::unordered_map<StreamId, ResumePosition> streamMap_;
 
+  std::unordered_map<StreamId, std::string /* streamName */> streamNameMap_;
```
lexs: I think I would prefer this being something like `pendingStreams_` that where the name only exists until the first frame is send (at that point it's added to the streamMap). Thoughts?

src/temporary_home/RequestHandler.h
```diff
@@ -29,6 +29,10 @@ class RequestHandler {
...
+      std::string streamName) noexcept = 0;
```
lexs: should be `const std::string&`

somasun:
closing this.  Most of the code is obsolete here

PR #396, opened 2017-05-07
Simplify Observers::create declarations with a helper type alias

vitaut:
The proposed PR simplifies `Observers::create` method declarations and reduces duplication when checking callback signatures by moving all the checks into a separate type alias. This makes public API of `Observers` cleaner while providing the same diagnostics if the signatures are not compatible:

```
rsocket-cpp/experimental/yarpl/test/Observable_test.cpp:32:9: error: no matching function for call to 'create'
        Observers::create<int>([&v](const int* value) { v.push_back(*value); }));
        ^~~~~~~~~~~~~~~~~~~~~~
rsocket-cpp/experimental/yarpl/include/yarpl/observable/Observers.h:28:13: note: candidate template ignored: disabled by 'enable_if' [with T = int, Next = (lambda at
      /Users/viz/work/rsocket-cpp/experimental/yarpl/test/Observable_test.cpp:32:32)]
            std::is_callable<Next(T), void>::value &&
            ^
rsocket-cpp/experimental/yarpl/include/yarpl/observable/Observers.h:48:15: note: candidate function template not viable: requires 2 arguments, but 1 was provided
  static auto create(
              ^
rsocket-cpp/experimental/yarpl/include/yarpl/observable/Observers.h:61:15: note: candidate function template not viable: requires 3 arguments, but 1 was provided
  static auto create(
              ^
1 error generated.
```

### DIFF
--- a/experimental/yarpl/include/yarpl/observable/Observers.h
+++ b/experimental/yarpl/include/yarpl/observable/Observers.h
@@ -12,15 +12,28 @@ namespace observable {
 
 /// Helper methods for constructing subscriber instances from functions:
 /// one, two, or three functions (callables; can be lamda, for instance)
-/// may be specified, corresponding to onNext, onError and onSubscribe
+/// may be specified, corresponding to onNext, onError and onComplete
 /// method bodies in the subscriber.
 class Observers {
+ private:
+  /// Defined if Next, Error and Complete are signature-compatible with
+  /// onNext, onError and onComplete subscriber methods respectively.
+  template <
+      typename T,
+      typename Next,
+      typename Error = void (*)(std::exception_ptr),
+      typename Complete = void (*)()>
+  using EnableIfCompatible =
+    typename std::enable_if<
+            std::is_callable<Next(T), void>::value &&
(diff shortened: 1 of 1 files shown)


experimental/yarpl/include/yarpl/observable/Observers.h
```diff
@@ -12,15 +12,28 @@ namespace observable {
...
+  /// onNext, onError and onComplete subscriber methods respectively.
+  template <
+      typename T,
+      typename Next,
+      typename Error = void (*)(std::exception_ptr),
+      typename Complete = void (*)()>
+  using EnableIfCompatible =
+    typename std::enable_if<
+            std::is_callable<Next(T), void>::value &&
+            std::is_callable<Error(const std::exception_ptr), void>::value &&
```
alexmalyshev: I know that this is copied from elsewhere, but I don't think that `const` in `const std::exception_ptr` is necessary?

lehecka:
will close and reopen to spin another round of travis build

vitaut:
Ditched unnecessary `const` as suggested by Alex.

PR #399, opened 2017-05-08
Fix chained operators reference leak (take 2).

nickgg:
I'm hoping #389 fixed the asan issue.


nickgg:
result! thanks @alexmalyshev

PR #400, opened 2017-05-08
Encapsulate frame management into connectionAutomaton

ragansa:
Brings frame management/serialization/de-serialization inside ConnectionAutomaton and out of the Requester/Responder classes

### DIFF
--- a/src/ConnectionAutomaton.cpp
+++ b/src/ConnectionAutomaton.cpp
@@ -382,7 +382,7 @@ void ConnectionAutomaton::processFrameImpl(
   }
   auto streamId = *streamIdPtr;
   if (streamId == 0) {
-    onConnectionFrame(std::move(frame));
+    handleConnectionFrame(frameType, std::move(frame));
     return;
   }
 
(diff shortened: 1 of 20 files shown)


src/ConnectionAutomaton.cpp
```diff
@@ -568,8 +561,75 @@ void ConnectionAutomaton::onConnectionFrame(
...
+  auto it = streamState_->streams_.find(streamId);
+  if (it == streamState_->streams_.end()) {
+    handleUnknownStream(streamId, frameType, std::move(serializedFrame));
+    return;
+  }
+  auto &automaton = it->second;
+
+  switch (frameType) {
+    case FrameType::REQUEST_N: {
+      Frame_REQUEST_N frame_request_n;
```
lehecka: nit: let's use the same coding style and name the variables inCammelCase.

src/ConnectionAutomaton.h
```diff
@@ -186,7 +186,7 @@ class ConnectionAutomaton final
     return streamsFactory_;
   }
 
-  FrameSerializer& frameSerializer() const override;
+  FrameSerializer& frameSerializer() const;
```
lehecka: can you try to remove this method? I don't think it is used anywhere now.

PR #402, opened 2017-05-09
Remove unneeded uses of virtual, add override where possible

alexmalyshev:
override implies virtual.  Otherwise it doesn't compile.


lehecka:
thanks for the cleanup

PR #403, opened 2017-05-09
Channel Support (unproven)

benjchristensen:
Adding channel to the new APIs. This is not yet proven as I'm getting an exception when running the client/server example and have not debugging it. Posting for review before I spend that time.

### DIFF
--- a/examples/channel-hello-world/TARGETS
+++ b/examples/channel-hello-world/TARGETS
@@ -1,0 +1,28 @@
+cpp_binary(
+  name = 'tcp-server',
+  headers = subdir_glob([('.', '**/*.h')]),
+  srcs = [
+    'HelloStreamRequestHandler.cpp',
+    'HelloStreamSubscription.cpp',
+    'StreamHelloWorld_Server.cpp',
+  ],
+  deps = [
+    '@/folly/init:init',
+    '@/lithium/reactivesocket-cpp:rsocket',
+  ],
+)
+
+cpp_binary(
+  name = 'tcp-client',
+  headers = subdir_glob([('.', '**/*.h')]),
+  srcs = [
+    'HelloStreamRequestHandler.cpp',
+    'HelloStreamSubscription.cpp',
+    'StreamHelloWorld_Client.cpp',
+  ],
+  deps = [
+    '@/folly/init:init',
+    '@/lithium/reactivesocket-cpp/examples/util:subscriber',
+    '@/lithium/reactivesocket-cpp:rsocket',
+  ],
+)
--- a/experimental/rsocket-src/RSocketRequester.cpp
+++ b/experimental/rsocket-src/RSocketRequester.cpp
@@ -37,10 +37,40 @@ RSocketRequester::~RSocketRequester() {
   LOG(INFO) << "RSocketRequester => destroy";
 }
 
-std::shared_ptr<Subscriber<Payload>> RSocketRequester::requestChannel(
-    std::shared_ptr<Subscriber<Payload>> responseSink) {
-  // TODO need to runInEventBaseThread like other request methods
-  return reactiveSocket_->requestChannel(std::move(responseSink));
+yarpl::Reference<yarpl::flowable::Flowable<reactivesocket::Payload>>
+RSocketRequester::requestChannel(
+    yarpl::Reference<yarpl::flowable::Flowable<reactivesocket::Payload>>
+        payloads) {
+  auto& eb = eventBase_;
+  auto srs = reactiveSocket_;
+
+    LOG(INFO) << "requestChannel executing ";
+
+  return yarpl::flowable::Flowables::fromPublisher<Payload>([
+    &eb,
+    request = std::move(payloads),
+    srs = std::move(srs)
+  ](yarpl::Reference<yarpl::flowable::Subscriber<Payload>> subscriber) mutable {
+    // TODO eliminate OldToNew bridge
+    auto os = std::make_shared<OldToNewSubscriber>(std::move(subscriber));
+
+
+      LOG(INFO) << "requestChannel ABOUT TO RUN ON THREAD ";
+
+    eb.runInEventBaseThread([
+      request = std::move(request),
+      os = std::move(os),
+      srs = std::move(srs)
+    ]() mutable {
+
+        LOG(INFO) << "requestChannel RUNNING IN THREAD";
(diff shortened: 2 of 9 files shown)


examples/channel-hello-world/TARGETS
```diff
@@ -0,0 +1,28 @@
+cpp_binary(
+  name = 'tcp-server',
+  headers = subdir_glob([('.', '**/*.h')]),
+  srcs = [
+    'HelloStreamRequestHandler.cpp',
+    'HelloStreamSubscription.cpp',
+    'StreamHelloWorld_Server.cpp',
```
alexmalyshev: Did you mean to copy the TARGETS file over?  If so then we should at least do `s|Stream|Channel|g` on them.

experimental/rsocket-src/RSocketRequester.cpp
```diff
@@ -37,10 +37,40 @@ RSocketRequester::~RSocketRequester() {
...
-  // TODO need to runInEventBaseThread like other request methods
-  return reactiveSocket_->requestChannel(std::move(responseSink));
+yarpl::Reference<yarpl::flowable::Flowable<reactivesocket::Payload>>
+RSocketRequester::requestChannel(
+    yarpl::Reference<yarpl::flowable::Flowable<reactivesocket::Payload>>
+        payloads) {
+  auto& eb = eventBase_;
+  auto srs = reactiveSocket_;
+
+    LOG(INFO) << "requestChannel executing ";
```
alexmalyshev: Nit: Indent.

experimental/rsocket-src/RSocketRequester.cpp
```diff
@@ -37,10 +37,40 @@ RSocketRequester::~RSocketRequester() {
...
+
+  return yarpl::flowable::Flowables::fromPublisher<Payload>([
+    &eb,
```
alexmalyshev: Doing this along with `auto& eb = eventBase_;` makes me think you should just capture `this` by value and use `eventBase_` directly in the lambda.  It'll be less confusing I think.

experimental/rsocket-src/RSocketRequester.cpp
```diff
@@ -37,10 +37,40 @@ RSocketRequester::~RSocketRequester() {
...
+    request = std::move(payloads),
+    srs = std::move(srs)
+  ](yarpl::Reference<yarpl::flowable::Subscriber<Payload>> subscriber) mutable {
+    // TODO eliminate OldToNew bridge
+    auto os = std::make_shared<OldToNewSubscriber>(std::move(subscriber));
+
+
+      LOG(INFO) << "requestChannel ABOUT TO RUN ON THREAD ";
```
alexmalyshev: Nit: Indents on LOGs.

experimental/rsocket/RSocketRequestHandler.h
```diff
@@ -31,6 +32,29 @@ class RSocketRequestHandler {
   virtual yarpl::Reference<yarpl::flowable::Flowable<reactivesocket::Payload>>
   handleRequestStream(
       reactivesocket::Payload request,
-      reactivesocket::StreamId streamId) = 0;
+      reactivesocket::StreamId streamId) {
+    return yarpl::flowable::Flowables::error<reactivesocket::Payload>(
+        std::make_exception_ptr(
```
alexmalyshev: You can just pass in the `std::logic_error` directly.  There's an `Flowables::error<T, ExnType>` overload.

experimental/rsocket/RSocketRequestHandler.h
```diff
@@ -31,6 +32,29 @@ class RSocketRequestHandler {
...
+     * @param streamId
+     * @return
+     */
+  virtual yarpl::Reference<yarpl::flowable::Flowable<reactivesocket::Payload>>
+  handleRequestChannel(
+      yarpl::Reference<yarpl::flowable::Flowable<reactivesocket::Payload>>
+          requestStream,
+      reactivesocket::StreamId streamId) {
+    return yarpl::flowable::Flowables::error<reactivesocket::Payload>(
+        std::make_exception_ptr(
```
alexmalyshev: Ditto.

lehecka:
I will work on this. Thanks for starting this.

examples/channel-hello-world/TARGETS
benjchristensen: I saw them, just moved them along for the ride assuming we'd fix when they become needed. Yes, I should have renamed them and forgot to :-/

experimental/rsocket-src/RSocketRequester.cpp
benjchristensen: Odd. Don't know why that's like that.

experimental/rsocket-src/RSocketRequester.cpp
benjchristensen: My reasoning is that it is easy to end up capturing more by accident when using `this`. Is it a better convention to use `this` generally?

experimental/rsocket/RSocketRequestHandler.h
benjchristensen: Oh nice. Thanks.

experimental/rsocket-src/RSocketRequester.cpp
alexmalyshev: Capturing one object field by reference is effectively the same as capturing the entire `this` pointer as you're depending on the entire object not being destroyed.  You're not really capturing the other objects fields by accident, whether they are or aren't in scope is the same question as "is eventBase_ in scope".

The issue I take with doing it this way is that it's harder to see the dependence on the object, as it requires jumping through `[..., &eb, ...]` and `auto& eb = eventBase_;`.

lehecka: I am suggesting to capture eb = &eventBase_ which will capture pointer (by value). I also dislike capturing this because it brings questions about lifetime. Capturing pointer to EventBase should be fine since all executors should outlive everything else.

examples/channel-hello-world/TARGETS
lehecka: TARGETS will need to be updated after the big refactor anyways.

benjchristensen:
Finished and merged in https://github.com/rsocket/rsocket-cpp/pull/424

PR #407, opened 2017-05-10
Update Observables to match Flowables, make it actually work

alexmalyshev:
It didn't really compile in the first place...  Copy some Flowable tests over to
use Observable.


lehecka:
thanks for cleaning up the mess. I think we copy/pasted way too much code from Flowables. This change makes sense to me.

PR #417, opened 2017-05-10
Bump up folly version to v2017.05.08.00

alexmalyshev:


alexmalyshev:
Hmm, the compilation errors are weird, but I don't have the time to look into them right now.

PR #419, opened 2017-05-12
Flowables::fromGenerator operator

lehecka:
This will allow using more Flowable operators with non-copyable T.


experimental/yarpl/include/yarpl/Flowables.h
```diff
@@ -108,6 +108,25 @@ class Flowables {
...
   }
 
+  template <typename T, typename TGenerator>
+  static Reference<Flowable<T>> fromGenerator(TGenerator generator) {
+    auto lambda = [generator = std::move(generator)]
+    (Subscriber<T>& subscriber, int64_t requested) {
+      int64_t generated = 0;
+      try {
+        while (generated < requested) {
+          subscriber.onNext(generator());
```
lexs: if the subscriber calls cancel() in the onNext() won't this still deliver another one?

experimental/yarpl/include/yarpl/Flowables.h
```diff
@@ -108,6 +108,25 @@ class Flowables {
...
+          ++generated;
+        }
+        return std::make_tuple(static_cast<int64_t>(generated), false);
```
lexs: this isn't related to this diff, but I find this API weird, why does it have to return how many onNext it produced and if it was completed. both of those could be inferred by passing a `Subscriber` that keeps track

alexmalyshev: This was written by @vjn.  It's how `Flowable` keeps track of its credits / whether to emit or not.  It's meant to be agnostic to whatever `Subscriber` you pass in.

lehecka: yeah, we discussed that. I asked the same question originally. There was a push back on creating a wrapper around every subscriber to keep track of the count.

experimental/yarpl/include/yarpl/Flowables.h
lehecka: @alexmalyshev Should I just emit one value at a time and let Flowable go thru the lambda multiple times? Flowables::range has the same problem. We can generate values even after cancel.
Technically speaking that is ok for the reactive streams protocol. It says after cancel we may receive some more values...

alexmalyshev: I think we should leave this as is.  The spec has no guarantees about synchronizing `onNext()` with `cancel()`, and I don't think there's a great solution that fits all client use-cases.

experimental/yarpl/include/yarpl/Flowables.h
```diff
@@ -108,6 +108,25 @@ class Flowables {
...
+      } catch(const std::exception&) {
+        subscriber.onError(std::current_exception());
+        return std::make_tuple(static_cast<int64_t>(generated), true);
```
alexmalyshev: `generated` is already an `int64_t`.  No need for the `static_cast`.

experimental/yarpl/include/yarpl/Flowables.h
```diff
@@ -108,6 +108,25 @@ class Flowables {
...
+    auto lambda = [generator = std::move(generator)]
+    (Subscriber<T>& subscriber, int64_t requested) {
+      int64_t generated = 0;
+      try {
+        while (generated < requested) {
+          subscriber.onNext(generator());
+          ++generated;
+        }
+        return std::make_tuple(static_cast<int64_t>(generated), false);
+      } catch(const std::exception&) {
```
alexmalyshev: I think we should also have a case for `catch (...)`, and we can just pass through some default `std::runtime_error("Don't do that")` or something.

experimental/yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -51,6 +51,8 @@ class FlowableOperator : public Flowable<D> {
       subscriber_->onSubscribe(Reference<::yarpl::flowable::Subscription>(this));
     }
 
+    void onNext(U) override {}
```
alexmalyshev: If you think `Subscriber` is too abstract to have a default implementation of `onNext()` then I think `FlowableOperator` shouldn't have one either.

experimental/yarpl/include/yarpl/Flowables.h
lexs: It seems like the wrapper would result in the exact same code after optimization (it would be templated on the lambda). This API makes it easy to fuck up by either returning the wrong number (off-by-one seems easy) or `false` instead of `true` etc.

lehecka:
replacing with https://github.com/rsocket/rsocket-cpp/pull/461

PR #422, opened 2017-05-12
Make ServerConnectionAcceptor thread-safe

alexmalyshev:
Right now it's always tied to a specific worker thread.  The issue we hit is
that there is no synchronous way of closing all FrameTransports held by the
ServerConnectionAcceptor, and waiting for all their associated events to drain
from the worker thread's EventBase.

This diff adds ServerConnectionAcceptor::stop().  This method calls close() on
each FrameTransport, and then waits on a newly created folly::Baton until the
transport set is empty.  To make this truly safe, we need the callbacks from
the ConnectionHandler to finish executing before the transport is removed
from the set.

Using that, we can make ~RSocketServer thread-safe as well.  I've enabled
RSocketClientServer.SimpleConnect, and run it 50 times in parallel on my machine
successfully.

### DIFF
--- a/experimental/rsocket-src/RSocketServer.cpp
+++ b/experimental/rsocket-src/RSocketServer.cpp
@@ -47,11 +47,14 @@ RSocketServer::RSocketServer(
       acceptor_(ProtocolVersion::Unknown) {}
 
 RSocketServer::~RSocketServer() {
-  // Stop accepting new connections.
+  // Will stop forwarding connections from lazyAcceptor_ to acceptor_.
+  isShutdown_ = true;
+
+  // Stop accepting new connections from the network.
   lazyAcceptor_->stop();
 
-  // FIXME(alexanderm): This is where we /should/ close the FrameTransports
-  // sitting around in the ServerConnectionAcceptor, but we can't yet...
+  // Close out all outstanding FrameTransports in acceptor_.
+  acceptor_.stop();
 
   // Asynchronously close all existing ReactiveSockets.  If there are none, then
   // we can do an early exit.
@@ -92,8 +95,9 @@ void RSocketServer::start(OnAccept onAccept) {
                   folly::Executor& executor) {
         LOG(INFO) << "Going to accept duplex connection";
 
-        // FIXME(alexanderm): This isn't thread safe
-        acceptor_.accept(std::move(conn), connectionHandler_);
+        if (isShutdown_) {
+          acceptor_.accept(std::move(conn), connectionHandler_);
+        }
       })
       .onError([](const folly::exception_wrapper& ex) {
         LOG(FATAL) << "Failed to start ConnectionAcceptor: " << ex.what();
--- a/src/ServerConnectionAcceptor.cpp
+++ b/src/ServerConnectionAcceptor.cpp
@@ -9,13 +9,9 @@
 #include "src/FrameTransport.h"
 #include "src/Stats.h"
 
-#include <iostream>
-
 namespace reactivesocket {
 
-class OneFrameProcessor
-    : public FrameProcessor,
-      public std::enable_shared_from_this<OneFrameProcessor> {
+class OneFrameProcessor : public FrameProcessor {
  public:
   OneFrameProcessor(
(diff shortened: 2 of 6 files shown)


alexmalyshev:
cc: @lexs

alexmalyshev:
Note: the important bit here is the `To make this truly safe, we need the callbacks from
the ConnectionHandler to finish executing before the transport is removed
from the set.` part.  I'm a little worried about holding a lock on the transport set while running `ConnectionHandler` callbacks, but I can't see how we can not do that and have safe shutdown.

experimental/rsocket-src/RSocketServer.cpp
```diff
@@ -92,8 +95,9 @@ void RSocketServer::start(OnAccept onAccept) {
                   folly::Executor& executor) {
         LOG(INFO) << "Going to accept duplex connection";
 
-        // FIXME(alexanderm): This isn't thread safe
-        acceptor_.accept(std::move(conn), connectionHandler_);
+        if (isShutdown_) {
```
lexs: `!isShutdown_`?

How did this pass any tests? :) :(

src/ServerConnectionAcceptor.cpp
```diff
@@ -159,26 +156,55 @@ void ServerConnectionAcceptor::processFrame(
...
-          connectionHandler,
-          std::move(transport),
-          std::runtime_error("invalid"));
+      error("Invalid frame received instead of SETUP/RESUME");
       break;
     }
   }
 }
 
+////////////////////////////////////////////////////////////////////////////////
```
lexs: why?

src/ServerConnectionAcceptor.cpp
```diff
@@ -159,26 +156,55 @@ void ServerConnectionAcceptor::processFrame(
...
+    auto locked = connections_.lock();
+
+    shutdown_.emplace();
+
+    if (locked->empty()) {
+      return;
+    }
+
+    for (auto& connection : *locked) {
+      connection->close(std::runtime_error("ServerConnectionAcceptor closed"));
```
lexs: This will touch the FrameTransport on a different thread that it was created on, which isn't safe.

src/ServerConnectionAcceptor.cpp
alexmalyshev: I use those `/ * 80` comments as logical separators, i.e. separating fields from methods, internal methods from public ones, ctors from methods, etc.  Though I do have a tendency to put in more than I should...  I can remove it if you want.

experimental/rsocket-src/RSocketServer.cpp
alexmalyshev: *facepalm*  How did it pass tests indeed...

src/ServerConnectionAcceptor.cpp
alexmalyshev: Is that because of not having any guarantees about the subscriber/subscription pair inside it?  Cause the `FrameTransport` object itself does everything under a lock.

experimental/rsocket-src/RSocketServer.cpp
alexmalyshev: Oh I remember now, there are no tests that actually exercise `RSocketServer` serving requests.  This is because up until this diff, there was no safe way to run ~RSocketServer after it had started serving connections.

src/ServerConnectionAcceptor.cpp
lexs: Let's remove them, they're existence is quite subjective and I personally don't feel like they add much.

src/ServerConnectionAcceptor.cpp
lexs: In this case it might be safe (I honestly don't know), but that's probably more accidental than anything else. I know we're trying to kill `FrameTransport` too which will probably change this. cc @lehecka

lehecka:
I will take over this PR and fix it.

lehecka:
closing as this PR was replaced by https://github.com/rsocket/rsocket-cpp/pull/486

PR #423, opened 2017-05-13
Simplify ResumeCache

somasun:
This PR has two changes
- Remove the `map<StreamId, ResumePosition> streamMap_` from `ResumeCache`.
This map was never used.
- Add a set of `activeStreams_` in `ResumeCache`.

The main reason for doing this is driven by cold-resumption.  In cold-resumption,
we need to know all the streams which are active when the application restarts.

`streamMap_` contained an entry only if there are some frames buffered for a
stream.  If all the frames of a stream have been delivered to the other side,
the stream entry was removed.   And since `streamMap_` was not being used
anywhere else in RSocket, I have removed all the associated code.  This also
simplifies code.

On the other hand, `activeStreams_` stores StreamIds of all the streams which
are currently active.  The entries get removed only when the streams are
closed.  This can directly be used in cold resumption.  This is currently done 
only for REQUEST_STREAM

### DIFF
--- a/src/statemachine/RSocketStateMachine.cpp
+++ b/src/statemachine/RSocketStateMachine.cpp
@@ -378,10 +378,6 @@ void RSocketStateMachine::processFrameImpl(
 
   auto streamIdPtr = frameSerializer_->peekStreamId(*frame);
 
-  // TODO(tmont): If a frame is invalid, it will still be tracked. However, we
-  // actually want that. We want to keep
-  // each side in sync, even if a frame is invalid.
-  resumeCache_->trackReceivedFrame(*frame, frameType, streamIdPtr);
 
   if (!streamIdPtr) {
     // Failed to deserialize the frame.
@@ -389,6 +385,7 @@ void RSocketStateMachine::processFrameImpl(
     return;
   }
   auto streamId = *streamIdPtr;
+  resumeCache_->trackReceivedFrame(*frame, frameType, streamId);
   if (streamId == 0) {
     handleConnectionFrame(frameType, std::move(frame));
     return;
--- a/src/statemachine/RSocketStateMachine.cpp
+++ b/src/statemachine/RSocketStateMachine.cpp
@@ -313,6 +313,7 @@ bool RSocketStateMachine::endStreamInternal(
     StreamId streamId,
     StreamCompletionSignal signal) {
   VLOG(6) << "endStreamInternal";
+  resumeCache_->rmFromActiveStreams(streamId);
   auto it = streamState_->streams_.find(streamId);
   if (it == streamState_->streams_.end()) {
     // Unsubscribe handshake initiated by the connection, we're done.
--- a/src/statemachine/RSocketStateMachine.cpp
+++ b/src/statemachine/RSocketStateMachine.cpp
@@ -377,8 +377,6 @@ void RSocketStateMachine::processFrameImpl(
   stats_->frameRead(frameType);
 
   auto streamIdPtr = frameSerializer_->peekStreamId(*frame);
-
-
   if (!streamIdPtr) {
     // Failed to deserialize the frame.
     closeWithError(Frame_ERROR::invalidFrame());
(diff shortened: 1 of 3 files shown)


src/ResumeCache.h
```diff
@@ -93,7 +96,8 @@ class ResumeCache {
   // Inferred position of the rcvd frames
   ResumePosition impliedPosition_{0};
 
-  std::unordered_map<StreamId, ResumePosition> streamMap_;
+  // Only active REQUEST_STREAMs are preserved here
+  std::set<StreamId> activeStreams_;
```
lexs: `unordered_set`?

src/ResumeCache.h
```diff
@@ -77,6 +76,10 @@ class ResumeCache {
     return size_;
   }
 
+  void rmFromActiveStreams(StreamId streamId) {
```
lexs: `removeStream()`? instead, or `onStreamClosed()`

src/ResumeCache.h
```diff
@@ -93,7 +96,8 @@ class ResumeCache {
   // Inferred position of the rcvd frames
   ResumePosition impliedPosition_{0};
 
-  std::unordered_map<StreamId, ResumePosition> streamMap_;
+  // Only active REQUEST_STREAMs are preserved here
```
lexs: this isn't only STREAM right? Also channel etc

src/ResumeCache.cpp
```diff
@@ -62,6 +62,11 @@ void ResumeCache::trackSentFrame(
     // TODO(tmont): this could be expensive, find a better way to get length
     auto frameDataLength = serializedFrame.computeChainDataLength();
 
+    if (streamIdPtr) {
```
lexs: this should probably switch on `frameType` and only do this for `REQUEST_*` frames.

somasun:
Thanks for the review @lexs.  Updated the PR.
I went with `set` instead of `unordered_set`, so that I can preserve order and also makes it easier to write unittests.  Also, the performance isn't really different for small number of elements.  
The `activeStreams_` keeps track of only STREAMs and not CHANNEL.  I am targeting the first implementation of cold-resumption only for STREAMs

lexs:
>  I am targeting the first implementation of cold-resumption only for STREAMs

This is concerning, does this mean you'll cancel all streams after resumption or not know about them at all?

src/ConnectionAutomaton.cpp
```diff
@@ -375,12 +376,13 @@ void ConnectionAutomaton::processFrameImpl(
   auto frameType = frameSerializer_->peekFrameType(*frame);
   stats_->frameRead(frameType);
 
+  auto streamIdPtr = frameSerializer_->peekStreamId(*frame);
+
   // TODO(tmont): If a frame is invalid, it will still be tracked. However, we
   // actually want that. We want to keep
   // each side in sync, even if a frame is invalid.
-  resumeCache_->trackReceivedFrame(*frame, frameType);
+  resumeCache_->trackReceivedFrame(*frame, frameType, streamIdPtr);
```
lexs: Stream id should probably not be Optional, just put this after the `if` below (no stream id is a connection-level error).

lexs: Sorry for missing this previously.

src/ResumeCache.cpp
```diff
@@ -46,7 +46,13 @@ ResumeCache::~ResumeCache() {
 
 void ResumeCache::trackReceivedFrame(
     const folly::IOBuf& serializedFrame,
-    const FrameType frameType) {
+    const FrameType frameType,
+    const folly::Optional<StreamId> streamIdPtr) {
+  if (frameType == FrameType::REQUEST_STREAM && streamIdPtr) {
```
lexs: As long as this doesn't break warm resumption for channel, which is something we'll need. Note we'll need cold start resumption to also work in that it cancels the channel in that case.

src/ResumeCache.h
lexs: Why does a sorted set make testing easier? If you need it sorted you can just do that in the test?

lehecka:
accepting and pulling in to make sure this doesn't drift away..

PR #430, opened 2017-05-16
RSocketRequestHandler -> RSocketResponder

benjchristensen:
Name change to match protocol terminology


benjchristensen:
Forcing merge after asking Ondrej to avoid conflict with next changes about to work on.

PR #432, opened 2017-05-16
stream automata uses yarpl::Subscription/Subscriber

lehecka:
At this moment everything compiles. There are unit tests which are broken.
I also need to have a solution for concurrency tests.
Early feedback is welcome.

UPDATE: Unit tests are passing and leaking memory. I am still working on this. To unblock Ben, I am going to merge and work on the unit tests separately.


experimental/rsocket-src/RSocketConnectionHandler.cpp
```diff
@@ -22,20 +23,19 @@ class RSocketHandlerBridge : public reactivesocket::DefaultRequestHandler {
...
   }
 
-  std::shared_ptr<Subscriber<Payload>> handleRequestChannel(
+  yarpl::Reference<yarpl::flowable::Subscriber<Payload>> handleRequestChannel(
       Payload request,
       StreamId streamId,
-      const std::shared_ptr<Subscriber<Payload>>& response) noexcept override {
-    auto eagerSubscriber = std::make_shared<EagerSubscriberBridge>();
+      const yarpl::Reference<yarpl::flowable::Subscriber<Payload>>& response) noexcept override {
+    auto eagerSubscriber = make_ref<EagerSubscriberBridge>();
```
benjchristensen: Why do we still need this bridge?

experimental/rsocket-src/RSocketConnectionHandler.cpp
```diff
@@ -45,29 +45,29 @@ class RSocketHandlerBridge : public reactivesocket::DefaultRequestHandler {
...
+    flowable->subscribe(std::move(response));
 
     return eagerSubscriber;
   }
 
   void handleRequestResponse(
       Payload request,
       StreamId streamId,
-      const std::shared_ptr<Subscriber<Payload>>&
+      const yarpl::Reference<yarpl::flowable::Subscriber<Payload>>&
```
benjchristensen: We can change this to take SingleObserver<Payload> instead of FlowableSubscriber and then eliminate the bridge. But that may need to be a followup change as I imagine that has far reaching impact lower down.

experimental/rsocket-src/RSocketRequester.cpp
```diff
@@ -69,28 +66,26 @@ RSocketRequester::requestStream(Payload request) {
...
-    ]() mutable { srs->requestStream(std::move(request), std::move(os)); });
+    ]() mutable { srs->requestStream(std::move(request), std::move(subscriber)); });
   });
 }
 
 yarpl::Reference<yarpl::single::Single<reactivesocket::Payload>>
 RSocketRequester::requestResponse(Payload request) {
   // TODO bridge in use until SingleSubscriber is used internally
-  class SingleToSubscriberBridge : public Subscriber<Payload> {
+  class SingleToSubscriberBridge : public yarpl::flowable::Subscriber<Payload> {
```
benjchristensen: After this change lands we can change requestResponse to use SingleObserver instead of FlowableSubscriber.

experimental/rsocket-src/RSocketConnectionHandler.cpp
benjchristensen: Discussed ... ignoring this for now. It's the hard-to-understand inverted API since it's eager.

src/ConnectionAutomaton.cpp
```diff
@@ -286,7 +286,7 @@ void ConnectionAutomaton::reconnect(
 
 void ConnectionAutomaton::addStream(
     StreamId streamId,
-    std::shared_ptr<StreamAutomatonBase> automaton) {
+    yarpl::Reference<StreamAutomatonBase> automaton) {
```
lexs: I don't see why we'd change this? Why use our home-made shared_ptr instead of shared_ptr here?

lexs:
I have a few concerns:
* Why are we using yarpl::Reference<> for non-yarpl types? It doesn't seem to bring us anything over shared_ptr?
* Could we use folly::ExceptionWrapper instead of std::exception_ptr? Otherwise we'll be forcing exception handling on everyone.

src/ConnectionAutomaton.cpp
lehecka: This is because the StreamAutomatonBase derived classes also derive from the yarpl::flowable::Subscriber/Subscription which inherit from yarpl::Refcounted. We need to use only one memory management mechanism. The common one is yarpl::Reference.

lehecka:
The main reason why we are using std::exception_ptr instead of folly::ExceptionWrapper is that yarpl doesn't depend on folly. We were debating pros and const of this and decided to avoid dependency on folly.
Practically std::exception_ptr is used just like folly::ExceptionWrapper. You can rethrow the exception if you want, but you also don't have to

benjchristensen:
Moving to https://github.com/rsocket/rsocket-cpp/pull/434

lexs:
`std::exception_ptr` requires rethrowing the exception to get the message, no way around that (exception an ExceptionWrapper).

PR #437, opened 2017-05-16
fixing memory leaking tests

lehecka:
The Subscription interface is now used in ReactiveSocket library and the special behavior around lifetime is unnecessary and awkward to use leading to memory leaks.
The fix also focuses on breaking the reference cycles between subscribers and subscriptions


alexmalyshev:
The change seems like a good idea to me, I never liked the `Subscription` reference cycle...

Build is still failing though, so please fix and re-test everything with ASAN+LSAN if you can.

lehecka:
this is not fixing all travis issues, but fixes the memory leaking. I will commit this now and work on fixing the rest of the travis issues.

PR #438, opened 2017-05-17
fixing compiler issues for gcc 6.0

lehecka:


lehecka:
the compiler error was: error: invalid use of incomplete type ‘class yarpl::single::Single<T>’ [-Werror]

benjchristensen:
Other failures are known ... merging so we can move forward and build on top of this.

PR #439, opened 2017-05-17
Make debug default in cmake

lehecka:
Please note that this is not changing our travis builds. This changes the default build mode for the local development.


lehecka:
don't need travis to verify it because this change is not for travis.

PR #440, opened 2017-05-17
Fix ASAN issue in unit test

lehecka:


benjchristensen:
Step towards being healthy again ...

PR #445, opened 2017-05-17
Purge Obsolete Unit Tests

benjchristensen:
Most tests are now obsolete as they use old Subscriber/Subscription types and old ReactiveSocket.h APIs.
It was decided that it is much easier to clean house and rewrite new unit tests against the new APIs, so purging and will start rewriting over next couple days (after a few more API changes this purge make easier).


benjchristensen:
ASAN failure on a yarpl test. I didn't touch yarpl :-/

PR #447, opened 2017-05-17
Fix TakeOperator::onNext() deleting itself too early

alexmalyshev:
It would call FlowableOperator::Subscription::onComplete() and then
FlowableOperator::Subscription::cancel().  This doesn't work as ::onComplete()
will call release(), and we are no longer safe running any more methods on the
object.

Instead run onComplete() manually on the subscriber, and then cancel the
subscription.

### DIFF
--- a/yarpl/include/yarpl/flowable/FlowableOperator.h
+++ b/yarpl/include/yarpl/flowable/FlowableOperator.h
@@ -219,8 +219,8 @@ class TakeOperator : public FlowableOperator<T, T> {
         FlowableOperator<T, T>::Subscription::subscriber_->onNext(
             std::move(value));
         if (limit_ == 0) {
+          FlowableOperator<T, T>::Subscription::subscriber_->onComplete();
           FlowableOperator<T, T>::Subscription::cancel();
-          FlowableOperator<T, T>::Subscription::onComplete();
         }
       }
     }


alexmalyshev:
Test failures are leaks coming from the rsocket tests.  The yarpl-tests binary seems to have run just fine.

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -219,8 +219,8 @@ class TakeOperator : public FlowableOperator<T, T> {
         FlowableOperator<T, T>::Subscription::subscriber_->onNext(
             std::move(value));
         if (limit_ == 0) {
+          FlowableOperator<T, T>::Subscription::subscriber_->onComplete();
```
phoad: I believe we should also update onComplete() method in the FlowableOperator to check if the Subscription is null. At least we might assert in the Debug build, if the subscription is null and if we call onComplete(). 

Furthermore, in FlowableOperator, we call Release multiple times. We might assert in the Release function also. Because it is reducing the reference count more than once per call.

lehecka:
I fixed the terminating signals for all operators via https://github.com/rsocket/rsocket-cpp/pull/456.
Closing this diff as it is unnecessary now.

PR #450, opened 2017-05-18
Fix build on Arch

alexmalyshev:
Missing gflags includes strikes again


benchmarks/RequestResponseLatency.cpp
```diff
@@ -9,6 +9,8 @@
 #include <condition_variable>
 #include <iostream>
 #include <thread>
+#include <gflags/gflags.h>
```
phoad: very nit: order of the include files.

lehecka: should be fixed on the next round of running clang-format

PR #451, opened 2017-05-18
Hopefully fix RSocketStateMachine leak

alexmalyshev:
Seems like we need to close it in ~RSocketRequester.  LSAN warnings stop for me
locally, but I'm not 100% sure of the logic here.

### DIFF
--- a/src/RSocketRequester.cpp
+++ b/src/RSocketRequester.cpp
@@ -35,6 +35,7 @@ RSocketRequester::RSocketRequester(
 
 RSocketRequester::~RSocketRequester() {
   LOG(INFO) << "RSocketRequester => destroy";
+  stateMachine_->close(folly::exception_wrapper(), StreamCompletionSignal::CONNECTION_END);
 }
 
 yarpl::Reference<yarpl::flowable::Flowable<rsocket::Payload>>


src/RSocketRequester.cpp
```diff
@@ -35,6 +35,7 @@ RSocketRequester::RSocketRequester(
 
 RSocketRequester::~RSocketRequester() {
   LOG(INFO) << "RSocketRequester => destroy";
+  stateMachine_->close(folly::exception_wrapper(), StreamCompletionSignal::CONNECTION_END);
```
phoad: stateMachine_'s destructor has a check 
  DCHECK(isDisconnectedOrClosed()); // the instance should be closed by via

Do you observe if this check is failing if you do not call the close() method in here? 
Should we call the close() method at the destructor of the stateMachine_ ?

I saw that in the close() method of the RSocketStateMachine, we call lots of callback functions. Any of these functions might fail and if might cause early termination of the function. But at the start of the function we call
  isClosed_ = true;
Should we place some try/catch in the close function and place the isClosed_ = true at the end of this function?

alexmalyshev: > Do you observe if this check is failing if you do not call the close() method in here?

The check doesn't fail because we leak the `RSocketStateMachine`, hence we never call `~RSocketStateMachine`.
Do you observe if this check is > Do you observe if this check is failing if you do not call the close() method in here?

The check doesn't fail because we leak the `RSocketStateMachine`, hence we never call `~RSocketStateMachine`.  This is because of the smart pointer cycle of
`FrameTransport <-> RSocketStateMachine`.

phoad: I see, so you break the cycle by calling the close().

lehecka: yep, this is a correct fix. We were not calling close.

> I saw that in the close() method of the RSocketStateMachine, we call lots of callback functions. Any of these functions might fail and if might cause early termination of the function. But at the start of the function we call

The current solution for this is that the callbacks are defined with noexcept keyword which means if there is an exception bubbling back to the caller, it will std::terminate the whole program. Maybe not ideal solution, but it eliminates the problem of early termination..

PR #458, opened 2017-05-18
TcpDuplexConnection supports multiple subscribes

lehecka:


lehecka:
was reviewed via https://github.com/rsocket/rsocket-cpp/pull/334

PR #463, opened 2017-05-18
[yarpl] Add skip operator for Flowable/Observable

jacobkahn:
* Implements [`skip`](http://reactivex.io/documentation/operators/skip.html) for both `Flowable` and `Observable`, which moves over arbitrary data in a stream

Usage example:
```
Flowables::range(0, 10)
    ->skip(8)
    ->map([] (int64_t n) { return n + 1; })
    ->subscribe(mySubscriber);
```
Returns `{9, 10}`

### DIFF
--- a/yarpl/include/yarpl/flowable/Flowable.h
+++ b/yarpl/include/yarpl/flowable/Flowable.h
@@ -95,6 +95,8 @@ class Flowable : public virtual Refcounted {
 
   auto take(int64_t);
 
+  auto skip(int64_t);
+
   auto subscribeOn(Scheduler&);
 
   /**
@@ -371,6 +373,12 @@ template <typename T>
 auto Flowable<T>::take(int64_t limit) {
   return Reference<Flowable<T>>(
       new TakeOperator<T>(Reference<Flowable<T>>(this), limit));
+}
+
+template <typename T>
+auto Flowable<T>::skip(int64_t offset) {
+  return Reference<Flowable<T>>(
+    new SkipOperator<T>(Reference<Flowable<T>>(this), offset));
 }
 
 template <typename T>
--- a/yarpl/include/yarpl/flowable/FlowableOperator.h
+++ b/yarpl/include/yarpl/flowable/FlowableOperator.h
@@ -319,8 +319,8 @@ class SkipOperator : public FlowableOperator<T, T> {
 
   void subscribe(Reference<Subscriber<T>> subscriber) override {
     FlowableOperator<T, T>::upstream_->subscribe(
-        Reference<Subscription>(new Subscription(
-            Reference<Flowable<T>>(this), offset_, std::move(subscriber))));
+      make_ref<Subscription>(
+          Reference<Flowable<T>>(this), offset_, std::move(subscriber)));
   }
 
  private:
@@ -331,33 +331,29 @@ class SkipOperator : public FlowableOperator<T, T> {
         Reference<Flowable<T>> flowable,
         int64_t offset,
         Reference<Subscriber<T>> subscriber)
-        : FlowableOperator<T, T>::Subscription(
-             std::move(flowable),
-             std::move(subscriber)),
+        : Super(std::move(flowable), std::move(subscriber)),
          offset_(offset) {}
 
-    void onNext(T value) override {
-      if (offset_ <= 0) {
-        if (pending_ > 0) {
-          --pending_;
-          Super::subscriberOnNext(
-              std::move(value));
-        }
-      } else {
-        --offset_;
-      }
-    }
+     void onNext(T value) override {
+       if (offset_ > 0) {
+         --offset_;
+       } else {
+         Super::subscriberOnNext(
+             std::move(value));
+       }
+     }
 
-    void request(int64_t delta) override {
-      if (delta > 0) {
-        pending_ += delta;
-        Super::request(delta);
-      }
-    }
+     void request(int64_t delta) override {
+       if (firstRequest_) {
+         firstRequest_ = false;
+         delta = delta + offset_;
+       }
+       Super::request(delta);
+     }
 
    private:
-    int64_t pending_{0};
     int64_t offset_;
+    bool firstRequest_{true};
   };
 
   const int64_t offset_;
--- a/yarpl/include/yarpl/flowable/FlowableOperator.h
+++ b/yarpl/include/yarpl/flowable/FlowableOperator.h
@@ -312,6 +312,58 @@ class TakeOperator : public FlowableOperator<T, T> {
 };
 
 template <typename T>
+class SkipOperator : public FlowableOperator<T, T> {
+ public:
+  SkipOperator(Reference<Flowable<T>> upstream, int64_t offset)
+      : FlowableOperator<T, T>(std::move(upstream)), offset_(offset) {}
+
+  void subscribe(Reference<Subscriber<T>> subscriber) override {
+    FlowableOperator<T, T>::upstream_->subscribe(
+        Reference<Subscription>(new Subscription(
+            Reference<Flowable<T>>(this), offset_, std::move(subscriber))));
+  }
+
+ private:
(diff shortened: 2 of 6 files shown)


yarpl/include/yarpl/flowable/Flowable.h
```diff
@@ -337,6 +339,12 @@ auto Flowable<T>::take(int64_t limit) {
 }
 
 template <typename T>
+auto Flowable<T>::skip(int64_t offset) {
+  return Reference<Flowable<T>>(
```
alexmalyshev: This return could be -

  `return make_ref<SkipOperator<T>>(Reference<Flowable<T>>(this), offset);`

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -246,6 +246,58 @@ class TakeOperator : public FlowableOperator<T, T> {
...
 
 template <typename T>
+class SkipOperator : public FlowableOperator<T, T> {
+ public:
+  SkipOperator(Reference<Flowable<T>> upstream, int64_t offset)
+      : FlowableOperator<T, T>(std::move(upstream)), offset_(offset) {}
+
+  void subscribe(Reference<Subscriber<T>> subscriber) override {
+    FlowableOperator<T, T>::upstream_->subscribe(
+        Reference<Subscription>(new Subscription(
```
alexmalyshev: `make_ref<Subscription>(Reference<Flowable<T>>(this), offset_, std::move(subscriber))`

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -246,6 +246,58 @@ class TakeOperator : public FlowableOperator<T, T> {
...
+
+ private:
+  class Subscription : public FlowableOperator<T, T>::Subscription {
+    using Super = typename FlowableOperator<T,T>::Subscription;
+   public:
+    Subscription(
+        Reference<Flowable<T>> flowable,
+        int64_t offset,
+        Reference<Subscriber<T>> subscriber)
+        : FlowableOperator<T, T>::Subscription(
```
alexmalyshev: Can replace with `: Super(std::move(flowable), std::move(subscriber))`

yarpl/include/yarpl/observable/Observable.h
```diff
@@ -146,6 +148,12 @@ auto Observable<T>::take(int64_t limit) {
 }
 
 template <typename T>
+auto Observable<T>::skip(int64_t offset) {
+  return Reference<Observable<T>>(
```
alexmalyshev: `make_ref` ditto

yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -226,6 +226,47 @@ class TakeOperator : public ObservableOperator<T, T> {
...
 
 template <typename T>
+class SkipOperator : public ObservableOperator<T, T> {
+ public:
+  SkipOperator(Reference<Observable<T>> upstream, int64_t offset)
+      : ObservableOperator<T, T>(std::move(upstream)), offset_(offset) {}
+
+  void subscribe(Reference<Observer<T>> observer) override {
+    ObservableOperator<T, T>::upstream_->subscribe(
+        Reference<Subscription>(new Subscription(
```
alexmalyshev: `make_ref`

yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -226,6 +226,47 @@ class TakeOperator : public ObservableOperator<T, T> {
...
+
+ private:
+  class Subscription : public ObservableOperator<T, T>::Subscription {
+    using Super = typename ObservableOperator<T,T>::Subscription;
+   public:
+    Subscription(
+       Reference<Observable<T>> observable,
+       int64_t offset,
+       Reference<Observer<T>> observer)
+       : ObservableOperator<T, T>::Subscription(
```
alexmalyshev: `Super` ditto

yarpl/test/FlowableTest.cpp
```diff
@@ -180,6 +180,18 @@ TEST(FlowableTest, SimpleTake) {
   EXPECT_EQ(std::size_t{0}, Refcounted::objects());
 }
 
+TEST(FlowableTest, SimpleSkip) {
+  ASSERT_EQ(std::size_t{0}, Refcounted::objects());
+  EXPECT_EQ(run(Flowables::range(0, 10)->skip(8)), std::vector<int64_t>({8, 9}));
```
alexmalyshev: Can you also add a test for something like `range(0, 10)->skip(5)` but where the CollectingSubscriber only does `request(2)` and then `cancel()` when it gets the 2.  I believe `run()` does `request(100)` by default, which kind of makes the whole thing behave like an Observable for tests with < 100 items.

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -246,6 +246,58 @@ class TakeOperator : public FlowableOperator<T, T> {
...
+             std::move(subscriber)),
+         offset_(offset) {}
+
+    void onNext(T value) override {
+      if (offset_ <= 0) {
+        if (pending_ > 0) {
+          --pending_;
+          Super::subscriberOnNext(
+              std::move(value));
+        }
```
phoad: Should we request further from the Super, up till we fulfill all the pending requests?

yarpl/include/yarpl/flowable/Flowable.h
jacobkahn: The way it's written now is the same for each other operator - changing it would break that convention. Do it anyways?

yarpl/include/yarpl/flowable/FlowableOperator.h
jacobkahn: Same as above — this syntax is used for all other operators. Can change now or wait for a codemod.

lehecka:
can you please rebase on top of master and update the PR? I will work with you to get this in.

yarpl/include/yarpl/flowable/Flowable.h
jacobkahn: Type mismatch with inferred type means this won't work (discussed offline).

yarpl/include/yarpl/observable/Observable.h
jacobkahn: Type mismatch with inferred type means this won't work (discussed offline).

yarpl/test/FlowableTest.cpp
```diff
@@ -286,6 +286,34 @@ TEST(FlowableTest, SimpleTake) {
...
+  EXPECT_EQ(std::size_t{0}, Refcounted::objects());
+  EXPECT_EQ(run(Flowables::range(0, 10)->skip(12)), std::vector<int64_t>({}));
+  EXPECT_EQ(std::size_t{0}, Refcounted::objects());
+}
+
+TEST(FlowableTest, SkipPartial) {
+  ASSERT_EQ(std::size_t{0}, Refcounted::objects());
+  auto collector = make_ref<CollectingSubscriber<int64_t>>(2);
+  auto flowable = Flowables::range(0, 10)
+    ->skip(5);
```
somasun: run clang-format ?

yarpl/include/yarpl/flowable/Flowable.h
```diff
@@ -95,6 +95,8 @@ class Flowable : public virtual Refcounted {
 
   auto take(int64_t);
 
+  auto skip(int64_t);
```
somasun: why aren't these uint ?

jacobkahn: just copied the signature for take. `int64_t` is being used in a few other places, and we're using these numbers with `request(int64_t)`, and I wanted to try to eliminate implicit casts

PR #470, opened 2017-05-21
cleaning up server connection creation

lehecka:
Tested connection creation on examples.

The current known issues are around terminating and closing RSocketServer. Also some of the examples crash. I will fix them in the next PR.


benjchristensen:
Reviewing now.

lexs:
Could we bring back the deleted test of this class? It tested very real problems we experienced in prod.

benjchristensen:
Which test are you referring to?

PR #473, opened 2017-05-22
RequestResponse Cancellation Test

benjchristensen:
Assert that a requestResponse can be cancelled.


benjchristensen:
Apparently running the unit tests locally didn't check all the build :-) Will fix and repost.

PR #476, opened 2017-05-23
Fix reading an undefined boolean member

alexmalyshev:
SingleSubscription::cancelled_ could be read before ever being set.  UBSAN
catches this.


benjchristensen:
Nice. That confirms something I was trying to fix as well. Thank you!

PR #480, opened 2017-05-23
Experiment with test subscriber (IGNORE)

benjchristensen:
An experiment that I need the Travis ASAN build for.


benjchristensen:
That experiment worked. Will push that change by itself.

PR #483, opened 2017-05-23
RequestChannelTest

benjchristensen:
Simple unit test of requestChannel with finite successful request and response streams.


benjchristensen:
https://github.com/rsocket/rsocket-cpp/pull/481 fixed the issues with these tests so it looks like we can merge them now.

PR #486, opened 2017-05-25
fixing thread safety in RSocketServer and clean shut down

lehecka:
Tested this by setting number of tcp server worker threads to 8 and modified one of the client examples to create thousands of connection in a infinite loop. Then I terminated the server and verified that the cleanup happened cleanly and all connections got closed.

### DIFF
--- a/src/RSocketServer.cpp
+++ b/src/RSocketServer.cpp
@@ -14,7 +14,11 @@ namespace rsocket {
 RSocketServer::RSocketServer(
     std::unique_ptr<ConnectionAcceptor> connectionAcceptor)
     : duplexConnectionAcceptor_(std::move(connectionAcceptor)),
-      setupResumeAcceptors_([]{return new rsocket::SetupResumeAcceptor(ProtocolVersion::Unknown); }) {}
+      setupResumeAcceptors_([]{
+        return new rsocket::SetupResumeAcceptor(
+            ProtocolVersion::Unknown,
+            folly::EventBaseManager::get()->getExistingEventBase());
+      }) {}
 
 RSocketServer::~RSocketServer() {
   // Will stop forwarding connections from duplexConnectionAcceptor_ to
@@ -71,14 +75,12 @@ void RSocketServer::start(OnSetupConnection onSetupConnection) {
       ->start([ this, onSetupConnection = std::move(onSetupConnection) ](
           std::unique_ptr<DuplexConnection> connection,
           folly::EventBase & eventBase) {
-        auto* acceptor = setupResumeAcceptors_.get();
-
         if (isShutdown_) {
           // connection is getting out of scope and terminated
-          //
-          // if we created a new acceptor its ok, we aren't queuing anything new
           return;
         }
+
+        auto* acceptor = setupResumeAcceptors_.get();
 
         VLOG(2) << "Going to accept duplex connection";
 
@@ -191,7 +193,7 @@ void RSocketServer::addConnection(
 }
 
 void RSocketServer::removeConnection(
-    const std::shared_ptr<rsocket::RSocketStateMachine>& socket) {
+    const std::shared_ptr<RSocketStateMachine>& socket) {
   auto locked = sockets_.lock();
   locked->erase(socket);
 
--- a/src/RSocketServer.cpp
+++ b/src/RSocketServer.cpp
@@ -16,14 +16,23 @@ namespace rsocket {
 RSocketServer::RSocketServer(
     std::unique_ptr<ConnectionAcceptor> connectionAcceptor)
     : duplexConnectionAcceptor_(std::move(connectionAcceptor)),
-      setupResumeAcceptor_(ProtocolVersion::Unknown) {}
+      setupResumeAcceptors_([]{return new rsocket::SetupResumeAcceptor(ProtocolVersion::Unknown); }) {}
 
 RSocketServer::~RSocketServer() {
+  // Will stop forwarding connections from duplexConnectionAcceptor_ to
+  // setupResumeAcceptors_
+  isShutdown_ = true;
+
   // Stop accepting new connections.
   duplexConnectionAcceptor_->stop();
 
-  // FIXME(alexanderm): This is where we /should/ close the FrameTransports
-  // sitting around in the SetupResumeAcceptor, but we can't yet...
+  std::vector<folly::Future<folly::Unit>> closingFutures;
+  for (auto& acceptor : setupResumeAcceptors_.accessAllThreads()) {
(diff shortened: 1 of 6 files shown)


src/internal/SetupResumeAcceptor.cpp
```diff
@@ -59,19 +60,26 @@ SetupResumeAcceptor::SetupResumeAcceptor(
     defaultFrameSerializer_ =
         FrameSerializer::createFrameSerializer(protocolVersion);
   }
+  eventBase_ = folly::EventBaseManager::get()->getExistingEventBase();
```
lexs: Please pass this in instead, otherwise it's impossible to test this class.

src/internal/SetupResumeAcceptor.cpp
```diff
@@ -59,19 +60,26 @@ SetupResumeAcceptor::SetupResumeAcceptor(
...
-  }
+  close().get();
 }
 
 void SetupResumeAcceptor::processFrame(
     std::shared_ptr<FrameTransport> transport,
     std::unique_ptr<folly::IOBuf> frame,
     SetupResumeAcceptor::OnSetup onSetup,
     SetupResumeAcceptor::OnResume onResume) {
+  DCHECK(eventBase_ == folly::EventBaseManager::get()->getExistingEventBase());
```
lexs: `DCHECK(eventBase_.isInEventBaseThread())`

src/internal/SetupResumeAcceptor.cpp
```diff
@@ -59,19 +60,26 @@ SetupResumeAcceptor::SetupResumeAcceptor(
...
+
+  if(!connections_) {
```
lexs: nit: spacing

src/internal/SetupResumeAcceptor.cpp
```diff
@@ -198,13 +206,42 @@ void SetupResumeAcceptor::closeAndRemoveConnection(
...
+  connections_->erase(transport);
+}
+
+folly::Future<folly::Unit> SetupResumeAcceptor::close() {
+  folly::Promise<folly::Unit> closingPromise;
+  auto closingFuture = closingPromise.getFuture();
+
+  if(eventBase_->isInEventBaseThread()) {
+    closeAllConnections();
+    closingPromise.setValue();
```
lexs: you could just do `return makeFuture<folly::Unit>()` here and move the Promise into the `else`

src/internal/SetupResumeAcceptor.cpp
```diff
@@ -198,13 +206,42 @@ void SetupResumeAcceptor::closeAndRemoveConnection(
...
+  } else {
+    auto queued = eventBase_->runInEventBaseThread([this, closingPromise = std::move(closingPromise)]() mutable {
```
lexs: if you do folly::via() here you won't have to create a promise at all

src/internal/SetupResumeAcceptor.h
```diff
@@ -52,11 +55,16 @@ class SetupResumeAcceptor final {
       folly::exception_wrapper ex);
   void removeConnection(const std::shared_ptr<FrameTransport>& transport);
 
+  void closeAllConnections();
+
   std::shared_ptr<FrameSerializer> getOrAutodetectFrameSerializer(
       const folly::IOBuf& firstFrame);
 
-  std::unordered_set<std::shared_ptr<FrameTransport>> connections_;
+  std::shared_ptr<std::unordered_set<std::shared_ptr<FrameTransport>>> connections_{
```
lexs: this doesn't need to be a `shared_ptr` and I think a `bool destroyed_` is clearer than nulling this out anyway.

src/internal/SetupResumeAcceptor.cpp
```diff
@@ -59,19 +60,26 @@ SetupResumeAcceptor::SetupResumeAcceptor(
...
   }
+  eventBase_ = folly::EventBaseManager::get()->getExistingEventBase();
+  CHECK(eventBase_);
 }
 
 SetupResumeAcceptor::~SetupResumeAcceptor() {
-  for (auto& connection : connections_) {
-    connection->close(std::runtime_error("SetupResumeAcceptor closed"));
-  }
+  close().get();
```
lexs: This will still be called even if we explicitly call `close()` before and cause a thread hop, probably fine I guess.

src/internal/SetupResumeAcceptor.cpp
```diff
@@ -198,13 +206,42 @@ void SetupResumeAcceptor::closeAndRemoveConnection(
...
+}
+
+void SetupResumeAcceptor::closeAllConnections() {
+  if(!connections_) {
+    return;
+  }
+
+  folly::exception_wrapper closingEx = std::runtime_error("shutting down");
+  for(auto& connection : *connections_) {
+    connection->close(closingEx);
```
alexmalyshev: I think it's better to just do `connection->close(std::runtime_error("shutting down"))`.  Otherwise you end up copying a lot of `folly::exception_wrapper` objects which I think is marginally more expensive than constructing them and passing them as rvalues.

src/RSocketServer.cpp
```diff
@@ -174,7 +191,7 @@ void RSocketServer::addConnection(
 }
 
 void RSocketServer::removeConnection(
-    std::shared_ptr<RSocketStateMachine> socket) {
+    const std::shared_ptr<rsocket::RSocketStateMachine>& socket) {
```
alexmalyshev: Nit: `rsocket::` not necessary here.

src/RSocketServer.cpp
```diff
@@ -64,10 +71,18 @@ void RSocketServer::start(OnSetupConnection onSetupConnection) {
       ->start([ this, onSetupConnection = std::move(onSetupConnection) ](
           std::unique_ptr<DuplexConnection> connection,
           folly::EventBase & eventBase) {
+        auto* acceptor = setupResumeAcceptors_.get();
+
+        if (isShutdown_) {
+          // connection is getting out of scope and terminated
+          //
+          // if we created a new acceptor its ok, we aren't queuing anything new
```
alexmalyshev: Why not do the `if (isShutdown_) { ... }` check before `auto acceptor = setupResumeAcceptors_.get();` then?

lehecka: As I am thinking about the isShutdown_ check here, I am realizing that you are right. There will be no race or any problem with having this check earlier.

src/internal/SetupResumeAcceptor.cpp
lehecka: its a thread hop on shut down which I think is acceptable

PR #487, opened 2017-05-25
Fix integer overflow in flowable::SkipOperator

alexmalyshev:
Changes CollectingSubscriber to default to requesting INT_MAX.  This exposes the
overflow on master, and is a more intuitive default than 100.

### DIFF
--- a/yarpl/include/yarpl/flowable/FlowableOperator.h
+++ b/yarpl/include/yarpl/flowable/FlowableOperator.h
@@ -346,7 +346,13 @@ class SkipOperator : public FlowableOperator<T, T> {
      void request(int64_t delta) override {
        if (firstRequest_) {
          firstRequest_ = false;
-         delta = delta + offset_;
+
+         // TODO: This should be helper like `clampAdd` or something.
+         if (delta <= std::numeric_limits<int64_t>::max() - offset_) {
+           delta = delta + offset_;
+         } else {
+           delta = std::numeric_limits<int64_t>::max();
+         }
        }
        Super::request(delta);
      }
--- a/yarpl/test/FlowableTest.cpp
+++ b/yarpl/test/FlowableTest.cpp
@@ -16,7 +16,8 @@ void unreachable() {
 template <typename T>
 class CollectingSubscriber : public Subscriber<T> {
  public:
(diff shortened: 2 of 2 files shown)


yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -346,7 +346,13 @@ class SkipOperator : public FlowableOperator<T, T> {
      void request(int64_t delta) override {
        if (firstRequest_) {
          firstRequest_ = false;
-         delta = delta + offset_;
+
+         // TODO: This should be helper like `clampAdd` or something.
+         if (delta <= std::numeric_limits<int64_t>::max() - offset_) {
```
benjchristensen: Consider using `SubscriptionHelper` https://github.com/rsocket/rsocket-cpp/blob/master/yarpl/src/yarpl/flowable/utils/SubscriptionHelper.h#L39

It handles the limits, overflow, etc in one standardized and tested place.

yarpl/test/FlowableTest.cpp
```diff
@@ -16,7 +16,8 @@ void unreachable() {
 template <typename T>
 class CollectingSubscriber : public Subscriber<T> {
  public:
-  explicit CollectingSubscriber(int64_t requestCount = 100)
```
benjchristensen: Consider using `TestSubscriber` which is the standardized implementation for doing this. We shouldn't need this `CollectingSubscriber` impl.

alexmalyshev: We'd have to at least add `std::vector<T>& TestSubscriber::values();` and a ctor that controls the initial request count, but yeah I can see they're more or less the same.

yarpl/include/yarpl/flowable/FlowableOperator.h
alexmalyshev: Hmm it does, but only for `std::atomic<int64_t>`, and I have an `int64_t` here.  I could add another overload however.

alexmalyshev:
I'll close this off in favor of https://github.com/rsocket/rsocket-cpp/pull/492.  I'll try to do `s|CollectingSubscriber|TestSubscriber|g` in a separate PR.

PR #492, opened 2017-05-26
Move static SubscriptionHelper class to credits:: namespace

alexmalyshev:
Codify all NO_FLOW_CONTROL and CANCELED constants into one header/module.

Also add a version of credits::add() that works with non-atomic integers.  Use
it to handle the overflow case in Flowable::SynchronousSubscription and in
SkipOperator::Subscription.


alexmalyshev:
Uhh whoops, I realized I have `credits.cpp` in `yarpl/include`.  I'll fix that...

PR #497, opened 2017-06-02
Delete Refcounted::objects()

alexmalyshev:
We already have LSAN as a means for detecting memory leaks automatically when
running tests.  Keeping around the conditional compilation around NDEBUG isn't
worth it. Many optimized build environments define NDEBUG and this makes them
incapable of compiling the test code.


phoad:
Did you validate that when we remove this line
if we have a leak that this line was catching, it is also caught by just running the unit tests?

alexmalyshev:
Yes, running with LSAN will catch any leaks that checking `Refcounted::objects() == 0` at the end of a test will find.  I didn't validate this here on this diff directly, but I've seen it on previous diffs plenty of times.

yarpl/include/yarpl/Refcounted.h
```diff
@@ -14,32 +14,24 @@ namespace yarpl {
...
+/// NOTE: Only derive using "virtual public" inheritance.
 class Refcounted {
  public:
-#if !defined(NDEBUG)
-  Refcounted();
-  virtual ~Refcounted();
-
-  // Return the number of live refcounted objects.  For testing.
-  static std::size_t objects();
+  virtual ~Refcounted() = default;
```
phoad: Should we have a warning, logging (at least in the debug build), if we delete a Refcounted object whose reference count is greater than 1?

lehecka:
I don't feel super strongly about this change. I liked the fact that we could have explicit assumptions expressed in the test code. Also knowing that not everyone is running LSAN with their diffs. I would accept the PR if we make LSAN part of Travis.

alexmalyshev:
LSAN is part of Travis.  All ASAN runs done on Linux include LSAN.

PR #498, opened 2017-06-05
Systematic usage of log levels in rsocket-cpp

phoad:
Sending an early PR to get feedback.
Internal: T18976114

LOG(ERROR) 
- A error from which the process can recover itself, but affects connections or data frames.
- Examples:  Unparseable Frames; Unsupported versions; Handler not provided by application; SETUP asks for resumption, but server does not provide support.

LOG(INFO)
- Data which might interest the user of rsocket.
- Examples: Server/Client startup/shutdown.  OnError events.  Maybe Resumption events?

VLOG(N)
- These log levels should be for the assistance of the developer.

Example:
VLOG(1) - Details at connection/stream level - connection setup/teardown events; stream begin/onComplete events.
VLOG(2) - Details at Frame level - in/out frames - t18912650
VLOG(3) - Details at byte level - in/out hexDump
VLOG(4) - State machines changes ?
VLOG(5) - Important object construction/destruction.
and so on ...

We could rate limit with LOG_FIRST_N or LOG_EVERY_N for repetitive events.

### DIFF
--- a/src/Payload.cpp
+++ b/src/Payload.cpp
@@ -10,13 +10,16 @@ namespace rsocket {
 Payload::Payload(
     std::unique_ptr<folly::IOBuf> _data,
     std::unique_ptr<folly::IOBuf> _metadata)
-    : data(std::move(_data)), metadata(std::move(_metadata)) {}
+    : data(std::move(_data)), metadata(std::move(_metadata)) {
+  VLOG(5) << "Created Payload.";
+}
 
 Payload::Payload(const std::string& _data, const std::string& _metadata)
     : data(folly::IOBuf::copyBuffer(_data)) {
   if (!_metadata.empty()) {
     metadata = folly::IOBuf::copyBuffer(_metadata);
   }
+  VLOG(5) << "Created PayLoad.";
 }
 
 void Payload::checkFlags(FrameFlags flags) const {
--- a/src/RSocketResponder.cpp
+++ b/src/RSocketResponder.cpp
@@ -7,12 +7,14 @@ namespace rsocket {
 
 yarpl::Reference<yarpl::single::Single<rsocket::Payload>>
 RSocketResponder::handleRequestResponse(rsocket::Payload request, rsocket::StreamId streamId) {
+  VLOG(4) << "handleRequestResponse not implemented.";
   return yarpl::single::Singles::error<rsocket::Payload>(
       std::logic_error("handleRequestResponse not implemented"));
 }
 
 yarpl::Reference<yarpl::flowable::Flowable<rsocket::Payload>>
 RSocketResponder::handleRequestStream(rsocket::Payload request, rsocket::StreamId streamId) {
+  VLOG(4) << "handleRequestStream not implemented.";
   return yarpl::flowable::Flowables::error<rsocket::Payload>(
       std::logic_error("handleRequestStream not implemented"));
 }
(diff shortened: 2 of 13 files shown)


src/statemachine/RSocketStateMachine.cpp
```diff
@@ -175,7 +185,7 @@ void RSocketStateMachine::close(
   isClosed_ = true;
   stats_->socketClosed(signal);
 
-  VLOG(6) << "close";
+  VLOG_EVERY_N(4, 100) << "close";
```
lexs: I don't think this needs to be VLOG_EVERY_N given that it's already VLOG

lexs: Same above

src/statemachine/RSocketStateMachine.cpp
```diff
@@ -354,18 +364,23 @@ void RSocketStateMachine::closeStreams(StreamCompletionSignal signal) {
...
 
 void RSocketStateMachine::resumeStreams() {
+  VLOG(4) << "Resuming streams.";
 //  for (auto& streamKV : streamState_->streams_) {
 //    streamKV.second->resumeStream(*requestHandler_);
 //  }
 }
 
 void RSocketStateMachine::processFrame(std::unique_ptr<folly::IOBuf> frame) {
+  VLOG_EVERY_N(2, 10000) << "processFrame - frame length=" << frame->length() << std::endl
```
lexs: I think same for all of these, if I want this logging I want it for all frames, not just one every 10 seconds. Also 2 seems too low of a level for frame logging?

src/framing/FrameSerializer_v1_0.cpp
```diff
@@ -65,6 +66,7 @@ static void serializeMetadataInto(
 
   // Use signed int because the first bit in metadata length is reserved.
   if (metadata->length() > kMaxMetadataLength) {
+    LOG(ERROR) << "Metadata is too big to serialize.";
     CHECK(false) << "Metadata is too big to serialize";
```
lexs: the CHECK here will already log (even though it should really be `LOG(FATAL)` instead).

src/Payload.cpp
```diff
@@ -10,13 +10,16 @@ namespace rsocket {
...
+    : data(std::move(_data)), metadata(std::move(_metadata)) {
+  VLOG(5) << "Created Payload.";
+}
 
 Payload::Payload(const std::string& _data, const std::string& _metadata)
     : data(folly::IOBuf::copyBuffer(_data)) {
   if (!_metadata.empty()) {
     metadata = folly::IOBuf::copyBuffer(_metadata);
   }
+  VLOG(5) << "Created PayLoad.";
```
lexs: Could these log the contents instead? Otherwise they seem kinda useless.

src/RSocketResponder.cpp
```diff
@@ -7,12 +7,14 @@ namespace rsocket {
 
 yarpl::Reference<yarpl::single::Single<rsocket::Payload>>
 RSocketResponder::handleRequestResponse(rsocket::Payload request, rsocket::StreamId streamId) {
+  VLOG(4) << "handleRequestResponse not implemented.";
```
lehecka: it just makes me think , how much are these strings going to increase the size of the binary even if the log level is skipping the log statements...

phoad: It sure do increase the size of the binary. I will eliminate any that is not required. There is a way to decrease the size increase a bit. If we want we can get rid of all INFO messages (the string values they use) in prod easily, it will also cause getting rid of all VLOG statements' cost:

Strip Logging Messages
 #define GOOGLE_STRIP_LOG 1
 #include <glog/logging.h>
The compiler will remove the log messages whose severities are less than the specified integer value. VLOG's severity level INFO (0), so they will be removed as well as INFO log statements.

PR #499, opened 2017-06-05
adding {Flowable, Observable}::justOnce operator

lehecka:
this is a version of the just operator which works with movable types. The aim is to enable using just operator with the restriction that you can subscribe to the flowable/observable only once.


yarpl/include/yarpl/observable/Observables.h
```diff
@@ -47,6 +47,21 @@ class Observables {
     return Observable<T>::create(std::move(lambda));
   }
 
+  // this will generate an observable which can be subscribed to only once
+  template <typename T>
+  static Reference<Observable<T>> justOnce(T value) {
+    auto lambda = [value = std::move(value), used = false](Reference<Observer<T>> observer) mutable {
+      CHECK(!used) << "justOnce value was already used";
```
lexs: I think this should resolve with an exception in the other case, crashing doesn't seem correct.

yarpl/test/FlowableTest.cpp
```diff
@@ -97,6 +97,28 @@ TEST(FlowableTest, SingleFlowable) {
   EXPECT_EQ(std::size_t{0}, Refcounted::objects());
 }
 
+TEST(FlowableTest, SingleMovableFlowable) {
+  ASSERT_EQ(std::size_t{0}, Refcounted::objects());
+
+  auto value = std::make_unique<int>(123456);
+
+  auto flowable = Flowables::justOnce(std::move(value));
+  EXPECT_EQ(std::size_t{1}, Refcounted::objects());
```
lexs: Isn't this getting removed?

yarpl/include/yarpl/observable/Observables.h
lehecka: I was going back and forth about this. On one hand I agree with you that crashing is not good. On the other hand this is violation of the API and the check is just checking that you are using it correctly. Debugging in the case of CHECK is much easier to debug.
I can also add DCHECK and exception instead of CHECK.
Anybody else want's to weigh in?

alexmalyshev: I think we shouldn't crash here.  If someone tries to resubscribe to a JustOnce operator, then they should probably get sent an onError.

yarpl/test/Observable_test.cpp
```diff
@@ -307,6 +303,22 @@ TEST(Observable, Just) {
       std::vector<const char*>({"ab", "pq", "yz"}));
 }
 
+TEST(Observable, SingleMovable) {
+  auto value = std::make_unique<int>(123456);
+
+  auto observable = Observables::justOnce(std::move(value));
```
phoad: nit: one more case to check if it fires the exception correctly

lehecka: Will cover in the next diff. Thank you for your feedback!

PR #500, opened 2017-06-05
Ability to pretty print in/out frames in rsocket-cpp

phoad:
When we are debugging the program, we would want to see increasing levels of details with the --verbosity (-v) flag.

Say for --verbosity=3, a simple request-response should look like something like this.

Out: SETUP FrameLen=45 DataLen=245 Encoding=abc 
Out: REQUEST_RESPONSE FrameLen=240 DataLen=156 Data=hello Metadata=World
In: PAYLOAD FrameLen=230 Flags=NEXT|COMPLETE DataLen=23 Data=hello Metadata=Buddy

Then for --verbosity=3, we could print the hexDump.

### DIFF
--- a/src/statemachine/RSocketStateMachine.cpp
+++ b/src/statemachine/RSocketStateMachine.cpp
@@ -272,7 +272,7 @@ void RSocketStateMachine::closeWithError(Frame_ERROR&& error) {
       error.payload_.data->cloneAsValue().moveToFbString().toStdString());
 
   if (frameSerializer_) {
-    outputFrameOrEnqueue(frameSerializer_->serializeOut(std::move(error)));
+    outputFrameOrEnqueue(std::move(error));
   }
   close(std::move(exception), signal);
 }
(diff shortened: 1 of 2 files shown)


phoad:
I need to update the default operator<< for the Frame and Payload for a better output.

This is a sample line: 
I0605 16:19:59.468442 88027136 RSocketStateMachine.cpp:638] In:REQUEST_STREAM[0000000000000000, 1](2147483647, [metadata: <null> data: 4]
I0605 16:19:59.514919 88027136 RSocketStateMachine.h:144] Out: PAYLOAD[0000000100100000, 1], ([metadata: 8 data: 13]

src/statemachine/RSocketStateMachine.cpp
```diff
@@ -774,12 +790,15 @@ void RSocketStateMachine::requestFireAndForget(Payload request) {
...
+  VLOG(3) << "Out: " << frame;
   outputFrameOrEnqueue(frameSerializer_->serializeOut(std::move(frame)));
 }
 
 void RSocketStateMachine::metadataPush(std::unique_ptr<folly::IOBuf> metadata) {
+  Frame_METADATA_PUSH metadataPushFrame(std::move(metadata));
+  VLOG(3) << "Out: " << metadataPushFrame;
   outputFrameOrEnqueue(
-      frameSerializer_->serializeOut(Frame_METADATA_PUSH(std::move(metadata))));
+      frameSerializer_->serializeOut(std::move(metadataPushFrame)));
```
phoad: nit: remove "frameSerializer_->serializeOut" in these places too.

phoad: We need to keep it only if we call outputFrameOrEnqueue on frameTransport_ object, not on this object.
   frameTransport_->outputFrameOrEnqueue(frameSerializer_->serializeOut

PR #502, opened 2017-06-06
Passing setup parameters to RSocketClient

lehecka:
this allows client application code to pass setup parameters to the RSocketClient as well as other parameters. All of them have a default value so the client doesn't have to specify them.


lehecka:
merging to unblock myself in another PR. Still interested in feedback.

PR #504, opened 2017-06-06
Improve the pretty print for operator<<Frame, operator<<Payload

phoad:
Improve the pretty print code to match the Flag name with the AllowedFlags data member in the Frame classes.

Sample output:
I0606 16:14:25.145932 237998080 RSocketStateMachine.h:144] Out: PAYLOAD[METADANEXT, 1], ([Metadata(8): 'metadata', Data(13): 'Hello Jane 1!']
I0606 16:14:25.146798 237998080 RSocketStateMachine.h:144] Out: PAYLOAD[METADANEXT, 1], ([Metadata(8): 'metadata', Data(14): 'Hello Jane 10!']
I0606 16:14:25.146890 237998080 RSocketStateMachine.h:144] Out: PAYLOAD[LEASE|COMPLE, 1], ([Metadata(0): <nullptr>, Data(0): <nullptr>]
I0606 16:14:30.147658 237998080 RSocketStateMachine.cpp:445] In: KEEPALIVE[RESUEN|KALIVR|FOLLOW, 0](<0>)
I0606 16:14:30.147709 237998080 RSocketStateMachine.cpp:699] Out: KEEPALIVE[EMPTY, 0](<0>)


src/framing/Frame.cpp
```diff
@@ -94,78 +94,135 @@ std::ostream& operator<<(std::ostream& os, ErrorCode errorCode) {
...
+      && !!(frameFlags & FrameFlags::METADATA)) {
+    ss << delimeter << "METADA";
     delimeter = "|";
+    foundFlags |= FrameFlags::METADATA;
   }
-  if (!!(frameFlags & FrameFlags::FOLLOWS)) {
-    ss << delimeter << "FOLLOWS";
+  if (!!(allowedFlags & FrameFlags::RESUME_ENABLE)
+      && !!(frameFlags & FrameFlags::RESUME_ENABLE)) {
+    ss << delimeter << "RESUEN";
```
ragansa: might be a typo RESUEN ->RESUME?

src/framing/Frame.cpp
```diff
@@ -94,78 +94,135 @@ std::ostream& operator<<(std::ostream& os, ErrorCode errorCode) {
...
+  }
+  if (!!(allowedFlags & FrameFlags::FOLLOWS)
+      && !!(frameFlags & FrameFlags::FOLLOWS)) { // FOLLOWS = 0x80,
+    ss << delimeter << "FOLLOW";
+    delimeter = "|";
+    foundFlags |= FrameFlags::FOLLOWS;
+  }
+  if (!!(allowedFlags & FrameFlags::COMPLETE)
+      && !!(frameFlags & FrameFlags::COMPLETE)) { // COMPLETE = 0x40
+    ss << delimeter << "COMPLE";
```
ragansa: COMPLE -> COMPLETE?

phoad: I was trying to use short version of these enum values. Let me use the whole word version of them. Seems like it will be problem in the future if I use the short versions.

src/framing/Frame.cpp
```diff
@@ -94,78 +94,135 @@ std::ostream& operator<<(std::ostream& os, ErrorCode errorCode) {
...
+    const std::pair<FrameFlags, FrameFlags>& frameFlagsAndAllowedFlags) {
+  FrameFlags frameFlags = frameFlagsAndAllowedFlags.first;
+  FrameFlags allowedFlags = frameFlagsAndAllowedFlags.second;
+  FrameFlags foundFlags = FrameFlags::EMPTY;
 
   std::stringstream ss;
   std::string delimeter = "";
-  if (!!(frameFlags & FrameFlags::NEXT)) {
-    ss << "NEXT";
+  if (!!(allowedFlags & FrameFlags::IGNORE)
```
ragansa: Would it be possible to loop over these flags instead of all the if statements?
You can use the to_string function to print out the flags.

phoad:
I am making updates now related to Scott's comments.

phoad:
Previous commit (in this CR) was printing out multiple values, instead of a single one for FrameFlags' value of 80 / 40.. (Thanks for catching it Scott) 
Now, it writes the correct Flag value for the given <Frame, FrameFlags>'s value.

enum clas.. {
  RESUME_ENABLE = 0x80,
  KEEPALIVE_RESPOND = 0x80,
  FOLLOWS = 0x80,
}

**Before:**
I0606 16:14:30.147658 237998080 RSocketStateMachine.cpp:445] In: KEEPALIVE[RESUME_ENABLE|KEEPALIVE_RESPOND|FOLLOW, 0](<0>)
**After:**
I0607 17:48:27.079867 247017472 RSocketStateMachine.cpp:699] Out: KEEPALIVE[KEEPALIVE_RESPOND, 0](<0>)

As seen KEEPALIVE_RESPOND is the only Flag for Frame_KEEPALIVE, but as it's value was matching with the RESUME_ENABLE, KEEPALIVE_RESPOND and FOLLOW, previous version of this PR was writing all of them. This update fixed it.

Out: PAYLOAD[NEXT|METADATA, 1], ([Metadata(8): 'metadata', Data(13): 'Hello Jane 1!']
Out: PAYLOAD[NEXT|METADATA, 1], ([Metadata(8): 'metadata', Data(14): 'Hello Jane 10!']
Out: PAYLOAD[COMPLETE, 1], ([Metadata(0): <nullptr>, Data(0): <nullptr>]
In: KEEPALIVE[KEEPALIVE_RESPOND, 0](<0>)
Out: KEEPALIVE[EMPTY, 0](<0>)

phoad:
I am just updating the code further to fix the build errors.

src/framing/Frame.cpp
```diff
@@ -94,78 +94,120 @@ std::ostream& operator<<(std::ostream& os, ErrorCode errorCode) {
...
+            << "(" << frame.requestN_ << ")";
 }
 
 std::ostream& operator<<(
     std::ostream& os,
     const Frame_REQUEST_RESPONSE& frame) {
-  return os << frame.header_ << ", " << frame.payload_;
+  static const std::vector<std::pair<FrameFlags, std::string>> allowedFlags {
+      {FrameFlags::METADATA, sFrameFlags_METADATA},
+      {FrameFlags::FOLLOWS, sFrameFlags_FOLLOWS} };
```
somasun: Any reason why we are not using the AllowedFlags member variable of the Frame struct (instead of redefining it here) ? Then you could write a simple utility function which takes a list of FrameFlags and returns a string (or vector of strings).

```
std::ostream& operator<<(std::ostream& os, const Frame_REQUEST_RESPONSE& frame) {
     return os << frame.header_ << ", " << frame.payload_ << toString(frame.AllowedFlags);
}
```

ragansa: I think the problem is that the allowed flags doesn't tell you which actual flags that frame can have.  Just what values of flags that the frame can have.  (There is overlap in flag values)
i.e. An AllowedFlags values tells you that the frame can have a flag of value 0x80 but doesn't specify if that means RESUME_ENABLE, KEEPALIVE_RESPOND, or FOLLOWS

phoad: Even though the FrameFlags is an enum class, it is used as old C/++ way and OR/AND operations are applied on it to get a bitset, that represents a set of flags. Furthermore, this enum has members which has the same int value. So to be able to differentiate between what's the corresponding Flag enum member for the given int value, we need information from the Frame type. It is not possible to apply switch/case on them.

We write the "flags" when we write the "header". So the information about the flags is existing in the "header" data member. So, this line will not provide the desired result.
```     return os << frame.header_ << ", " << frame.payload_ << toString(frame.AllowedFlags); ```

We have a set of flags that are accepted by each Frame type. We should match these flags with the given flag values in the "header.flags". If the frame accepts RESUME_ENABLE, and if the given flag value is 0x80, we should only match with RESUME_ENABLE, but not with KEEPALIVE_RESPOND or FOLLOWS. For this we need to pass information about the frame along with the header to "operator<<" function.

I have tried to do no change in the declaration of the Frame classes and do all the updates in the cpp code.

File: src/Payload.cpp [hunk 5]
```diff
@@ -31,15 +31,15 @@ std::ostream& operator<<(std::ostream& os, const Payload& payload) {
                           payload.metadata->computeChainDataLength())
                     : "0")
             << (payload.metadata
-                    ?  "): '" + payload.metadata->cloneAsValue().moveToFbString().substr(0, 80).toStdString() + "'"
+                    ?  "): '" + payload.metadata->cloneAsValue().moveToFbString().substr(0, 40).toStdString() + "'"
```
somasun: please clang-format the code.

src/framing/Frame.cpp
```diff
@@ -94,78 +94,120 @@ std::ostream& operator<<(std::ostream& os, ErrorCode errorCode) {
   return os << "ErrorCode(" << static_cast<uint32_t>(errorCode) << ")";
 }
 
-std::ostream& operator<<(std::ostream& os, FrameFlags frameFlags) {
-  // TODO Match the Flag names with the AllowedFlags in the Frame declarations
+constexpr const char sFrameFlags_EMPTY[] = "EMPTY";
```
somasun: nit. EMPTY can be interpreted as a flag.  I would not print any flag.

phoad: Yes, you are right. It is defined as 0x00, better to keep empty.

phoad:
One test is randomly failing, which is not related to this update. Even though @lehecka has approved the PR, it did not get into the master branch. So I close/open to get a successful Travis run, to see if it is being blocked to merged to master because of this failure.

phoad:
@alexmalyshev can you take a look to this PR and merge it?

src/framing/Frame.cpp
```diff
@@ -95,78 +95,120 @@ std::ostream& operator<<(std::ostream& os, ErrorCode errorCode) {
   return os << "ErrorCode(" << static_cast<uint32_t>(errorCode) << ")";
 }
 
-std::ostream& operator<<(std::ostream& os, FrameFlags frameFlags) {
-  // TODO Match the Flag names with the AllowedFlags in the Frame declarations
+constexpr const char sFrameFlags_EMPTY[] = "0x00";
```
alexmalyshev: These should be `const char*`.  Normally `char[]` is reserved for mutable character buffers.

src/framing/Frame.cpp
```diff
@@ -95,78 +95,120 @@ std::ostream& operator<<(std::ostream& os, ErrorCode errorCode) {
...
+constexpr const char sFrameFlags_METADATA[] = "METADATA";
+constexpr const char sFrameFlags_RESUME_ENABLE[] = "RESUME_ENABLE";
+constexpr const char sFrameFlags_LEASE[] = "LEASE";
+constexpr const char sFrameFlags_KEEPALIVE_RESPOND[] = "KEEPALIVE_RESPOND";
+constexpr const char sFrameFlags_FOLLOWS[] = "FOLLOWS";
+constexpr const char sFrameFlags_COMPLETE[] = "COMPLETE";
+constexpr const char sFrameFlags_NEXT[] = "NEXT";
```
alexmalyshev: I wouldn't name these `sFrameFlags_NEXT`.  The convention for string literals is `kStringLiteralName`, and I think we can get away with doing `kNext` instead of duplicating the `sFrameFlags` prefix everywhere.

src/framing/Frame.cpp
```diff
@@ -95,78 +95,120 @@ std::ostream& operator<<(std::ostream& os, ErrorCode errorCode) {
...
 std::ostream& operator<<(std::ostream& os, const FrameHeader& header) {
   return os << header.type_ << "[" << header.flags_ << ", " << header.streamId_
             << "]";
 }
+
+std::ostream& operator<<(
+    std::ostream& os,
+    const std::pair<
+        const FrameHeader&,
+        const std::vector<std::pair<FrameFlags, std::string>>&>&
```
alexmalyshev: This is really rough to understand...  The second element in this pair is a `vector<pair<Flags, std::string>>&` which is also inside a wrapping pair<...,...>&.  All for the purpose of fitting everything into a single argument for `operator<<`.

There's no need to use vectors to track which flags are allowed for which frame header.  You can write a `FrameFlags allowedFlags(FrameType);` function and that can use the type in the header.  Then have a `const char* toString(FrameFlags, FrameType);` that'll decide which string to return (need the type to differentiate between flags like `FOLLOWS` and `KEEPALIVE_RESPOND`).

I think instead of writing
`std::ostream& operator<<(std::ostream&, const std::vector<std::pair<FrameFlags, std::string>>&>&);`
I think we can do everything we want with
`std::ostream& operator<<(std::ostream&, const FrameHeader&);`

CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ find_package(Threads)
 string(TOLOWER "${CMAKE_BUILD_TYPE}" BUILD_TYPE_LOWER)
 
 if(APPLE)
-  set(OPENSSL_ROOT_DIR "/usr/local/opt/openssl")
+  set(OPENSSL_ROOT_DIR "/usr/local/opt/openssl;/usr/local/homebrew/opt/openssl")
```
alexmalyshev: Should this be part of this PR?

src/framing/Frame.cpp
phoad: Good suggestion. I have also noticed today that we have a matching FrameType per Frame class type. Now its easier to do the thing.

CMakeLists.txt
phoad: I needed it to build in my box, though it does not need to be here.

src/framing/Frame.cpp
phoad: I do not need the FrameFlags information of the Frame class types. Though I need a mapping between each flag type to corresponding string value per Frame class type. I prepared a map for this. I am open for suggestion, though it seemed required such a map to me.

src/framing/Frame.cpp
```diff
@@ -95,41 +96,85 @@ std::ostream& operator<<(std::ostream& os, ErrorCode errorCode) {
   return os << "ErrorCode(" << static_cast<uint32_t>(errorCode) << ")";
 }
 
-std::ostream& operator<<(std::ostream& os, FrameFlags frameFlags) {
-  // TODO Match the Flag names with the AllowedFlags in the Frame declarations
+std::ostream& operator<<(std::ostream& os, std::pair<FrameFlags, FrameType> frameFlagsAndType) {
```
alexmalyshev: I still think it's better to make this a function instead of an operator<< overload, i.e. `std::ostream& writeFlags(std::ostream&, FrameType, FrameFlags);`.  Having to make a pair just to pass in two arguments is silly.

phoad: Sure. I thought if we write a function we will not be able to write
os << x << y << z;
Let me update the code accordingly.

yschimke:
@fuat this seems like it might be annoying for terminal output when payloads contain binary.  Have you considered hex formatting or filtering unprintable characters?

phoad:
@yschimke , hexdump seems a better way to go for payloads. Can you provide an example of the terminal output for me to figure out the best way to fix it?

PR #505, opened 2017-06-07
Separating RSocket connection lifetime management into RSocketConnectionManager

lehecka:
tested on examples with vlog level set to 3 to see the extra logging. All works as expected.

Both client and server RSocket instances now share the same lifetime management. The key behavior elements are:
* all RSocket instances are closed when the RSocketClient/RSocketServer objects are destroyed.
* when the RSocket is closed from the other side, it is automatically removed from the collection of active RSocket instances
* user code can explicitly close the RSocket via RSocketRequester::close


lehecka:
merging to unblock the following PR

PR #507, opened 2017-06-07
RSocketClient from TCP socket

lehecka:
This allows creating RSocketClient directly from the opened TCP socket (folly::AsyncSocket) via 2 synchronous calls.


lehecka:
there is a known race condition in the async tests causing travis to fail. I will unblock myself and land the PR and then follow up on the test failure.

PR #509, opened 2017-06-09
[ToBeDeleted] Update DuplexConnection implementations to use yarpl Subscriber/Subscription

phoad:
* Work in progress - I will write unit tests

Other than code cleanup, the major change is in TcpDuplexConnection.h/cpp files. The SubscriptionBase and SubscriberBase classes are deleted and the code that is using these classes are updated to use yarpl::flowable::Subscription/Subscriber.

Next I will write unit tests to verify these:
 - Call setInput/getOutput multiple times
 - Closing input / output should not effect each other / independent
 - Closing input & output together should not cause the connection to close, only dtor of DuplexConnection should close it
 - Write the tests such that different duplex connections can be swapped and the same tests can be run easily for different implementations.

T19169160 - internal reference, stating the remaining updates to do in dependent places.


src/RSocketSetup.h
```diff
@@ -22,7 +24,7 @@ class RSocketStats;
 class RSocketSetup {
  public:
   RSocketSetup(
-      std::shared_ptr<FrameTransport> frameTransport,
+      yarpl::Reference<FrameTransport> frameTransport,
```
lexs: Is there any benefit of changing this? It just seems risky and makes all users have to use custom types instead of a standard shared_ptr.

src/RSocketClient.cpp
```diff
@@ -43,11 +44,21 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
     networkStats = std::move(networkStats),
     promise = std::move(promise)](
       std::unique_ptr<DuplexConnection> connection,
+      bool isFramedConnection,
```
lehecka: I would suggest, since this is already a large change, to generate a new diff with this particular change.
The idea for the diff would be to make FramedDuplexConnection an internal type which would not be used by the application code or implementors of DuplexConnection interface. Instead they will provide a boolean parameter to RSocketClient/RSocketServer whether the connection is framed or not (so far all implementations we have are not framed). Then internally we would do what the lines 51-59 are doing.

The main motivator for this change is to avoid application code to specify protocol version for FramedDuplexConnection.
cc: @somasun

src/RSocketSetup.h
lehecka: this is implied by the FrameTransport implementing yarpl::Subscriber which is a Reference type.
Please keep in mind the FrameTransport type in on its way out.

yarpl/include/yarpl/Refcounted.h
```diff
@@ -146,12 +143,20 @@ class Reference {
 
  private:
   void inc() {
+    static_assert(
```
lehecka: btw. the move of the code here is to be able to declare yarpl::Reference<T> where T is not fully defined. Its very useful feature for header files usages.

src/RSocketClient.cpp
phoad: Moved the related code to: https://github.com/rsocket/rsocket-cpp/pull/514

phoad:
I have sent another CR that is up to date with the master branch. So I am closing this.

https://github.com/rsocket/rsocket-cpp/pull/550

PR #510, opened 2017-06-09
[rsocket-cpp] Enable tck tests in travis builds

somasun:
- Enable tck tests to run in travis builds
- Make the server terminate with signal (instead of stdin - which doesnt work well with background jobs)
- Reuse the same requester for all requests

### DIFF
--- a/tck-test/TestInterpreter.cpp
+++ b/tck-test/TestInterpreter.cpp
@@ -18,8 +18,8 @@ namespace tck {
 
 TestInterpreter::TestInterpreter(
     const Test& test,
-    std::unique_ptr<RSocketRequester> requester)
-    : requester_(std::move(requester)), test_(test) {
+    RSocketRequester* requester)
+    : requester_(requester), test_(test) {
   DCHECK(!test.empty());
 }
 
--- a/tck-test/client.cpp
+++ b/tck-test/client.cpp
@@ -37,6 +37,8 @@ int main(int argc, char* argv[]) {
   folly::SocketAddress address;
   address.setFromHostPort(FLAGS_ip, FLAGS_port);
 
+  LOG(INFO) << "Creating client to connect to " << address.describe();
+
   // create a client which can then make connections below
   auto rsf = RSocket::createClient(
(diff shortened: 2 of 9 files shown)


lehecka:
looks like there are 2 test issues in travis.
I believe there is a race condition in Request*Test.Hello
The second issue is some memory leak in TCK tests.

tck-test/client.cpp
```diff
@@ -37,6 +37,8 @@ int main(int argc, char* argv[]) {
   folly::SocketAddress address;
   address.setFromHostPort(FLAGS_ip, FLAGS_port);
 
+  LOG(INFO) << "Creating client to connect to " << address.describe();
```
alexmalyshev: `operator<<(std::ostream&, const folly::SocketAddress&)` already exists I think, no need for `.describe()`.

tck-test/TestInterpreter.cpp
```diff
@@ -18,8 +18,8 @@ namespace tck {
 
 TestInterpreter::TestInterpreter(
     const Test& test,
-    std::unique_ptr<RSocketRequester> requester)
-    : requester_(std::move(requester)), test_(test) {
+    RSocketRequester* requester)
```
alexmalyshev: I would probably have this as `RSocketRequester&`, but this is fine too.

tck-test/server.cpp
```diff
@@ -119,9 +118,11 @@ class ServerResponder : public RSocketResponder {
   MarbleStore marbles_;
 };
 
-[[noreturn]] static void signal_handler(int signal) {
+std::promise<void> terminate;
+
+static void signal_handler(int signal) {
   LOG(INFO) << "Terminating program after receiving signal " << signal;
-  exit(signal);
+  terminate.set_value();
```
alexmalyshev: What are the chances this is safe to do in a signal handler...  Regardless, I don't have a better solution as `SignalBarrier` isn't open source and it's very likely Linux-only code.

Any particular reason for `std::future` instead of `folly::Future`?

PR #514, opened 2017-06-12
Internally create FramedConnection

phoad:
As @lehecka suggested, I am splitting this PR into two: https://github.com/rsocket/rsocket-cpp/pull/509/files/e1ad97fa94dee680066277f3c78ca1798e591155#diff-f99ef4eeb8d8c84cdd6b217134f898e8 

Make FramedDuplexConnection an internal type which would not be used by the application code or implementors of DuplexConnection interface. Instead they will provide a boolean parameter to RSocketClient/RSocketServer whether the connection is framed or not (so far all implementations we have are not framed). Internally create FramedConnection.

The main motivator for this change is to avoid application code to specify protocol version for FramedDuplexConnection.
cc: Manikandan Somasundaram, Ondrej Lehecka

### DIFF
--- a/src/RSocketClient.cpp
+++ b/src/RSocketClient.cpp
@@ -48,16 +48,9 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
     folly::EventBase& eventBase) mutable {
     VLOG(3) << "onConnect received DuplexConnection";
 
-    std::unique_ptr<DuplexConnection> framedConnection;
-    if (isFramedConnection) {
-      framedConnection = std::move(connection);
-    } else {
-      framedConnection = std::make_unique<FramedDuplexConnection>(
-          std::move(connection), setupParameters.protocolVersion, eventBase);
-    }
-
     auto rsocket = fromConnection(
-        std::move(framedConnection),
+        std::move(connection),
+        isFramedConnection,
         eventBase,
         std::move(setupParameters),
         std::move(responder),
@@ -72,6 +65,7 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
 
 std::unique_ptr<RSocketRequester> RSocketClient::fromConnection(
     std::unique_ptr<DuplexConnection> connection,
+    bool isFramedConnection,
     folly::EventBase& eventBase,
     SetupParameters setupParameters,
     std::shared_ptr<RSocketResponder> responder,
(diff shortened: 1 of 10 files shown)


src/RSocketClient.cpp
```diff
@@ -43,11 +44,22 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
     networkStats = std::move(networkStats),
     promise = std::move(promise)](
       std::unique_ptr<DuplexConnection> connection,
+      bool isFramedConnection,
       folly::EventBase& eventBase) mutable {
     VLOG(3) << "onConnect received DuplexConnection";
 
+    std::unique_ptr<DuplexConnection> framedConnection;
+    if(isFramedConnection) {
```
somasun: clang-format ?

phoad:
I will also put the OnDuplexConnectionAccept related update into this PR. (I see that I missed that part, thanks @somasun )

phoad:
Applying it to OnDuplexConnectionAccept was not part of the original implementation that I have parted this PR. Though it is a good suggestion that I will update the code for it now!

src/RSocketClient.cpp
phoad: @somasun , I have made updates on the PR related to your comments.

src/transports/tcp/TcpConnectionFactory.cpp
```diff
@@ -72,27 +71,11 @@ TcpConnectionFactory::~TcpConnectionFactory() {
...
 
-void TcpConnectionFactory::connect(OnConnect cb) {
+void TcpConnectionFactory::connect(OnDuplexConnectionConnect cb) {
   worker_.getEventBase()->runInEventBaseThread(
       [ this, fn = std::move(cb) ]() mutable {
         new ConnectCallback(address_, std::move(fn));
       });
 }
 
-std::unique_ptr<DuplexConnection>
```
lehecka: please keep this method as it is used by the RT GW.

the body of the method should be just:

return std::make_unique<TcpDuplexConnection>(
      std::move(socket), eventBase, std::move(stats));

src/RSocketClient.cpp
```diff
@@ -41,13 +42,22 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
...
-    promise = std::move(promise)](
-      std::unique_ptr<DuplexConnection> connection,
-      folly::EventBase& eventBase) mutable {
+    promise = std::move(promise)
+  ](std::unique_ptr<DuplexConnection> connection,
+    bool isFramedConnection,
+    folly::EventBase& eventBase) mutable {
     VLOG(3) << "onConnect received DuplexConnection";
 
+    std::unique_ptr<DuplexConnection> framedConnection;
```
lehecka: lets move this block of code including the isFramedConnection parameter to RSocketClient::fromConnection method because it is the counterpart of the TcpConnectionFactory::createDuplexConnectionFromSocket used by the RTGW.

PR #516, opened 2017-06-15
Delete SubscriptionHelper.{cpp,h}

alexmalyshev:
These aren't used anymore as they were replaced by credits.{cpp,h}.


phoad:
Looks like there is no reference to these files in the CMakeLists.txt either.
LGTM.

phoad:
Do you know why it is named Credits.h/cpp?

alexmalyshev:
Because the functionality in those files is all about adding, consuming, and cancelling credits.

PR #518, opened 2017-06-15
Try upgrading clang to 4.0

alexmalyshev:
Been getting strange linker errors from clang-3.9, want to see if clang-4.0
exists and if it is any better.  Switch the toolchain to trusty, to match the
distribution we're using.

Delete unused (Are they used?  I don't know yaml...) yaml anchors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,20 +9,20 @@ language: cpp
 matrix:
 
    include:
-    - env: CLANG_VERSION=3.9 BUILD_TYPE=Debug CPP_VERSION=14 ASAN=On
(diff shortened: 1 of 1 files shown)


alexmalyshev:
Example nonsensical linker error that I've only seen in clang-3.9 so far

https://travis-ci.org/rsocket/rsocket-cpp/jobs/243080844

alexmalyshev:
Well that didn't work.  Can't replicate the missing make_unique<> issues [https://travis-ci.org/rsocket/rsocket-cpp/jobs/243092942] locally on my end, and it seems like it should transitively include <memory> and all that.

vitaut:
#526 fixes the clang build.

PR #519, opened 2017-06-15
Fixing build on Arch

alexmalyshev:
Again missing gflags includes


phoad:
LGTM. I wonder if "clang-format" tries to delete these headers automatically. @somasun

alexmalyshev:
No clang-format isn't related here.  These files weren't being built until recently, they have had the includes in the first place.

PR #520, opened 2017-06-15
Try upgrading llvm-toolchain to trusty

alexmalyshev:
Maybe clang issues can be fixed by matching llvm-toolchain package distro with
travis distro.


alexmalyshev:
Nope

PR #521, opened 2017-06-15
TCP server should be able to support any version

somasun:
Tested the following combinations:
```
./tckserver  and  ./tckclient   -rs_use_protocol_version 0.1
./tckserver and  ./tckclient  -rs_use_protocol_version 1.0
```

Previously, the following used to fail (since the server defaulted to v0.1)
```
./tckserver and  ./tckclient  -rs_use_protocol_version 1.0
```


phoad:
LGTM.

PR #523, opened 2017-06-16
Fix races in closing a TcpReaderWriter

alexmalyshev:
Each TcpReaderWriter is only accessed by a single thread, but it interleaves
callbacks in ways that aren't safe.  When a TcpInputSubscription or
TcpOutputSubscriber tries to close a TcpReaderWriter, it would only close the
socket, and then let the folly::AsyncSocket callbacks handle the rest in
follow-up events.  Depending on what events are already in the EventBase, we can
run into an inconsistent state.

Fix this by having TcpReaderWriter::close{,Err}() which go through the
{subscriber,subscription,socket}.  They are idempotent, so they are safe to call
in follow-up read{EOF,Err}() events.


alexmalyshev:
Hmm, LSAN says this leaks memory.  I'll take a look...

PR #528, opened 2017-06-19
Export listener port in ConnectionAcceptor and RSocketServer

alexmalyshev:
Lets us assign port 0 to servers, and then read the automatically assigned port
and connect a client.


src/RSocketServer.h
```diff
@@ -85,6 +85,12 @@ class RSocketServer {
 
   void shutdownAndWait();
 
+  /**
+   * Gets the port the ConnectionAcceptor is listening on.  Returns folly::none
+   * if this server is not listening on a port.
+   */
+  folly::Optional<uint16_t> listeningPort() const;
```
lehecka: Please let's not add this method to ConnectionAcceptor or RSocketServer. It looks like TCP specific method which would fit better to TcpConnectionAcceptor. 
For HTTP, Thrift and other servers, the method would mean nothing.

test/RSocketTests.h
```diff
@@ -16,21 +15,11 @@ namespace rsocket {
...
-  std::random_device device;
-  std::uniform_int_distribution<uint16_t> dis(9000, 65000);
-
-  auto const n = dis(device);
-  return static_cast<uint16_t>(n);
-}
-
 inline std::unique_ptr<RSocketServer> makeServer(
-    uint16_t port,
     std::shared_ptr<rsocket::RSocketResponder> responder) {
```
lehecka: can we make the signature as follows?
std::unique_ptr<RSocketServer> makeServer(
      std::shared_ptr<rsocket::RSocketResponder> responder,
      int& usedPort);

so that tests can use the usedPort value to makeClient?

src/RSocketServer.h
alexmalyshev: I'm not sure I understand.  Why would HTTP and Thrift not be able to expose the port they're listening on?  This has nothing to do with TCP port 0, this is just "what port is my server listening on".

test/RSocketTests.h
alexmalyshev: I wouldn't use an out-param, I would do something like

  `pair<unique_ptr<RSocketServer>, uint16_t> makeServer(shared_ptr<RSocketResponder>);`

but I prefer the usage of `listeningPort()` far more.

src/transports/tcp/TcpConnectionAcceptor.cpp
```diff
@@ -114,4 +114,12 @@ void TcpConnectionAcceptor::stop() {
       [this] { serverSocket_.reset(); });
   serverThread_.reset();
 }
+
+folly::Optional<uint16_t> TcpConnectionAcceptor::listeningPort() const {
+  if (!serverSocket_) {
+    return folly::none;
```
phoad: ConnectionAcceptor's constructor accepts Options parameter. This struct has a "port" value in it. 
So if we don't have serverSocket_, I believe we should provide this "port" value.
Perhaps, we should not ask for serverSocket_->getAddress().getPort() at all if the port value in the options parameter is not 0.

src/transports/tcp/TcpConnectionAcceptor.h
```diff
@@ -53,6 +53,11 @@ class TcpConnectionAcceptor : public ConnectionAcceptor {
    */
   void stop() override;
 
+  /**
+   * Get the port being listened on.
+   */
+  folly::Optional<uint16_t> listeningPort() const override;
```
phoad: Does this need to be optional at all? Just return value of the options.port should be ok?

src/transports/tcp/TcpConnectionAcceptor.cpp
alexmalyshev: That's more confusing than it needs to be.  The point of the function is to return the port currently being listened on.  If there is no port being listened on, then we should return some kind of "null" value.  Returning the value in options.port is odd because that's the port we _wanted_ to listen on, not the actual port we _are_ listening on (e.g. wanted to listen on TCP port 0, actually listening on TCP port 25023).

tl;dr I don't think this function should return port N if the acceptor isn't actually listening on port N at the given moment.

phoad: As the method name is "listeningPort" your way seems logical to me too.

test/RequestStreamTest.cpp
```diff
@@ -32,9 +32,8 @@ class TestHandlerSync : public rsocket::RSocketResponder {
 };
 
 TEST(RequestStreamTest, HelloSync) {
-  auto port = randPort();
-  auto server = makeServer(port, std::make_shared<TestHandlerSync>());
-  auto client = makeClient(port);
+  auto server = makeServer(std::make_shared<TestHandlerSync>());
+  auto client = makeClient(*server->listeningPort());
```
phoad: nit: I did not like the idea of having no listening port (optional), null return value. So we force the caller to always check if the return value is null, then the server might not be running. So even though the caller might be sure, they should always write if/else code. 
So we force this method to give information about if the server is really running or not.

alexmalyshev:
Easy way to see the effects of this PR is running

  `./tests --gtest_repeat=100 --gtest_filter='RSocketClientServer.*'`

On master that usually dies around iteration 30 for me with an 'address already in use error'.  Applying this PR and it never fails locally for me.

@lehecka May I push this?  I can't right now due to you requesting changes.

PR #536, opened 2017-06-20
Add a few override keywords, reflow lines past 80 columns

alexmalyshev:


phoad:
Did you run "clang-format" on these files? I think some of us directly run "clang-format" on the files and auto format them.

alexmalyshev:
I ran it on the 80 column violations, but nothing else.  clang-format doesn't automatically insert `override` keywords.

PR #538, opened 2017-06-21
Update folly to 2017.06.19 release

alexmalyshev:


alexmalyshev:
> To learn the policy, so we do bump up the version as soon as the new version becomes available?

We don't really have an official policy :/.  It's been very ad-hoc so far.  We should probably either try to update on every release, or every other release IMO.

This update is particularly important because it's 3 months over due, and it has an important change that fixes some ScopedEventBaseThread crashes we hit on shutdown (rarely).

alexmalyshev:
I just realized this PR probably doesn't rebuild folly at all, as it's using what's already been built in the Travis container cache (see https://travis-ci.org/rsocket/rsocket-cpp/jobs/245167261, `scripts/build_folly.sh` ran in 0.02s).

lexs:
You'll have to set FOLLY_INSTALL_DIR to include the version to fix the cache pollution issue.

vitaut:
I'll update the build config to automatically rebuild cache on version change.

vitaut:
Tried building this new version of folly and gcc wasn't super happy:
```
In file included from ./../folly/String.h:33:0,
                 from ./../folly/Format.h:27,
                 from detail/CacheLocality.cpp:28:
./../folly/FBVector.h: In lambda function:
./../folly/FBVector.h:1432:49: error: parameter packs not expanded with '...':
           M_construct(start, std::forward<Args>(args)...);
                                                 ^
./../folly/FBVector.h:1432:49: note:         'args'
```

alexmalyshev:
Closing in favor of https://github.com/rsocket/rsocket-cpp/pull/543

PR #539, opened 2017-06-21
Fix mishandling future in TcpConnectionAcceptor::start()

alexmalyshev:
It's supposed to return a future that will be fulfilled once the acceptor is
done initializing.  However what we're doing instead is making a manual call to
runInEventBaseThread() and then returning folly::unit synchronously.

Use via() to tie the asynchronous runInEventBaseThread() to the return future
value.  Also add a test for port collision to vet that we can guarantee we hit
it with two servers on the same port, without racing on when the acceptors
actually initialize.


alexmalyshev:
Okay that stack-buffer overflow looks scary, looking...

alexmalyshev:
Not needed anymore, the `folly::via` change landed already in a separate PR.  The test issue is unrelated here, and I'll address it in a follow up PR.

PR #540, opened 2017-06-21
Move build config from Travis to CMake

vitaut:
More build system improvements:

* Move build configuration from `.travis.yml` to CMake to make reproducing and debugging issues locally easier.
* Add `RSOCKET_CC` CMake variable for specifying the compiler, e.g. `cmake -DRSOCKET_CC=gcc-4.9`. Unlike Travis's `compiler` option it handles version correctly and unlike `CC` & `CXX` compiler can be specified once. This replaces a ton of repetitive ifs in the Travis config.
* Make build stop on the first failure (except for tests for now) on Travis.
* Replace `build_folly.sh` with a simple `ExternalProject_Add` command and added CMake import target `folly` to simplify setting up dependencies. Now you only need `target_link_libraries(<target> folly)` which will set up libraries, include directories and necessary flags. These will propagate transitively as expected. Folly installation is controlled by the CMake `RSOCKET_INSTALL_DEPS` variable.
* Moved code coverage from Travis to the build target `coverage` so that it can be run locally with `make coverage`.
* Write coverage files to the build instead of source directory.
* Cleanup `.travis.yml`.
* Don't install valgrind on Travis.


vitaut:
As before, tests are failing due to unrelated "Broken promise for type name" issue.

CMakeLists.txt
```diff
@@ -33,6 +83,48 @@ add_definitions(-DOSS)
...
+set_property(TARGET folly PROPERTY IMPORTED_LOCATION ${FOLLY_LIBRARY})
+set_property(TARGET folly
+  APPEND PROPERTY INTERFACE_LINK_LIBRARIES
+  ${EXTRA_LINK_FLAGS} ${CMAKE_THREAD_LIBS_INIT})
+if (TARGET folly-ext)
+  add_dependencies(folly folly-ext)
+endif ()
+
+# Folly includes are marked as system to prevent errors on non-standard
+# extensions when compiling with -pedantic and -Werror.
```
alexmalyshev: Neat.

.travis.yml
```diff
@@ -76,32 +78,37 @@ matrix:
             - *common_deps
             - g++-6
 
+env:
+  global:
+    - BUILD_TYPE=Debug CPP_VERSION=14 ASAN=On
```
alexmalyshev: I wonder if we really need the CPP_VERSION variable.  We make too much use of C++14 features to compile with `-std=c++11 -pedantic -Werror`.

vitaut: Probably not. Will remove in the next PR.

PR #542, opened 2017-06-22
Removing unused named parameters

alexmalyshev:
Don't use it, don't name it.  Vast majority of these cases were `StreamId
streamId` in an override of handleRequestStream().

Remove our use of -Wno-unused-parameter.


phoad:
LGTM

PR #543, opened 2017-06-22
Build system improvements

vitaut:
* Rebuild folly if version changes and update to version v2017.06.19.00.
* Use the same C++ compiler to compile both folly and rsocket, because folly v2017.06.19.00 fails to compile with the default g++-4.8 compiler.
* Add libevent and libpthreads to folly's interface link libraries.
* Use automatic dependency propagation for folly and pthreads instead of listing link libraries manually which is tedious and error-prone.
* Read the list of apt packages from .travis.yml in Vagrantfile.
* Set C++ standard version to 14 in CMake instead of Travis config as discussed in #540.

### DIFF
--- a/scripts/Vagrantfile
+++ b/scripts/Vagrantfile
@@ -3,6 +3,12 @@
 # Vagrant configuration that creates Travis-CI compatible build environment
 # for rsocket.
 
+# Get the APT dependencies from the Travis config.
+require 'yaml'
+yml = YAML.load_file(File.dirname(__dir__) + '/.travis.yml')
+apt = yml['addons']['apt']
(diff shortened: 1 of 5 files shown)


scripts/Vagrantfile
```diff
@@ -3,6 +3,12 @@
 # Vagrant configuration that creates Travis-CI compatible build environment
 # for rsocket.
 
+# Get the APT dependencies from the Travis config.
+require 'yaml'
+yml = YAML.load_file(File.dirname(__dir__) + '/.travis.yml')
```
alexmalyshev: Convenient.

PR #544, opened 2017-06-23
Remove assert for unimplemented resumption

nickgg:
This just aborts the server when a client tries to resume, instead let's close the connection semi-gracefully.


src/RSocketServer.cpp
```diff
@@ -119,7 +119,9 @@ void RSocketServer::onRSocketResume(
   // We don't need to check for isShutdown_ here since all callbacks are
   // processed by this time.
 
-  CHECK(false) << "not implemented";
+  LOG(ERROR) << "onRSocketResume: not implemented";
+  frameTransport->close(
```
alexmalyshev: Need to name the function's parameter, it got purged in a previous diff where I removed all unused parameters.

nickgg:
Test failures in the tck test for c: java and s:cpp - feels unrelated

nickgg:
Is this hung?

alexmalyshev:
Just needs a merge-to-master to pick up the latest CMake changes I think.

I'll handle merging in the PR past that, we've been hitting issues with port-in-use flakiness and a little bit of flakiness from the TCK tests.  Nothing a few kicks to the build doesn't fix.

nickgg:
That doesn't look like what I wanted to do, I'll close and make a new PR.

PR #545, opened 2017-06-24
Add -g flag to common CXX flags

alexmalyshev:
Noticed ASAN errors in Vagrant VM were missing line numbers.  Adding -g did the
trick.  Good to have on all builds, debug and opt.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,6 +197,7 @@ set(CMAKE_CXX_STANDARD 14)
 # Common configuration for all build modes.
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Woverloaded-virtual")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -173,6 +173,7 @@ endif()
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++${META_CXX_STD} -pedantic")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Woverloaded-virtual")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
```
vitaut: Isn't the default build mode debug? Also there is `RelWithDebInfo` build type for this.

alexmalyshev: I prefer debuginfo on all builds to be honest.  I feel like that's worth it, and doesn't come at too much of a cost here (small library, not going to bloat size or compile time too much).

vitaut: Fair enough.

PR #550, opened 2017-06-27
Update DuplexConnection implementations to use yarpl Subscriber/Subscription

phoad:
The SubscriptionBase and SubscriberBase classes are deleted and the code that is using these classes are updated to use yarpl::flowable::Subscription/Subscriber.

Other than code cleanup, the major change is in TcpDuplexConnection.h/cpp files.


yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -29,6 +29,10 @@ class Subscriber : public virtual Refcounted {
     subscription_.reset();
   }
 
+  void onError() {
```
alexmalyshev: This doesn't seem necessary.  The only use is in `FramedWriter::onError(eptr)`, and you can just call `SubscriberBase::onError(eptr)`.  It incurs a useless copy of `eptr`, but I think that's better than adding this overload.

yarpl/include/yarpl/flowable/Flowable.h
```diff
@@ -10,6 +10,7 @@
 #include <type_traits>
 #include <utility>
 
+#include <folly/ExceptionWrapper.h>
```
alexmalyshev: Why/where is this necessary?

yarpl/include/yarpl/Refcounted.h
```diff
@@ -176,35 +181,124 @@ bool operator==(const Reference<T>& lhs, const Reference<U>& rhs) noexcept {
   return lhs.get() == rhs.get();
 }
 
+template <typename T>
+bool operator==(const Reference<T>& lhs, std::nullptr_t) noexcept {
```
alexmalyshev: Are these overloads actually necessary?  I would figure `someRef == nullptr` would just forward to `operator==(const Reference<T>&, const Reference<T>&)` because `Reference<T>` has an implicit ctor for `std::nullptr_t`.

src/transports/tcp/TcpDuplexConnection.cpp
```diff
@@ -127,76 +135,75 @@ class TcpReaderWriter : public ::folly::AsyncTransportWrapper::WriteCallback,
...
 
-  void onCompleteImpl() noexcept override {
+  void onComplete() noexcept override {
     CHECK(tcpReaderWriter_);
     auto tcpReaderWriter = std::move(tcpReaderWriter_);
     tcpReaderWriter->close();
   }
 
-  void onErrorImpl(folly::exception_wrapper ew) noexcept override {
+  void onError(const std::exception_ptr ew) noexcept override {
```
alexmalyshev: Two nits: I'd rename `ew` to `eptr` and drop the `const`.

src/internal/Common.h
```diff
@@ -25,6 +25,8 @@ typedef Range<const char*> StringPiece;
 
 namespace rsocket {
 
+constexpr int64_t kMaxRequestN = std::numeric_limits<int32_t>::max();
```
alexmalyshev: `yarpl::credits::kNoFlowControl` in `yarpl/utils/credits.h`?  Any why is int32_t::max() being set into an int64_t?

src/framing/FramedWriter.h
```diff
@@ -4,9 +4,7 @@
 
 #include <folly/ExceptionWrapper.h>
```
alexmalyshev: Can likely drop this include now.

yarpl/include/yarpl/Refcounted.h
phoad: When I check folly/io/DelayedDesctructionBase.h file, which is similar to this implementation, I see that it is providing these functions also. Though it doesn't provide implicit constructor with std::nullptr_t.
As there is implicit constructor, it seems like these methods are not needed..

src/internal/Common.h
phoad: https://github.com/rsocket/rsocket/blob/master/Protocol.md#request_n-frame-0x08
If we check here, the limit is 2^31-1. Which is not the same as the limit of yarpl. I have created another issue to solve this
https://github.com/rsocket/rsocket-cpp/issues/561

src/framing/FramedDuplexConnection.cpp
```diff
@@ -7,37 +7,29 @@
...
 }
 
 void FramedDuplexConnection::setInput(
-    std::shared_ptr<Subscriber<std::unique_ptr<folly::IOBuf>>> framesSink) {
-  CHECK(!inputReader_);
-  inputReader_ = std::make_shared<FramedReader>(
-      std::move(framesSink), executor_, protocolVersion_);
-  connection_->setInput(inputReader_);
+    yarpl::Reference<Subscriber<std::unique_ptr<folly::IOBuf>>> framesSink) {
+  if(!inputReader_) {
```
phoad: When I try to switch the subscriber of a connection, I am getting failure - this check fails. I need to check if CHECK(!inputReader_) should stay or I need to make this update?

yarpl/include/yarpl/Refcounted.h
phoad: These lines are added for optimization purposes. std::shared_ptr also has the same lines. It saves the compiler from generating an instance of Reference<nullptr_t>. Otherwise, for each such check
```if (variable == nullptr)```
we would be creating an instance of Reference<nullptr_t>.

File: src/framing/FrameSerializer.cpp [hunk 8]
```diff
@@ -17,7 +17,7 @@ namespace rsocket {
 
 // TODO: this should default to 1.0 when we deploy successfully
 constexpr const ProtocolVersion ProtocolVersion::Latest =
-    FrameSerializerV1_0::Version;
+    FrameSerializerV0_1::Version;
```
lehecka: This line should be reverted. It has nothing to do with this change.

src/framing/FramedReader.cpp
```diff
@@ -134,47 +133,37 @@ void FramedReader::parseFrames() {
...
-  payloadQueue_.move(); // equivalent to clear(), releases the buffers
-  if (auto subscription = std::move(streamSubscription_)) {
-    subscription->cancel();
-  }
-  if (auto subscriber = std::move(frames_)) {
-    subscriber->onComplete();
-  }
+void FramedReader::setInput(yarpl::Reference<Subscriber<std::unique_ptr<folly::IOBuf>>>
+              frames) {
+  CHECK(!frames_) <<"FrameReader should be closed before setting another input.";
```
lehecka: nit: space

src/framing/FrameSerializer.cpp
phoad: Yes. Good catch. It is recently upgraded to 1_0.

src/framing/FramedReader.cpp
phoad: let me apply clang-format on this file.

phoad:
@lehecka @alexmalyshev 
I have made updates on the diff related to your comments.

src/transports/tcp/TcpDuplexConnection.cpp
```diff
@@ -64,20 +76,20 @@ class TcpReaderWriter : public ::folly::AsyncTransportWrapper::WriteCallback,
     if (auto socket = std::move(socket_)) {
       socket->close();
     }
+    if (auto outputSubscription = std::move(outputSubscription_)) {
```
alexmalyshev: Why is `subscription_->cancel()` moved up above `subscriber_->onComplete()` here in `close(), but not in `closeErr()`?

.gitignore
```diff
@@ -62,12 +62,14 @@ install_manifest.txt
 # vim temp files
 *.swp
 *.un~
+*.*~
```
alexmalyshev: Can we drop these changes?  This PR is big enough :).  Feel free to put them in a separate PR and I'll stamp it.

yarpl/include/yarpl/Refcounted.h
alexmalyshev: Hmm calling `Reference<T>(std::nullptr_t)` should get inlined, but I'm fine with leaving this as is.

src/internal/Common.h
alexmalyshev: Good point.

src/transports/tcp/TcpDuplexConnection.cpp
```diff
@@ -127,76 +136,80 @@ class TcpReaderWriter : public ::folly::AsyncTransportWrapper::WriteCallback,
...
+  void request(int64_t n) noexcept override {
+    DCHECK(tcpReaderWriter_);
+    DCHECK(n == kMaxRequestN)
+        << "TcpDuplexConnection doesnt support proper flow control";
   }
 
-  void cancelImpl() noexcept override {
-    tcpReaderWriter_->close();
+  void cancel() noexcept override {
+    tcpReaderWriter_->setInput(nullptr);
```
alexmalyshev: Why replace `close()` with `setInput(nullptr)`?

phoad: We would like to be able to call the setInput & getOutput functions on an instance of TcpReaderWriter multiple times. So when an instance of TcpInputSubscription calls cancel, it should not close the connection. One subscription/subscriber might cancel and then can be replaced with another one.

src/transports/tcp/TcpDuplexConnection.cpp
phoad: Good catch. Because the closeErr method is a new method, which should also be updated. Updating now.

phoad:
I have completed the implementation of the unit tests for this PR. 

https://github.com/phoad/rsocket-cpp/pull/11/files

To be able to see the difference more clearly, I have made a merge request on top of this update, to be checked in as this PR goes in.

src/framing/FramedReader.cpp
```diff
@@ -66,35 +69,31 @@ size_t FramedReader::readFrameLength() const {
...
 
-void FramedReader::onSubscribeImpl(
-    std::shared_ptr<Subscription> subscription) noexcept {
-  CHECK(!streamSubscription_);
-  streamSubscription_ = std::move(subscription);
-  frames_->onSubscribe(shared_from_this());
+void FramedReader::onSubscribe(
+    yarpl::Reference<Subscription> subscription) noexcept {
+  SubscriberBase::onSubscribe(subscription);
+  subscription->request(kMaxRequestN);
```
phoad: This "request" in here makes the Http2StreamReader to fail.

```
void Http2StreamHandler::request(int64_t n) noexcept {
  // since proxygen doesn't support flow control, there is no point in allowing
  // more than n = 1. The underalying http stream can be transported over
  // as many chunks as necessary and we will always call onNext with the chain
  // of all chunks we received so far
  CHECK_LE(n, 1) << "can be called with n <= 1";
```
So I have removed this line and tested that. It seems still working fine without this line. I will verify with @lehecka .

phoad: I have synced with Ondrej. We will keep this line as is and update Http2StreamHandler.

Issue #488, opened 2017-05-26
Failed Build: Broken promise


benjchristensen:
Failed build here: https://travis-ci.org/rsocket/rsocket-cpp/jobs/236401561#L3140 and here: https://travis-ci.org/rsocket/rsocket-cpp/jobs/236401562#L3119

```
I0526 15:10:25.343734 23830 RSocketServer.cpp:72] Starting RSocketServer
I0526 15:10:25.344848 23830 TcpConnectionAcceptor.cpp:83] Starting TCP listener on port 46979 with 2 request threads
I0526 15:10:25.346344 23848 TcpConnectionAcceptor.cpp:103] Listening on 0.0.0.0:46979
I0526 15:10:25.347144 23830 TcpConnectionAcceptor.cpp:111] Shutting down TCP listener
unknown file: Failure
C++ exception with description "Broken promise for type name `St10shared_ptrIN7rsocket16RSocketRequesterEE`" thrown in the test body.
[  FAILED  ] RequestStreamTest.HelloAsync (6 ms)
[----------] 2 tests from RequestStreamTest (19 ms total)
```
and
```
[ RUN      ] RequestResponseTest.Hello
I0526 15:12:15.132006 23954 RSocketServer.cpp:72] Starting RSocketServer
I0526 15:12:15.143609 23954 TcpConnectionAcceptor.cpp:83] Starting TCP listener on port 49071 with 2 request threads
I0526 15:12:15.152891 23954 TcpConnectionAcceptor.cpp:111] Shutting down TCP listener
I0526 15:12:15.156368 23959 TcpConnectionAcceptor.cpp:103] Listening on 0.0.0.0:49071
unknown file: Failure
C++ exception with description "Broken promise for type name `St10shared_ptrIN7rsocket16RSocketRequesterEE`" thrown in the test body.
[  FAILED  ] RequestResponseTest.Hello (29 ms)
```

It is two different tests non-deterministically failing on server shutdown. So it seems there are issues with changes done in https://github.com/rsocket/rsocket-cpp/pull/486

alexmalyshev:
#552 got rid of these test failures.

PR #552, opened 2017-06-27
Export listener port in ConnectionAcceptor and RSocketServer

alexmalyshev:
Lets us assign port 0 to servers, and then read the automatically assigned port
and connect a client.


vitaut:
Does this fix #488?

alexmalyshev:
Most of the errors on Travis were because of "address-in-use" errors, which this PR definitely got rid of.

PR #554, opened 2017-06-28
s|const std::exception_ptr|std::exception_ptr|g

alexmalyshev:
Don't put `const` on a class type in a function's parameter list.

### DIFF
--- a/examples/util/ExampleSubscriber.cpp
+++ b/examples/util/ExampleSubscriber.cpp
@@ -55,7 +55,7 @@ void ExampleSubscriber::onComplete() noexcept {
   terminalEventCV_.notify_all();
 }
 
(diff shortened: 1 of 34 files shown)


phoad:
Putting "const" in front of std::exception_ptr forces the implementation to do not change the provided parameter's value, isn't it? Otherwise the callee can update the passed parameter?

Passing by value is ok, good that we don't need "const std::exception_ptr &eptr".
  http://en.cppreference.com/w/cpp/error/exception_ptr

alexmalyshev:
Using function declarations of the form `TRet func(const TValue);` isn't a common practice in C++ because it's actually the same declaration as `TRet func(TValue);`.  And for definitions the common practice is to keep the parameters mutable.

It feels weird to break this pattern just for `std::exception_ptr`, it's not different than any other struct type.

PR #556, opened 2017-06-28
Replace almost all uses of CollectingSubscriber with TestSubscriber

alexmalyshev:
The remaining places are where we use noncopyable types, something which
TestSubscriber doesn't support.

### DIFF
--- a/yarpl/test/FlowableTest.cpp
+++ b/yarpl/test/FlowableTest.cpp
@@ -3,17 +3,18 @@
 #include <type_traits>
 #include <vector>
 #include <gtest/gtest.h>
+
 #include "yarpl/Flowable.h"
+#include "yarpl/flowable/TestSubscriber.h"
 #include "yarpl/utils/ExceptionString.h"
 
 namespace yarpl {
 namespace flowable {
 namespace {
 
-void unreachable() {
-  EXPECT_TRUE(false);
-}
-
+/*
+ * Used in place of TestSubscriber where we have move-only types.
+ */
 template <typename T>
 class CollectingSubscriber : public Subscriber<T> {
  public:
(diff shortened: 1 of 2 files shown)


yarpl/test/FlowableTest.cpp
```diff
@@ -279,116 +281,106 @@ TEST(FlowableTest, IgnoreElements) {
...
+      std::make_exception_ptr(std::runtime_error(kMsg)));
+  auto subscriber = make_ref<TestSubscriber<int>>();
+  flowable->subscribe(subscriber);
 
-  EXPECT_EQ(collector->complete(), false);
-  EXPECT_EQ(collector->error(), true);
-  EXPECT_EQ(collector->errorMsg(), "something broke!");
+  EXPECT_FALSE(subscriber->isComplete());
+  EXPECT_TRUE(subscriber->isError());
+  subscriber->assertOnErrorMessage(kMsg);
```
benjchristensen: Should we use EXPECT_EQ with a getErrorMsg method for consistency? Or use subscriber->assertComplete() etc instead of EXPECT on the others?

benjchristensen: For example there is assertSuccess()

alexmalyshev: We should be using EXPECT.  It lets the test continue running and is more consistent with the rest of our gtests.

I can add a getErrorMsg() method.

phoad:
Is it possible to make the TestSubscriber to work with value of T being std::unique_ptr?

DuplexConnection accepts Subscriber of std::unique_ptr<folly::IOBuf>
```
  virtual void setInput(
      yarpl::Reference<yarpl::flowable::Subscriber<std::unique_ptr<folly::IOBuf>>>
          framesSink) = 0;
```

It would be helpful to have TestSubscriber to be used for DuplexConnection also.

alexmalyshev:
> Is it possible to make the TestSubscriber to work with value of T being std::unique_ptr?

No, because TestSubscriber supports forwarding its T items over to an inner subscriber.  This implies that T must be copyable, as it has to be stored in TestSubscriber's vector and sent to the inner subscriber's onNext(T).

We'd have to split that functionality out into another class if we wanted to support move-only types like unique_ptr.

PR #557, opened 2017-06-29
Warm-resumption

somasun:
### Client:
```
auto client = RSocket::createClient(...);
auto requester = client->connect();
...
client->disconnect();
...
client->resume()
  .via(my_evb)
  .then([]{ 
    // resumption succeeded; continue with the same requester 
  })
  .onError([] { 
    // resumption failure; create a new client
  });
```
Previously RSocketClient::connect() would return a new RSocketRequester everytime. It seemed a little unconventional, as you typically connect() a client only once and disconnect() it before attempting to reconnect.  With this implementation, the application can call connect() only once (if connect is successful).  And then it can resume()/disconnect() on the connection.  If application wants multiple RSocketRequesters, it has to create multiple RSocketClients.

### Server:
The ConnectionManager class holds the RSocketStateMachines and the ResumptionTokens.  So the server can resume connections without the application having to do anything.  In a subsequent diff I will let `RSocketServer::start()` take a callback which will be invoked for each resumption attempt.  The server-application can then decide (using that callback) whether to allow resumption or not.  This will be useful in cases where the server does not want to resume any further connections because of load or any other reason.  It is helpful even just for stats.

## Testing
Wrote a simple warm-resumption server and client. The server can be started in resumable/non-resumable modes.  Verified the correct behavior with both v1.0 and v0.1 interactions.  The frames get buffered in StreamState when the FrameTransport is disconnected.  The Frames get resent if the client/server request older positions.  The server sends ERROR frames for non-resumable cases. 

## TODO
* Ability to migrate RSocketStateMachine to different eventBase (this is needed to run multiple worker threads for the server)
* Add resumeCallback for RSocketServer::start()
* Make RSocketClient thread safe
* Write integration tests for warm-resumption
* Add more checks
* The sample warm-resumption client/server could be more exhaustive in handling the different cases.

### DIFF
--- a/examples/warm-resumption/WarmResumption_Client.cpp
+++ b/examples/warm-resumption/WarmResumption_Client.cpp
@@ -66,7 +66,7 @@ class HelloSubscriber : public virtual yarpl::Refcounted,
 };
 }
 
-std::unique_ptr<RSocketClient> getClient(
+std::unique_ptr<RSocketClient> getClientAndRequestStream(
     yarpl::Reference<HelloSubscriber> subscriber) {
   folly::SocketAddress address;
   address.setFromHostPort(FLAGS_host, FLAGS_port);
@@ -85,7 +85,7 @@ int main(int argc, char* argv[]) {
   folly::init(&argc, &argv);
 
   auto subscriber1 = yarpl::make_ref<HelloSubscriber>();
-  auto client = getClient(subscriber1);
+  auto client = getClientAndRequestStream(subscriber1);
 
   subscriber1->request(7);
 
@@ -110,10 +110,19 @@ int main(int argc, char* argv[]) {
         subscriber1->cancel();
       })
       .onError([](folly::exception_wrapper ex) {
-        LOG(INFO) << "Resumption Failed: " << ex;
+        LOG(INFO) << "Resumption Failed: " << ex.what();
+        try {
+          ex.throw_exception();
+        } catch (const ResumptionException& e) {
+          LOG(INFO) << "ResumeException";
+        } catch (const ConnectionException& e) {
+          LOG(INFO) << "ConnectionException";
+        } catch (const std::exception& e) {
+          LOG(INFO) << "UnknownException " << typeid(e).name();
+        }
         // Create a new client
         auto subscriber2 = yarpl::make_ref<HelloSubscriber>();
-        auto client = getClient(subscriber2);
+        auto client = getClientAndRequestStream(subscriber2);
         subscriber2->request(7);
         while (subscriber2->rcvdCount() < 7) {
           std::this_thread::yield();
--- a/examples/warm-resumption/WarmResumption_Client.cpp
+++ b/examples/warm-resumption/WarmResumption_Client.cpp
@@ -1,0 +1,118 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#include <iostream>
+
+#include <folly/init/Init.h>
+#include <folly/io/async/ScopedEventBaseThread.h>
+#include <folly/portability/GFlags.h>
+
+#include "examples/util/ExampleSubscriber.h"
+#include "src/RSocket.h"
+#include "src/transports/tcp/TcpConnectionFactory.h"
+
+#include "yarpl/Flowable.h"
+
+using namespace rsocket_example;
+using namespace rsocket;
+
+DEFINE_string(host, "localhost", "host to connect to");
+DEFINE_int32(port, 9898, "host:port to connect to");
+
+namespace {
+
+class HelloSubscriber : public virtual yarpl::Refcounted,
+                        public yarpl::flowable::Subscriber<Payload> {
+ public:
+  void request(int n) {
+    LOG(INFO) << "... requesting " << n;
+    while (!subscription_) {
+      std::this_thread::yield();
+    }
+    subscription_->request(n);
+  }
+
+  void cancel() {
+    if (auto subscription = std::move(subscription_)) {
+      subscription->cancel();
+    }
+  }
+
+  int rcvdCount() const {
+    return count_;
+  };
+
+ protected:
+  void onSubscribe(yarpl::Reference<yarpl::flowable::Subscription>
+                       subscription) noexcept override {
+    subscription_ = subscription;
+  }
+
+  void onNext(Payload element) noexcept override {
+    LOG(INFO) << "Received: " << element.moveDataToString() << std::endl;
+    count_++;
+  }
+
+  void onComplete() noexcept override {
+    LOG(INFO) << "Received: onComplete";
+  }
+
+  void onError(std::exception_ptr) noexcept override {
+    LOG(INFO) << "Received: onError ";
+  }
+
+ private:
+  yarpl::Reference<yarpl::flowable::Subscription> subscription_{nullptr};
+  std::atomic<int> count_{0};
+};
+}
+
+int main(int argc, char* argv[]) {
+  FLAGS_logtostderr = true;
+  FLAGS_minloglevel = 0;
+  folly::init(&argc, &argv);
+
+  folly::SocketAddress address;
+  address.setFromHostPort(FLAGS_host, FLAGS_port);
+
+  // create a client which can then make connections below
+  auto client = RSocket::createClient(
+      std::make_unique<TcpConnectionFactory>(std::move(address)));
+
+  // connect and wait for connection
+  SetupParameters setupParameters;
+  setupParameters.resumable = true;
+  auto rs = client->connect(std::move(setupParameters)).get();
+
+  // subscriber
+  auto subscriber = yarpl::make_ref<HelloSubscriber>();
+
+  // perform request on connected RSocket
+  rs->requestStream(Payload("Jane"))->subscribe(subscriber);
+
+  subscriber->request(7);
+
+  while (subscriber->rcvdCount() < 3) {
+    std::this_thread::yield();
+  }
+
+  client->disconnect(std::runtime_error("disconnect triggered from client"));
+
+  // This request should get buffered
+  subscriber->request(3);
+
+  client->resume(std::make_unique<DefaultResumeCallback>());
+
+  subscriber->request(3);
+
+  while (subscriber->rcvdCount() < 13) {
+    std::this_thread::yield();
+  }
+
+  std::getchar();
+
+  subscriber->cancel();
+
+  std::getchar();
+
+  return 0;
+}
--- a/examples/warm-resumption/WarmResumption_Client.cpp
+++ b/examples/warm-resumption/WarmResumption_Client.cpp
@@ -66,53 +66,62 @@ class HelloSubscriber : public virtual yarpl::Refcounted,
 };
 }
 
+std::unique_ptr<RSocketClient> getClient(
+    yarpl::Reference<HelloSubscriber> subscriber) {
+  folly::SocketAddress address;
+  address.setFromHostPort(FLAGS_host, FLAGS_port);
+  auto client = RSocket::createClient(
+      std::make_unique<TcpConnectionFactory>(std::move(address)));
+  SetupParameters setupParameters;
+  setupParameters.resumable = true;
+  auto rs = client->connect(std::move(setupParameters)).get();
+  rs->requestStream(Payload("Jane"))->subscribe(subscriber);
+  return client;
+}
+
 int main(int argc, char* argv[]) {
   FLAGS_logtostderr = true;
   FLAGS_minloglevel = 0;
   folly::init(&argc, &argv);
 
-  folly::SocketAddress address;
-  address.setFromHostPort(FLAGS_host, FLAGS_port);
+  auto subscriber1 = yarpl::make_ref<HelloSubscriber>();
+  auto client = getClient(subscriber1);
 
-  // create a client which can then make connections below
-  auto client = RSocket::createClient(
-      std::make_unique<TcpConnectionFactory>(std::move(address)));
+  subscriber1->request(7);
 
-  // connect and wait for connection
-  SetupParameters setupParameters;
-  setupParameters.resumable = true;
-  auto rs = client->connect(std::move(setupParameters)).get();
-
-  // subscriber
-  auto subscriber = yarpl::make_ref<HelloSubscriber>();
-
-  // perform request on connected RSocket
-  rs->requestStream(Payload("Jane"))->subscribe(subscriber);
-
-  subscriber->request(7);
-
-  while (subscriber->rcvdCount() < 3) {
+  while (subscriber1->rcvdCount() < 3) {
     std::this_thread::yield();
   }
-
   client->disconnect(std::runtime_error("disconnect triggered from client"));
 
   // This request should get buffered
-  subscriber->request(3);
+  subscriber1->request(3);
 
-  client->resume(std::make_unique<DefaultResumeCallback>());
+  folly::ScopedEventBaseThread worker_;
 
-  subscriber->request(3);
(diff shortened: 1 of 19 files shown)


yschimke:
Could you split out the reformat into a separate PR and immediately land.  Some of the files are > 50% reformat so its a toucher review than it needs to be.

src/RSocketClient.cpp
```diff
@@ -60,6 +63,35 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
   return future;
 }
 
+folly::Future<folly::Unit> RSocketClient::resume(
```
lexs: Why does this take a Future as well as a callback?

src/RSocketClient.cpp
```diff
@@ -41,11 +44,11 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
...
     networkStats = std::move(networkStats),
-    promise = std::move(promise)](
-      std::unique_ptr<DuplexConnection> connection,
-      folly::EventBase& eventBase) mutable {
+    promise = std::move(promise)
+  ](std::unique_ptr<DuplexConnection> connection,
+    folly::EventBase & eventBase) mutable {
     VLOG(3) << "onConnect received DuplexConnection";
-
+    evb_ = &eventBase;
```
lexs: This doesn't seem great? Switching threads for every resumption?

src/RSocketClient.h
```diff
@@ -56,13 +60,26 @@ class RSocketClient {
...
+      std::unique_ptr<ClientResumeStatusCallback> resumeCallback =
+          std::make_unique<DefaultResumeCallback>());
+
+  void disconnect(folly::exception_wrapper ex);
 
  private:
   std::unique_ptr<ConnectionFactory> connectionFactory_;
   std::unique_ptr<RSocketConnectionManager> connectionManager_;
+
+  std::shared_ptr<RSocketStateMachine> stateMachine_;
```
lexs: Aren't we really close to this being the old `ReactiveSocket` class? We moved away from that for some reason right?

src/RSocketClient.cpp
```diff
@@ -60,6 +63,35 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
...
+    stateMachine_->tryClientResume(
+        resumeToken_, frameTransport, std::move(resumeCallback));
+    promise.setValue(folly::Unit());
+  });
+  return future;
+}
+
+void RSocketClient::disconnect(folly::exception_wrapper) {
+  CHECK(stateMachine_);
+  evb_->runInEventBaseThreadAndWait([this] {
```
lexs: you should not wait here

src/RSocketClient.cpp
somasun: The RSocketClient just keeps track of the evb which the client-connection is using, so that later it can use it for scheduling disconnect in the right thread and to ensure resumption is happening in the same evb-thread.

src/RSocketClient.h
somasun: Not sure I have the context or understand your question completely. I think the previous implementation was ambiguous on ReactiveSocket ownership.  In this implementation the ConnectionManager takes care of properly closing/destroying the RSocketStateMachines.  It is the responsibility of the client to hold the RSocketClient instance.

somasun:
Removed the clang-format changes which my editor automatically added.

benjchristensen:
> The server-application can then decide (using that callback)

That should be `RSocketParameters::ResumeParameters` I believe.

https://github.com/rsocket/rsocket-cpp/blob/master/src/RSocketParameters.h#L47

examples/warm-resumption/WarmResumption_Client.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+  subscriber->request(7);
+
+  while (subscriber->rcvdCount() < 3) {
+    std::this_thread::yield();
+  }
+
+  client->disconnect(std::runtime_error("disconnect triggered from client"));
+
+  // This request should get buffered
```
benjchristensen: I thought we weren't buffering frames in this impl?

examples/warm-resumption/WarmResumption_Client.cpp
```diff
@@ -0,0 +1,118 @@
...
+  // perform request on connected RSocket
+  rs->requestStream(Payload("Jane"))->subscribe(subscriber);
+
+  subscriber->request(7);
+
+  while (subscriber->rcvdCount() < 3) {
+    std::this_thread::yield();
+  }
+
+  client->disconnect(std::runtime_error("disconnect triggered from client"));
```
benjchristensen: Can we also show a version where the connection is broken in a non-clean way? For example, killing the DuplexConnection so the client gets an error and then retries with a resume?

src/RSocketClient.cpp
benjchristensen: Threading is supposed to all be handled at only a single layer, by the "Scheduled" wrappers here: https://github.com/rsocket/rsocket-cpp/tree/master/src/internal

So why is this needed?

benjchristensen: I can sort of understand disconnect ... but what about resumption needs this?

examples/warm-resumption/WarmResumption_Client.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+  // This request should get buffered
+  subscriber->request(3);
+
+  client->resume();
```
benjchristensen: Why doesn't this need a callback?

src/RSocketClient.cpp
```diff
@@ -60,6 +63,30 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
...
+    folly::EventBase & evb) mutable {
+    CHECK(evb_ == &evb);
+    auto frameTransport =
+        std::make_shared<FrameTransport>(std::move(connection));
+    stateMachine_->tryClientResume(
+        resumeToken_, std::move(frameTransport), std::move(resumeCallback));
+  });
+}
+
+void RSocketClient::disconnect(folly::exception_wrapper ex) {
```
benjchristensen: If this is going to move here, then `RSocketRequester` should not have `closeSocket`: https://github.com/rsocket/rsocket-cpp/blob/master/src/RSocketRequester.h#L96

benjchristensen: At least I don't think so.

src/RSocketClient.h
benjchristensen: The benefit of the way it's currently working, prior to this change is that a user only needs to hold the reference to the `RSocketRequester`, they don't need to hold both the `RSocketClient` and `RSocketRequester`.

benjchristensen: Not quite happy with how this change is affecting the API and relationship between RSocketClient and RSocketRequester. Haven't quite figured out the solution yet, but I don't want this merged as is yet.

src/RSocketSetup.h
```diff
@@ -34,6 +34,7 @@ class RSocketSetup {
 
   void createRSocket(
       std::shared_ptr<RSocketResponder> requestResponder,
+      bool resumable = false,
```
benjchristensen: Should we have another approach to injecting this, or is this truly the only bool we'll ever pass in here? I don't want to accumulate a list of primitive args in this API.

For example, what about keepalive timeout config?

src/RSocketClient.cpp
somasun: The "Scheduled" wrappers are for Subscriber/Subscription/Responder, not for stateMachine/duplexConnection.  Here the client schedules a disconnect() directly with the stateMachine, so I have used a raw-evb.  

The check in resumption is for ensuring that the resumed connection is also running in the same evb as the old connection.   With TcpConnectionFactory, this check is unnecessary as the subsequent clients will be in the same worker_.evb.  But a custom ConnectionFactory could give out clients in different evbs for every connect().  At that point we dont want to run our stateMachine bound to old evb_ in this new evb_.  Ondrej has a task to add ability to migrate stateMachines across evbs.  That could change what we do here.

src/RSocketClient.h
somasun: Prior to this change (and after this change) the application would need to store both RSocketClient and RSocketRequester.  The RSocketClient holds ConnectionManager instance.  The ConnectionManager destructor closes and destroys all RSocketStateMachines.  This design seems intentional.  So the application has to hold the RSocketClient instance in addition to the RSocketRequester.   

But I agree  that the RSocketClient-RSocketRequester interaction/ownership is kind of awkward here.  The clients shouldn't have to worry about life-times of both these objects.  Maybe RSocketClient can hold a instance of RSocketRequester?  That way, the client needs to worry about only holding the RSocketClient.

examples/warm-resumption/WarmResumption_Client.cpp
somasun: There is buffering at two places.  
(1) The RSocketStateMachine buffers frames if the frameTransport (DuplexConnection) is disconnected.
(2) Frames and positions are stored in ResumeCache as and when frames arrive/leave the RSocketStateMachine.  This is used for warm-resumption.

I was talking about buffering in (1) here.

examples/warm-resumption/WarmResumption_Client.cpp
somasun: Sure.  It is one of the items I have in the TODO part of the description (6th bullet)

examples/warm-resumption/WarmResumption_Client.cpp
somasun: I left the callback optional, but I think I should make it mandatory.  I don't see a case where the client would want to resume() without a callback.

src/RSocketClient.cpp
somasun: Agreed.  Will remove!

src/RSocketSetup.h
somasun: If there is a better approach we should pursue it.

src/internal/RSocketConnectionManager.cpp
```diff
@@ -87,8 +88,20 @@ void RSocketConnectionManager::manageConnection(
       *this, socket, eventBase);
   newNetworkStats->inner = std::move(socket->networkStats());
   socket->networkStats() = std::move(newNetworkStats);
+  sockets_.lock()->insert({std::move(socket), {eventBase, resumeToken}});
```
lexs: Does this mean that we'll have every connection in this map? Will this cause scaling issues?

somasun: The ConnectionManager stores shared_ptrs to all RSocketStateMachines.  It later uses it to close/destroy the RSocketStateMachines properly on exit (whenever RSocketServer/RSocketClient are destroyed).  We anyway store all the StateMachines. I don't think adding a pointer to these stateMachines would affect scaling signifcantly.  We do acquire lock before adding/removing the StateMachines, but I don't think it will be in the critical path.

src/RSocketClient.cpp
benjchristensen: Thanks for the explanation.

src/RSocketClient.h
benjchristensen: We talked more in person. Will remove close() from Requester. Right now leave connect() as is returning a Requester. Discussed that we could always add later a getRequester() depending on ergonomics as we use this more. Choosing not to add anything new now.

benjchristensen:
> allowing us to merge RSocketRequester and RSocketResponder. A win for symmetry!

Not always a good thing. Rarely does a client have a responder, and it's very different on a server when requester versus responder are being used. Every attempt I've seen of them being blended increases the awkwardness and confusion. 

Also, having an RSocketClient "be" a Requester is also not great, as until it is connected those methods would need to throw an exception. That is why the connect() method returns a Requester that is then ready to use. And why a client would be injected a Responder. It would not be both a Responser and Requester, as that makes no sense.

src/RSocketSetup.h
```diff
@@ -34,6 +34,7 @@ class RSocketSetup {
 
   void createRSocket(
       std::shared_ptr<RSocketResponder> requestResponder,
+      ServerOptions serverOptions = ServerOptions(),
```
benjchristensen: I like this better. Far more flexible going forward.

src/RSocketClient.cpp
```diff
@@ -60,6 +63,51 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
...
+      ResumeCallback(folly::Promise<folly::Unit> promise)
+          : promise_(std::move(promise)) {}
+      void onResumeOk() noexcept override {
+        promise_.setValue(folly::Unit());
+      }
+      void onResumeError(folly::exception_wrapper ex) noexcept override {
+        promise_.setException(ex);
+      }
+      void onConnectionError(folly::exception_wrapper ex) noexcept override {
+        promise_.setException(ex);
```
lexs: Not being able to distinguish a resume error from a connection error will make it impossible to implement retries correctly.

lexs:
> Also, having an RSocketClient "be" a Requester is also not great, as until it is connected those methods would need to throw an exception.

On mobile we currently depend on being able to request* before a connection has been established. This is also needed for fast start.

somasun:
Rebased on top #514.  The `RSocketClient` now remembers the `protocolVersion` which was used in `connect()` and uses that to create FramedDuplexConnections in all subsequent `resume()` operations.

The `resume().onError()` now gets typed exception (ConnectionError vs ResumptionError).

src/internal/RSocketConnectionManager.cpp
lehecka: I understand the intention to keep the application code as simple as possible and so storing the ResumeIdentificationToken here. My concern is that the library is now in the business of storing it, looking up state machine by RID, dealing with problems with duplicity or empty values, ...
I was preferring in pushing this problem into the application code.

src/RSocketRequester.cpp
```diff
@@ -26,14 +26,6 @@ RSocketRequester::~RSocketRequester() {
   }
 }
 
-void RSocketRequester::closeSocket() {
```
lehecka: currently RSocketRequester::closeSocket is the only method to close one rsocket instance on the server (method used by RTGW today). If we remove it, what is the solution for allowing the server to close rsocket instance?

src/internal/RSocketConnectionManager.cpp
somasun: Right. Since most applications are going to repeat the same logic, I felt it is better to embed this in the protocol implementation.  The application wouldn't have to know about StateMachines and ResumeTokens.  May be we could keep this logic in the protocol implementation, but the application can override it if needed ? Anyway, eagerly looking forward to see what you have in your mind.

src/RSocketRequester.cpp
somasun: That is a good point. I did not know about the usage. 
What if we don't create Requester on the server - should we have ways to close individual instance from RSocketResponder as well?  In general I think Requester/Responder should just request/respond.  Should we add ability in RSocketServer to close specific instances ?

lexs: I want us to clear this confusion up and solve this for real. I suggest something like this:
```cpp
struct RSocket {
  // emphasis on non-owning
  virtual void RSocketRequester& getRemote();
  virtual void disconnect() = 0;
  virtual void close() = 0;
};

struct RSocketClient : public RSocket {
  virtual Future resume();
};

// if it doesn't need extra methods, can also be a hard typedef
using RSocketServer = RSocket; 
```

somasun:
The warm-resumption changes are being made in #591 and #601.  Closing this.

PR #558, opened 2017-06-29
Fix rsocket-server to send ERROR frame for unsupported resumption

somasun:
Previously, the rsocket-server would just crash on receiving RESUME frame.  With this change it would send out an ERROR frame to the client and close the DuplexConnection.

Testing:
Ran a client which attempts resumption against this server.  Without this change the server crashes.  With this change the server sends out ERROR and closes the connection 
screen shot 2017-06-29 at 4 47 47 pm

### DIFF
--- a/src/internal/SetupResumeAcceptor.cpp
+++ b/src/internal/SetupResumeAcceptor.cpp
@@ -145,7 +145,12 @@ void SetupResumeAcceptor::processFrame(
 
       if (!onResume(transport, std::move(resumeParams))) {
         transport->outputFrameOrEnqueue(frameSerializer->serializeOut(
-            Frame_ERROR::connectionError("Resumption Failure")));
+            Frame_ERROR::rejectedResume("Resumption Failure")));
+        // We are currently in FrameTransport's callstack where it would
+        // have already acquired a recursive_mutex.  After the following block
+        // of code, FrameTransport goes out of scope and will get destroyed. But
+        // we cant destroy it while holding a recursive_mutex.  So schedule this
+        // block of code to be executed after callstack has unwinded.
         eventBase_->runInEventBaseThread(
             [ this, transport = std::move(transport) ] {
               closeAndRemoveConnection(
--- a/src/internal/SetupResumeAcceptor.cpp
+++ b/src/internal/SetupResumeAcceptor.cpp
@@ -143,8 +143,19 @@ void SetupResumeAcceptor::processFrame(
         break;
       }
 
-      removeConnection(transport);
-      onResume(std::move(transport), std::move(resumeParams));
+      if (!onResume(transport, std::move(resumeParams))) {
+        transport->outputFrameOrEnqueue(frameSerializer->serializeOut(
+            Frame_ERROR::connectionError("Resumption Failure")));
+        eventBase_->runInEventBaseThread(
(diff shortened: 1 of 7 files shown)


alexmalyshev:
The new use of `onResume` in `SetupResumeAcceptor` is awkward.  I think it's better to remove the FrameTransport from the connections set first, and then call `onResume`(what it does now).  It's then the responsibility of `onResume` to close it properly if an error occurs.

Is there any way we could get a small client-server test for this?

src/internal/SetupResumeAcceptor.cpp
```diff
@@ -143,8 +143,19 @@ void SetupResumeAcceptor::processFrame(
         break;
       }
 
-      removeConnection(transport);
-      onResume(std::move(transport), std::move(resumeParams));
+      if (!onResume(transport, std::move(resumeParams))) {
+        transport->outputFrameOrEnqueue(frameSerializer->serializeOut(
+            Frame_ERROR::connectionError("Resumption Failure")));
+        eventBase_->runInEventBaseThread(
```
alexmalyshev: (Post IRL conversation) This definitely deserves a comment, because to the untrained eye (mine) it looks like an unnecessary use of `runInEventBaseThread()`.

src/internal/SetupResumeAcceptor.cpp
```diff
@@ -143,8 +143,19 @@ void SetupResumeAcceptor::processFrame(
         break;
       }
 
-      removeConnection(transport);
-      onResume(std::move(transport), std::move(resumeParams));
+      if (!onResume(transport, std::move(resumeParams))) {
+        transport->outputFrameOrEnqueue(frameSerializer->serializeOut(
+            Frame_ERROR::connectionError("Resumption Failure")));
```
yschimke: Shouldn't this be REJECTED_RESUME	0x00000004?

yschimke: Will need to be added here https://github.com/rsocket/rsocket-cpp/blob/4407a4a0cb0781da3f8217160d253e0742d2ed01/src/framing/Frame.h#L62

lexs: (y) Also "Resumption Failure" seems a bit dishonest, can it be "Resumption not supported" instead?

somasun: `onResume` could return false for a number of reasons.  It just happens that in the current version, resumption is not supported.  Hence I went with a generic "Resumption Failure" message.  Later, we could augment this code to get the actual failure reason from `onResume`

somasun: @yschimke right!  I will change it to REJECTED_RESUME.  It will also involve changing the client side to process this new type.

src/internal/SetupResumeAcceptor.cpp
somasun: will do !

somasun:
Once my other PR #557 gets in, it should be straightforward to write tests.  This is how I tested it locally

```

server.start( /* non_resumable */ mode);
...
auto client = RSocket::createClient()
client.connect();
...
expect client.resume() error out.
```

alexmalyshev:
Let's get this in now as its blocking other tasks, and do the due diligence of adding tests later.

PR #560, opened 2017-06-30
Update default version to v1.0

somasun:
Verified all the tests work with both versions.


lexs:
Can we make these tests parametrized on the version?

PR #562, opened 2017-07-04
Fix coveage and ignore test directories and dependencies

lexs:
Turns out coveralls upload wasn't working because we were running it from the wrong directory (so the coverage file wasn't found).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,6 +118,6 @@ script:
 
 after_success:
     # Upload to coveralls.
-  - if [ -f build/coverage.info ]; then
-      coveralls-lcov --repo-token ${COVERALLS_TOKEN} build/coverage.info;
+  - if [ -f coverage.info ]; then
+      coveralls-lcov --repo-token=${COVERALLS_TOKEN} coverage.info;
     fi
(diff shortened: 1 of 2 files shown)


lexs:
@benjchristensen Coveralls is throwing this error now:
```
$ if [ -f build/coverage.info ]; then coveralls-lcov --repo-token=${COVERALLS_TOKEN} build/coverage.info; fi
#<Net::HTTPUnprocessableEntity 422 Unprocessable Entity readbody=true>
{"message":"Couldn't find a repository matching this job.","error":true}
```
Does the token no longer match or is there some setting than requires updating?

lexs:
https://github.com/lemurheavy/coveralls-public/issues/984 suggests this might only happen on PRs, let's try merging this as we at least now run coveralls.

yschimke:
@lexs should the build script take any of that into account?

The travis rsocket-java command is different for master, PRs, and tagged releases etc

https://github.com/rsocket/rsocket-java/blob/1.0.x/gradle/buildViaTravis.sh

.travis.yml
```diff
@@ -118,6 +118,6 @@ script:
 
 after_success:
     # Upload to coveralls.
-  - if [ -f build/coverage.info ]; then
-      coveralls-lcov --repo-token ${COVERALLS_TOKEN} build/coverage.info;
+  - if [ -f coverage.info ]; then
```
vitaut: This doesn't look right considering that we build and run everything except for tck in the build dir.

lexs: It seems like `after_success` runs in whatever directory `script` ended in, which now is `build`. Coveralls is working again now.

lexs:
Coveralls is now running again: https://coveralls.io/repos/159872/builds

PR #563, opened 2017-07-05
Ignore further temporary file types

phoad:


.gitignore
```diff
@@ -62,12 +62,14 @@ install_manifest.txt
...
 *.swp
 *.un~
+*.*~
 
 # Emacs files:
 # VCS comparison files:
 *.~*~
 # Autosave files
 .#*
+#*#
```
alexmalyshev: I get the other two changes, but isn't this just a comment as it begins with `#`?  Or am I not reading something correctly.

phoad: [#]*[#]
It should have been this format, to ignore temporary files with this name #filename#

PR #564, opened 2017-07-05
Slightly saner {Flowable,Observable}Operator shutdown semantics

alexmalyshev:
An operator is between an upstream Flowable and a downstream
subscriber/subscription.  The stream of signals it receives from one is not
synchronized with the other.  e.g. We can't assume that we still have a
downstream subscriber when FlowableOperator::onComplete() is called from
the upstream as the subscriber might have sent a cancel just before.

We consider an operator "terminated" if it has lost its references to its
upstream and downstream.  A terminated operator can still have methods called
upon it, so we need to make sure we don't send any signals from one side to the
other.

Enables the Observable.SimpleTake test, which previously hit ASAN errors because
of this bad behavior.


yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -55,48 +55,110 @@ class FlowableOperator : public Flowable<D> {
...
+      upstream_.reset();
+
+      if (auto subscriber = std::move(subscriber_)) {
+        subscriber->onComplete();
+      }
+
+      Refcounted::decRef(*this);
+    }
+
+    void onError(std::exception_ptr eptr) override {
```
phoad: This method looks like the same with terminateErr method.

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -55,48 +55,110 @@ class FlowableOperator : public Flowable<D> {
...
 
-    void onComplete() override {
-      assert(upstream_ && "subscription was already terminated");
-      upstream_.reset(); // breaking the cycle
-      subscriber_->onComplete();
-      Refcounted::decRef(*this);
+    // Subscription.
+
+    void request(int64_t delta) override {
+      if (upstream_) {
```
phoad: Should we record the request and do the request when the upstream_ becomes available? Or should we log/error in here stating that there is no upstream?

alexmalyshev: I don't believe so, because if `upstream_ == nullptr` then we've terminated this operator.  Any requests still flowing in from downstream are "late" so to speak, and should be dropped safely.

yarpl/include/yarpl/flowable/FlowableOperator.h
alexmalyshev: It isn't.  `terminateErr()` cancels the upstream whereas this does not.  The different comes from which part of the pipeline we're getting a termination signal.  `onError()` means the upstream has sent the termination, so we only need to null out the `upstream_` reference.  `terminateErr()` is used when the operator itself wants to terminate, so we need to send a `cancel()` to the upstream (and an `onError()` to the downstream as well).

alexmalyshev:
> Good update. I would ask if we should favor simplicity over code duplication. It looks like all the terminate/terminateErr/cancel/onComplete/onError functions are almost the same.

I'll see if I can merge them a bit.

PR #565, opened 2017-07-06
Add trailing newlines to files missing them

alexmalyshev:
Somebody's editor doesn't add a trailing newline, which is making some linters
unhappy.


phoad:
I have also seen the failures. Thanks for fixing.

PR #568, opened 2017-07-06
Don't return a folly::Future from ConnectionAcceptor::start()

alexmalyshev:
Starting a ConnectionAcceptor should be synchronous.  Returning a folly::Future
was mostly a convenient way of wrapping up thrown exceptions so users would be
forced to handle them.  The reality ended up being every user just did .get() on
the future and rethrew any exceptions anyway.

### DIFF
--- a/src/transports/tcp/TcpConnectionAcceptor.cpp
+++ b/src/transports/tcp/TcpConnectionAcceptor.cpp
@@ -3,6 +3,7 @@
 #include "TcpConnectionAcceptor.h"
 
 #include <folly/ThreadName.h>
+#include <folly/futures/Future.h>
 #include <folly/io/async/ScopedEventBaseThread.h>
 
 #include "src/framing/FramedDuplexConnection.h"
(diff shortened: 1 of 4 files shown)


src/transports/tcp/TcpConnectionAcceptor.cpp
```diff
@@ -84,25 +83,29 @@ folly::Future<folly::Unit> TcpConnectionAcceptor::start(
...
+        }
+
+        serverSocket_->listen(options_.backlog);
+        serverSocket_->startAccepting();
+
+        for (auto& i : serverSocket_->getAddresses()) {
+          LOG(INFO) << "Listening on " << i.describe();
+        }
+      })
+      .get();
```
phoad: Can we use "wait" in here and do better logging?

Should we clear onAccept_ variable in case of failure, to be able to retry again?

alexmalyshev: I don't think there's anything meaningful to log here.  If we successfully start up then we can log that in `RSocketServer` if we care, and if there's an error then we want to throw it upwards and let whatever is above us on the stack handle it as they please.

Clearing `onAccept_` and retrying is interesting, but I can't see that being useful.  For most cases, throwing an exception and destroy everything on server failure is fine I think.

PR #569, opened 2017-07-07
Ensure server and client are in sync about resumability

somasun:
If the client is running in a resumable mode, the server should be aware of it.  This is especially important in RSocket v0.1, since the the KEEPALIVE frame serialization format depends on whether resumption is enabled or not.  If the client has resumption enabled and the server is not aware of it, then it could result in mangled KEEPALIVE frames.

### DIFF
--- a/src/statemachine/RSocketStateMachine.cpp
+++ b/src/statemachine/RSocketStateMachine.cpp
@@ -62,14 +62,15 @@ RSocketStateMachine::~RSocketStateMachine() {
 
 void RSocketStateMachine::setResumable(bool resumable) {
   debugCheckCorrectExecutor();
-  DCHECK(isDisconnectedOrClosed()); // we allow to set this flag before we are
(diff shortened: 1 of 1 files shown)


src/statemachine/RSocketStateMachine.cpp
```diff
@@ -62,14 +62,15 @@ RSocketStateMachine::~RSocketStateMachine() {
 
 void RSocketStateMachine::setResumable(bool resumable) {
   debugCheckCorrectExecutor();
-  DCHECK(isDisconnectedOrClosed()); // we allow to set this flag before we are
-  // connected
+  DCHECK(isDisconnectedOrClosed()); // we allow to set this flag only before
```
alexmalyshev: Can you line up these `//` comments?  Maybe just move them above `DCHECK(isDisconnectedOrClosed())` and align with that.

PR #570, opened 2017-07-07
Delete TcpConnectionFactory::createDuplexConnectionFromSocket()

alexmalyshev:
It's a very verbose way of doing std::make_unique\<TcpDuplexConnection\>(...)

### DIFF
--- a/src/transports/tcp/TcpConnectionFactory.cpp
+++ b/src/transports/tcp/TcpConnectionFactory.cpp
@@ -80,12 +80,4 @@ void TcpConnectionFactory::connect(OnDuplexConnectionConnect cb) {
       });
 }
 
-std::unique_ptr<DuplexConnection>
-TcpConnectionFactory::createDuplexConnectionFromSocket(
-    folly::AsyncSocket::UniquePtr socket,
-    folly::EventBase& eventBase,
-    std::shared_ptr<RSocketStats> stats) {
-  return std::make_unique<TcpDuplexConnection>(std::move(socket), eventBase, std::move(stats));
-}
-
 } // namespace rsocket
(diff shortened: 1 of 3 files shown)


src/transports/tcp/TcpConnectionFactory.cpp
```diff
@@ -80,12 +80,4 @@ void TcpConnectionFactory::connect(OnDuplexConnectionConnect cb) {
       });
 }
 
-std::unique_ptr<DuplexConnection>
-TcpConnectionFactory::createDuplexConnectionFromSocket(
```
phoad: I have deleted this function in my previous PR and @lehecka has informed that this function needs to stay. I am not sure if the concern of him still exists.

lehecka:
TcpDuplexConnection is supposed to be an internal type, not part of the public API. The idea is that TcpConnectionFactory::createDuplexConnectionFromSocket is returning DuplexConnection base type and hiding the way how is the instance created.

alexmalyshev:
> TcpDuplexConnection is supposed to be an internal type, not part of the public API.

I'd argue that if it's being returned at all to the user's code and they have to use it, then it's part of the public API.  No point with bothering with passthrough functions.

lehecka:
To be clear what I meant:
TcpDuplexConnection is not returned to the application code. It is returned as DuplexConnection (the base type).
There is nothing interesting TcpDuplexConnection so I would rather keep it "invisible" and not allow application code to depend on this type.

alexmalyshev:
Fair enough.  I didn't realize it was actually being returned as a `unique_ptr<DuplexConnection>`.

PR #571, opened 2017-07-09
Add unit tests for 'DuplexConnection to use yarpl Subscriber/Subscription'

phoad:
Added three unit tests
 - Call getOutput/setInput functions of DuplexConnection multiple times and observe that the connection is still useful via sending new messages over these new Subscriber/Subscriptions.
 - Close input / output of the connection separately and observe that closing one direction does not effect the other.
 - Close input / output and observe that the connection is still open.


test/transport/TcpDuplexConnectionTest.cpp
```diff
@@ -0,0 +1,96 @@
...
+    clientPromise.setValue();
+  });
+
+  serverPromise.getFuture().wait();
+  clientPromise.getFuture().wait();
+  return std::make_pair(std::move(server), std::move(client));
+}
+
+TEST(TcpDuplexConnection, MultipleSetInputGetOutputCalls) {
+  std::unique_ptr<DuplexConnection> serverConnection, clientConnection;
```
lehecka: "Do not declare more than one variable on a line (this helps accomplish the other guidelines, promotes commenting, is good for tagging)."
https://our.intern.facebook.com/intern/wiki/CppStyle

PR #572, opened 2017-07-10
Split FrameTransport::close() into ::close() and ::closeErr()

alexmalyshev:
Use explicit methods for onComplete() or onError() signals.  Add tests.

Adds mocks for DuplexConnection and FrameProcessor.


src/framing/FrameTransport.h
```diff
@@ -36,7 +36,14 @@ class FrameTransport :
...
   virtual void outputFrameOrEnqueue(std::unique_ptr<folly::IOBuf> frame);
-  virtual void close(folly::exception_wrapper ex);
+
+  /// Cancel the input, complete the output, and close the underlying
+  /// connection.
+  void close();
+
+  /// Cancel the input, error the output, and close the underlying connection.
+  /// This must be closed with a non-empty exception_wrapper.
+  void closeErr(folly::exception_wrapper);
```
lexs: can we call this `closeWithError()`? Clarifies the intent a bit.

src/framing/FrameTransport.cpp
```diff
@@ -69,10 +69,23 @@ void FrameTransport::setFrameProcessor(
...
+
+void FrameTransport::closeErr(folly::exception_wrapper ew) {
+  if (!ew) {
+    VLOG(1) << "FrameTransport::closeErr() called with empty exception";
+    ew = std::runtime_error("Undefined error");
+  }
+  closeImpl<true>(std::move(ew));
+}
+
+template <bool isErr>
```
lexs: I think this could use a bool arg instead, I don't see the reason for this to use a template. Or just use `folly::Optional`.

alexmalyshev: Yeah it should use a runtime value.  I'm not sure why I used a template in the first place...

src/framing/FrameTransport.h
alexmalyshev: I kind of like `closeErr()`, but I'm with `closeWithError()` too.

alexmalyshev:
> Something we've run in to again and again is not sending an error frame when we close in error. Is this something we could do in this diff too?

We could, but I'm not yet sure that all places where we're closing the `FrameTransport` means we want to be sending an error frame (maybe we do, I'm just not sure yet :) ).  I'll play around with sending an error frame and see what comes of it.

lexs:
Yeah, it's something we've been bitten by in prod but it's kinda tangential to this PR so let's get this in.

PR #576, opened 2017-07-11
expose isFramedConnection in RSocketServer::acceptConnection method

lehecka:
to replace https://github.com/rsocket/rsocket-cpp/pull/574

### DIFF
--- a/src/framing/FramedDuplexConnection.h
+++ b/src/framing/FramedDuplexConnection.h
@@ -25,6 +25,10 @@ class FramedDuplexConnection : public virtual DuplexConnection {
   void setInput(yarpl::Reference<yarpl::flowable::Subscriber<std::unique_ptr<folly::IOBuf>>>
                     framesSink) override;
 
+  bool isFramed() override {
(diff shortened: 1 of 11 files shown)


src/framing/FramedDuplexConnection.h
```diff
@@ -25,6 +25,10 @@ class FramedDuplexConnection : public virtual DuplexConnection {
   void setInput(yarpl::Reference<yarpl::flowable::Subscriber<std::unique_ptr<folly::IOBuf>>>
                     framesSink) override;
 
+  bool isFramed() override {
```
phoad: nit: checking if the input parameter "connection" of the constructor is not already a FramedDuplexConnection.

phoad:
I would like to thank for this update and for the suggestion of removing this bool isFramedConnection parameter. It has just helped me that one of the failing test, that took my two hours to figure out what's going on and could not figure out it, has just passed with this update!
Clean! Nice!
@lehecka @lexs @yschimke

PR #579, opened 2017-07-12
General FrameTransport cleanup

alexmalyshev:
* Remove virtual methods and make it final, there are no subclasses.

* Rename drainOutputFramesQueue() to drainWrites(), add a matching drainReads().
  Both of them take a lock_guard<> as a parameter, as all of their uses were
  already within the scope of one.

* Delete FrameTransport::duplexConnection().  It's not safe to be called in the
  general case as we can't control when the connection will be closed, and it's
  not necessary in tests if mocks are being used.

* Remove out-of-date TODOs about re-enabling VLOGs that may or may not be
  useful.

* Comments and a few tests.

### DIFF
--- a/src/framing/FrameTransport.cpp
+++ b/src/framing/FrameTransport.cpp
@@ -1,8 +1,13 @@
 // Copyright 2004-present Facebook. All Rights Reserved.
 
 #include "FrameTransport.h"
+
 #include <folly/ExceptionWrapper.h>
+#include <folly/io/IOBuf.h>
+#include <glog/logging.h>
+
 #include "src/DuplexConnection.h"
+#include "src/framing/FrameProcessor.h"
 
 namespace rsocket {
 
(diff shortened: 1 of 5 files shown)


src/framing/FrameTransport.cpp
```diff
@@ -175,62 +168,60 @@ void FrameTransport::request(int64_t n) noexcept {
...
-  // not been drained (e.g. we're looping in ::request).
-  // or we are disconnected
+  // We either have no allowance to perform the operation, or the queue has not
+  // been drained (e.g. we're looping in ::request), or we are disconnected.
   pendingWrites_.emplace_back(std::move(frame));
 }
 
-void FrameTransport::drainOutputFramesQueue() {
-  std::lock_guard<std::recursive_mutex> lock(mutex_);
+void FrameTransport::drainReads(const FrameTransport::Lock&) {
```
phoad: I liked it that we pass a Lock inside and force this method to be single threaded without locking the recursive_mutex again. 

But I wonder, if we give different locked mutexes to this function, then this function will not be single threaded. So it is not the same as the original code. Namely, now we use a single mutex object and it is not a problem, but might passing this parameter in or writing a comment have the same result?

Do you think with your current update, is it possible to change recursive_mutex to non-recursive mutex?

src/framing/FrameTransport.cpp
```diff
@@ -151,22 +144,22 @@ void FrameTransport::terminateFrameProcessor(folly::exception_wrapper ex) {
...
   VLOG(6) << "onComplete";
-  terminateFrameProcessor(folly::exception_wrapper());
+  terminateProcessor(folly::exception_wrapper());
 }
 
-void FrameTransport::onError(std::exception_ptr ex) noexcept {
+void FrameTransport::onError(std::exception_ptr ex) {
   VLOG(6) << "onError" << folly::exceptionStr(ex);
-  terminateFrameProcessor(folly::exception_wrapper(std::move(ex)));
+  terminateProcessor(folly::exception_wrapper(std::move(ex)));
```
alexmalyshev: Nice catch.  I can repro this in a test, so I'll fix it here.

src/framing/FrameTransport.cpp
alexmalyshev: > But I wonder, if we give different locked mutexes to this function, then this function will not be single threaded. So it is not the same as the original code. Namely, now we use a single mutex object and it is not a problem, but might passing this parameter in or writing a comment have the same result?

I don't think that's an issue, as there's only one mutex on the `FrameTransport` to lock.  If somebody makes another `std::recursive_mutex` on the stack somewhere then that's their problem :).

> Do you think with your current update, is it possible to change recursive_mutex to non-recursive mutex?

No, I don't think so.  The critical sections aren't just re-entered as `FrameTransport -> FrameTransport`, I think they're `FrameTransport -> FrameProcessor -> FrameTransport`.

src/framing/FrameTransport.cpp
```diff
@@ -149,9 +149,15 @@ void FrameTransport::onComplete() {
   terminateProcessor(folly::exception_wrapper());
 }
 
-void FrameTransport::onError(std::exception_ptr ex) {
-  VLOG(6) << "onError" << folly::exceptionStr(ex);
-  terminateProcessor(folly::exception_wrapper(std::move(ex)));
+void FrameTransport::onError(std::exception_ptr eptr) {
+  VLOG(6) << "onError" << folly::exceptionStr(eptr);
+
+  try {
```
phoad: I believe the input "eptr" should never be nullptr, but in case, should we check if eptr is null?

http://en.cppreference.com/w/cpp/error/rethrow_exception

alexmalyshev: I'd rather leave it assuming that `eptr` is non-null, the same way all other `onError()` methods work right now.

PR #581, opened 2017-07-13
Add rules for installing rsocket

alexmalyshev:
Installing rsocket and yarpl headers, along with libReactiveSocket.a and libyarpl.a


CMakeLists.txt
```diff
@@ -347,8 +347,17 @@ target_compile_options(
...
+# Install libReactiveSocket.a and yarpl headers normally.
+install(TARGETS ReactiveSocket DESTINATION lib)
+install(DIRECTORY yarpl/include/yarpl DESTINATION include FILES_MATCHING PATTERN "*.h")
+
+# RSocket headers get installed in $INSTALL_DIR/src/Foo.h initially, because
+# they use #includes of the form "src/RSocketServer.h".  We also need them to be
+# $INSTALL_DIR/rsocket/Foo.h.
+install(DIRECTORY src DESTINATION include FILES_MATCHING PATTERN "*.h")
+install(CODE "execute_process(COMMAND rm -r ${CMAKE_INSTALL_PREFIX}/include/rsocket)")
+install(CODE "execute_process(COMMAND cp -r ${CMAKE_INSTALL_PREFIX}/include/src ${CMAKE_INSTALL_PREFIX}/include/rsocket)")
```
phoad: Is it possible to create a symbolic link instead of copying?

alexmalyshev: Probably, but now that I've tried this hacky `src/` copy I realize it doesn't quite work... It's far too easy to transitively include both `"src/RSocketServer.h"` and `"rsocket/RSocketServer.h"`, leading to duplicate definition compiler errors.

I think I'm going to try to rename `src/` to `rsocket/` and then fix up all the #includes in one go before I ship this.

PR #582, opened 2017-07-13
Add test for server binding to an invalid port

alexmalyshev:
TcpConnectionAcceptor::start() should throw when it fails to bind to the port.

Note - I don't actually want to land this diff as is.  I'm looking to see if
anyone can tell me why the changes to TcpConnectionAcceptor::start() are
necessary for the test to pass in Linux/ASAN/gcc builds.

### DIFF
--- a/src/transports/tcp/TcpConnectionAcceptor.cpp
+++ b/src/transports/tcp/TcpConnectionAcceptor.cpp
@@ -84,7 +84,7 @@ void TcpConnectionAcceptor::start(OnDuplexConnectionAccept onAccept) {
 
   // The AsyncServerSocket needs to be accessed from the listener thread only.
   // This will propagate out any exceptions the listener throws.
-  folly::via(
+  auto future = folly::via(
       serverThread_->getEventBase(),
       [this] {
         folly::SocketAddress addr;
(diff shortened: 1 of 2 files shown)


src/transports/tcp/TcpConnectionAcceptor.cpp
```diff
@@ -103,8 +103,12 @@ void TcpConnectionAcceptor::start(OnDuplexConnectionAccept onAccept) {
...
           LOG(INFO) << "Listening on " << i.describe();
         }
-      })
-      .get();
+      });
+
+  // XXX: Ideally we'd just do `future.get()`.  I can't explain why, but that
+  // consistently hits an asan-stack-overflow error in the exceptional case when
+  // built with gcc on Linux.  Doing `wait().getTry()` magically works.
+  future.wait().getTry().throwIfFailed();
```
phoad: The old version of "fbcode/rsocket/transports/http2/HttpConnectionAcceptor.cpp" was having such a comment for the Future that's returned from server.start(). Though as we now don't return any Future, this code is deleted.

    // Need to synchronize on finishing HTTPServer::start() before ever calling
    // HTTPServer::stop().  Wait on either success or error callback to fire.  We
    // need to make a new future after waiting on the first one, otherwise the
    // calling .then() or .onError() fails with `setCallback called twice`.

    auto asTry = std::move(promise.getFuture().wait().getTry());
    return folly::makeFuture<folly::Unit>(std::move(asTry));

I am not sure if they are related but, as you see, they first wait and then do getTry and generate another future from the initial one.

alexmalyshev: I wrote that comment.  I'm not convinced it's the same issue, as I wasn't hitting an ASAN-specific error then.

For that case I'm pretty sure I had messed up a test by forgetting to call `.get()` on the result of `HttpConnectionAcceptor::start()`, and adding a `.get()` equivalent inside of it made things work.  I'm quite glad that `asTry` code is gone.

phoad:
Let's ask it to Learning C++ group?

alexmalyshev:
Closing out as I'm not looking at this right now.

PR #584, opened 2017-07-14
Split out Frame{Flags,Type} and ErrorCode out from Frame.h

alexmalyshev:
Trying to reduce the surface area of Frame.h and internal/Common.h.

Cut out some duplicate definitions of StreamId as well.  It'll stay defined in
internal/Common.h for now.


rsocket/framing/FrameType.cpp
```diff
@@ -0,0 +1,57 @@
...
+    case FrameType::RESUME:
+      return "RESUME";
+    case FrameType::RESUME_OK:
+      return "RESUME_OK";
+    case FrameType::EXT:
+      return "EXT";
+    default:
+      break;
+  }
+  LOG(FATAL) << "Unknown FrameType[" << static_cast<int>(type) << "]";
```
lexs: This method shouldn't fatal (not sure if you just copied this or not but lets fix it). Return unknown instead like for error codes.

alexmalyshev: My thinking was that we should never get in a situation where we're getting an invalid `FrameType`, as all cases of the enum class are well-defined (unlike `FrameFlags` which is an arbitrary bitset and `ErrorCode` which is allowed to have user-defined values).

That said, I'm fine with changing it to not fatal.  I'll do that.

lehecka:
I dont have anything against this change. What I am missing is more detailed motivation of this PR in the description.
Frame.h now includes all the new .h files so there is no change of surface there.
In Common.h we saved the declaration of enum class FrameType : uint8_t;

alexmalyshev:
It's to break apart the smaller frame enums into their own modules.  For modules like `rsocket/Payload.{cpp,h}` that only use one of them (`FrameFlags`), we can just include the single header we need instead of all of `Frame.h`.  I also don't like forward declaring enum classes because 1) the size of the enum always has to be specified which defeats the purpose of forward declaring a bit 2) they're small enough (in terms of code) that it's fine to just include their headers.

`Frame.h` has to include them yes, because it actually uses all of them, but I'm aiming to cut down `Common.h` because it it's just a monolithic header filled with all sorts of decls/defs.

PR #586, opened 2017-07-15
Rename src/ to rsocket/

alexmalyshev:
And fix up all #includes with sed.  This is on the path to installing rsocket
headers to /usr/local/include/rsocket.


benchmarks/RequestResponseLatency.cpp
```diff
@@ -2,13 +2,15 @@
 
 #include <benchmark/benchmark.h>
 #include <folly/io/async/ScopedEventBaseThread.h>
-#include <src/transports/tcp/TcpConnectionAcceptor.h>
+
```
phoad: Looks like 'sed' has left empty lines.

alexmalyshev: Ah right, I fixed up the `#include <src/...>` cases by hand as there were only a few.

PR #588, opened 2017-07-18
[ConnectionFactory] Return future instead of taking callback

somasun:
Makes error handling easier.  Fits with the result of the planned API changes

### DIFF
--- a/rsocket/ConnectionFactory.h
+++ b/rsocket/ConnectionFactory.h
@@ -14,8 +14,8 @@ class EventBase;
 namespace rsocket {
 
 struct ConnectionResult {
-  std::unique_ptr<rsocket::DuplexConnection> connection_;
-  folly::EventBase &evb_;
+  std::unique_ptr<rsocket::DuplexConnection> connection;
+  folly::EventBase &evb;
 };
 
 /**
--- a/rsocket/ConnectionFactory.h
+++ b/rsocket/ConnectionFactory.h
@@ -3,6 +3,8 @@
 #pragma once
 
 #include <folly/Function.h>
+#include <folly/futures/Future.h>
+
 #include "rsocket/DuplexConnection.h"
 
 namespace folly {
@@ -11,9 +13,10 @@ class EventBase;
 
 namespace rsocket {
 
-using OnDuplexConnectionConnect = folly::Function<void(
-    std::unique_ptr<rsocket::DuplexConnection>,
-    folly::EventBase&)>;
+struct ConnectionResult {
+  std::unique_ptr<rsocket::DuplexConnection> connection_;
+  folly::EventBase &evb_;
+};
 
 /**
  * Common interface for a client to create connections and turn them into
@@ -27,7 +30,9 @@ using OnDuplexConnectionConnect = folly::Function<void(
 class ConnectionFactory {
  public:
   ConnectionFactory() = default;
+
   virtual ~ConnectionFactory() = default;
+
   ConnectionFactory(const ConnectionFactory&) = delete; // copy
   ConnectionFactory(ConnectionFactory&&) = delete; // move
   ConnectionFactory& operator=(const ConnectionFactory&) = delete; // copy
@@ -42,6 +47,6 @@ class ConnectionFactory {
    *
    * Resource creation depends on the particular implementation.
    */
-  virtual void connect(OnDuplexConnectionConnect onConnect) = 0;
+  virtual folly::Future<ConnectionResult> connect() = 0;
 };
 } // namespace rsocket
--- a/rsocket/RSocketClient.cpp
+++ b/rsocket/RSocketClient.cpp
@@ -32,33 +32,26 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
     std::shared_ptr<RSocketNetworkStats> networkStats) {
   VLOG(2) << "Starting connection";
 
-  folly::Promise<std::unique_ptr<RSocketRequester>> promise;
-  auto future = promise.getFuture();
-
-  connectionFactory_->connect()
-      .then([this,
-             setupParameters = std::move(setupParameters),
-             responder = std::move(responder),
-             keepaliveTimer = std::move(keepaliveTimer),
-             stats = std::move(stats),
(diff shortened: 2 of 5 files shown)


rsocket/ConnectionFactory.h
```diff
@@ -11,9 +13,10 @@ class EventBase;
 
 namespace rsocket {
 
-using OnDuplexConnectionConnect = folly::Function<void(
-    std::unique_ptr<rsocket::DuplexConnection>,
-    folly::EventBase&)>;
+struct ConnectionResult {
+  std::unique_ptr<rsocket::DuplexConnection> connection_;
```
lehecka: for public members, I don't think we should be using _ postfix.

rsocket/RSocketClient.cpp
```diff
@@ -35,28 +35,28 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
...
+            std::move(keepaliveTimer),
+            std::move(stats),
+            std::move(networkStats));
+        promise.setValue(std::move(rsocket));
+      })
+      .onError([promise = std::move(promise)](folly::exception_wrapper ex) mutable {
+        promise.setException(ex);
+      });
 
   return future;
```
lehecka: you don't have to allocate promise anymore. you can just 
`return connectionFactory_->connect().then(...)`

test/transport/TcpDuplexConnectionTest.cpp
```diff
@@ -43,11 +43,11 @@ makeSingleClientServer(
...
-  client->connect(
+  client->connect().then(
       [&clientPromise, &clientConnection, &clientEvb](
-          std::unique_ptr<DuplexConnection> connection, EventBase& eventBase) {
-        clientConnection = std::move(connection);
-        *clientEvb = &eventBase;
+          ConnectionResult connResult) {
+        clientConnection = std::move(connResult.connection_);
+        *clientEvb = &connResult.evb_;
         clientPromise.setValue();
```
lehecka: I don't think you need the clientPromise anymore..

rsocket/RSocketClient.cpp
```diff
@@ -35,28 +35,44 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
...
+        setupParameters = std::move(setupParameters),
+        responder = std::move(responder),
+        keepaliveTimer = std::move(keepaliveTimer),
+        stats = std::move(stats),
+        networkStats = std::move(networkStats),
+        promise = std::move(promise)
+      ](ConnectionResult connResult) mutable {
+        VLOG(3) << "onConnect received DuplexConnection";
+        // Ensure we create the RSocketStateMachine in the right evb.
+        connResult.evb.runImmediatelyOrRunInEventBaseThreadAndWait([
```
lexs: Please don't block here. On mobile this eventbase is shared meaning you'll potentially wait for other requests (or block them).

rsocket/RSocketClient.cpp
```diff
@@ -35,28 +35,44 @@ folly::Future<std::unique_ptr<RSocketRequester>> RSocketClient::connect(
...
+          auto rsocket = fromConnection(
+              std::move(connResult.connection),
+              connResult.evb,
+              std::move(setupParameters),
+              std::move(responder),
+              std::move(keepaliveTimer),
+              std::move(stats),
+              std::move(networkStats));
+
+          promise.setValue(std::move(rsocket));
```
lexs: As @lehecka mentioned you can just chain these instead.

rsocket/RSocketClient.cpp
lexs: Instead of this you should be able to ensure the `then()` clause happens on the correct eventbase using `via()` on the caller side.

rsocket/transports/tcp/TcpConnectionFactory.cpp
```diff
@@ -47,19 +48,20 @@ class ConnectCallback : public folly::AsyncSocket::ConnectCallback {
         std::move(socket_), RSocketStats::noop());
     auto evb = folly::EventBaseManager::get()->getExistingEventBase();
     CHECK(evb);
-    onConnect_(std::move(connection), *evb);
+    connectionResult_.setValue(ConnectionResult{std::move(connection), *evb});
   }
 
-  void connectErr(const folly::AsyncSocketException& ex) noexcept override {
+  void connectErr(const folly::AsyncSocketException &ex) noexcept override {
```
lexs: you reformatted this (and `ConnectionResult` has weird formatting), why?

rsocket/RSocketClient.cpp
somasun: Right.  I could do connect().via(evb).then().  But the evb is available only in the lambda passed to then() (not before it), so it makes it tricky.   @lehecka @lexs - any suggestions of cleaner way of doing this ?

rsocket/RSocketClient.cpp
somasun: Yes chaining is possible.  But I cannot ensure then() executes in the correct eventBase (even with via()) as mentioned in the previous comment.

rsocket/RSocketClient.cpp
lexs: I believe the future will actually resolve on the same thread you call `Promise::setValue()` on, which should be the correct eventbase. So I think this already works, just add a CHECK for being on the current eventbase in the callback?

Otherwise what you can do is in the .then() do
```
return Future::via(evb, [] {
  // .. this will run on evb.
});
```

rsocket/RSocketClient.cpp
lexs: Why doesn't via() work?

rsocket/RSocketClient.cpp
somasun: "I believe the future will actually resolve on the same thread you call Promise::setValue()"  -> The lambda inside then() could run in any thread (the one which calls Promise::setValue() or the one which call .then()).  There is no guarantee unless you can via()

The evb is passed as argument to the lambda in then(), so the evb wont be available to schedule the lambda.

somasun:
Will get back to this after I finish other API changes.

PR #591, opened 2017-07-18
RSocket Client API

somasun:
Changing the RSocketClient API to be more intuitive.   

The only way to create clients is using RSocket class.  The clients can be created in two modes - regular-connected-client or cold-resumed-client.  The API forces the users to pass the appropriate parameters for both the constructors.  The only public method which RSocketClient exposes is `resume()`.  `client->resume()` would be a warm-resumption attempt.  All the methods return futures.

```
RSocket::createConnectedClient(...)
      .then([&client](std::unique_ptr<RSocketClient> cl) mutable {
        client = std::move(cl);  // store a reference to the client
        client->getRequester()
            ->requestStream(Payload("BM_Stream"))
            ->subscribe(...);
      });
```

`RSocket::createResumedClient()` has a similar usage.

TODO:
- Add more convenience constructors in RSocket to cover common cases.   All constructors would return `Future<RSocketClient>` and they would create a fresh-connected-client or a cold-resumed-client.
- Resumption part in this PR is only about the API.  It will updated and tested once the server side changes are also made.
- Error propagation from the ConnectionFactory to the user has to be implemented.  This would require probably a change in the ConnectionFactory::connect() method and RSocketClient::connect() method.

### DIFF
--- a/benchmarks/StreamThroughput.cpp
+++ b/benchmarks/StreamThroughput.cpp
@@ -186,20 +186,17 @@ BENCHMARK_DEFINE_F(BM_RsFixture, BM_Stream_Throughput)
 
   std::shared_ptr<RSocketClient> client;
 
-  RSocket::createConnectedClient(
-      std::make_unique<TcpConnectionFactory>(std::move(address)))
-      .then([&client, s = std::move(s) ](
-          std::shared_ptr<RSocketClient> cl) mutable {
-        LOG(INFO) << "Connected";
-        client = std::move(cl);
-        client->getRequester()
-            ->requestStream(Payload("BM_Stream"))
-            ->subscribe(std::move(s));
-      })
-      .onError([](folly::exception_wrapper ex) {
-        LOG(INFO) << "Exception received " << ex;
-        return;
-      });
+  try {
+    client = RSocket::createConnectedClient(
+                 std::make_unique<TcpConnectionFactory>(std::move(address)))
+                 .get();
+    client->getRequester()
+        ->requestStream(Payload("BM_Stream"))
+        ->subscribe(std::move(s));
+  } catch (const std::exception& ex) {
+    LOG(INFO) << "Exception received " << ex;
+    return;
+  }
 
   while (state.KeepRunning()) {
     std::this_thread::yield();
--- a/benchmarks/StreamThroughput.cpp
+++ b/benchmarks/StreamThroughput.cpp
@@ -184,12 +184,12 @@ BENCHMARK_DEFINE_F(BM_RsFixture, BM_Stream_Throughput)
 
   auto s = make_ref<BM_Subscriber>(state.range(0));
 
-  std::unique_ptr<RSocketClient> client;
+  std::shared_ptr<RSocketClient> client;
 
   RSocket::createConnectedClient(
       std::make_unique<TcpConnectionFactory>(std::move(address)))
       .then([&client, s = std::move(s) ](
-          std::unique_ptr<RSocketClient> cl) mutable {
+          std::shared_ptr<RSocketClient> cl) mutable {
         LOG(INFO) << "Connected";
         client = std::move(cl);
         client->getRequester()
--- a/benchmarks/StreamThroughput.cpp
+++ b/benchmarks/StreamThroughput.cpp
@@ -182,14 +182,20 @@ BENCHMARK_DEFINE_F(BM_RsFixture, BM_Stream_Throughput)
   folly::SocketAddress address;
   address.setFromHostPort(host_, port_);
 
-  auto clientRs = RSocket::createClient(
-      std::make_unique<TcpConnectionFactory>(std::move(address)));
-
   auto s = make_ref<BM_Subscriber>(state.range(0));
 
-  clientRs->connect().then([s](std::shared_ptr<RSocketRequester> rs) {
-    rs->requestStream(Payload("BM_Stream"))->subscribe(std::move(s));
-  });
+  std::unique_ptr<RSocketClient> client;
+
+  RSocket::createConnectedClient(
+      std::make_unique<TcpConnectionFactory>(std::move(address)))
+      .then([&client, s = std::move(s) ](
+          std::unique_ptr<RSocketClient> cl) mutable {
+        LOG(INFO) << "Connected";
+        client = std::move(cl);
+        client->getRequester()
+            ->requestStream(Payload("BM_Stream"))
+            ->subscribe(std::move(s));
+      });
 
   while (state.KeepRunning()) {
     std::this_thread::yield();
--- a/benchmarks/StreamThroughput.cpp
+++ b/benchmarks/StreamThroughput.cpp
@@ -195,6 +195,10 @@ BENCHMARK_DEFINE_F(BM_RsFixture, BM_Stream_Throughput)
         client->getRequester()
             ->requestStream(Payload("BM_Stream"))
             ->subscribe(std::move(s));
+      })
+      .onError([](folly::exception_wrapper ex) {
+        LOG(INFO) << "Exception received " << ex;
+        return;
       });
 
   while (state.KeepRunning()) {
--- a/examples/channel-hello-world/ChannelHelloWorld_Client.cpp
+++ b/examples/channel-hello-world/ChannelHelloWorld_Client.cpp
@@ -27,11 +27,11 @@ int main(int argc, char* argv[]) {
   folly::SocketAddress address;
   address.setFromHostPort(FLAGS_host, FLAGS_port);
 
-  std::unique_ptr<RSocketClient> client;
+  std::shared_ptr<RSocketClient> client;
 
   RSocket::createConnectedClient(
       std::make_unique<TcpConnectionFactory>(std::move(address)))
-      .then([&client](std::unique_ptr<RSocketClient> cl) mutable {
+      .then([&client](std::shared_ptr<RSocketClient> cl) mutable {
         client = std::move(cl);
         client->getRequester()
             ->requestChannel(Flowables::justN({"initialPayload", "Bob", "Jane"})
--- a/examples/channel-hello-world/ChannelHelloWorld_Client.cpp
+++ b/examples/channel-hello-world/ChannelHelloWorld_Client.cpp
@@ -27,20 +27,24 @@ int main(int argc, char* argv[]) {
   folly::SocketAddress address;
   address.setFromHostPort(FLAGS_host, FLAGS_port);
 
-  // create a client which can then make connections below
-  auto rsf = RSocket::createClient(
-      std::make_unique<TcpConnectionFactory>(std::move(address)));
+  std::unique_ptr<RSocketClient> client;
 
-  auto rs = rsf->connect().get();
-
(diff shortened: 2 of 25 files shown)


rsocket/RSocketClient.cpp
```diff
@@ -6,104 +6,147 @@
...
     std::shared_ptr<RSocketResponder> responder,
     std::unique_ptr<KeepaliveTimer> keepaliveTimer,
     std::shared_ptr<RSocketStats> stats,
-    std::shared_ptr<RSocketNetworkStats> networkStats) {
+    std::shared_ptr<RSocketNetworkStats> networkStats,
+    std::shared_ptr<ResumeManager> resumeManager,
+    std::shared_ptr<ColdResumeHandler> coldResumeHandler,
+    OnRSocketResume)
+    : connectionFactory_(std::move(connectionFactory)),
+      connectionManager_(std::make_unique<RSocketConnectionManager>()),
```
lexs: What's the point of this? It will only ever have one connection in it and doesn't seem like it adds anything.

lexs:
An alternative could be a static `connect()` that returns the client synchronously but also takes a callback for connection success/failure. That would work for us on mobile. However I would consider a builder API so don't have to repeat args, something like:
```
auto client = RSocket::clientBuilder(10 args).connect(callback);
// or
auto client = RSocket::clientBuilder(10 args).resume(callback);
```
This way I can write
```
auto builder =  RSocket::clientBuilder(10 args);
if (has_state) {
  client = builder.resume();
} else {
 client = builder.connect()
}
```

benchmarks/StreamThroughput.cpp
```diff
@@ -182,14 +182,20 @@ BENCHMARK_DEFINE_F(BM_RsFixture, BM_Stream_Throughput)
...
-    rs->requestStream(Payload("BM_Stream"))->subscribe(std::move(s));
-  });
+  std::shared_ptr<RSocketClient> client;
+
+  RSocket::createConnectedClient(
+      std::make_unique<TcpConnectionFactory>(std::move(address)))
+      .then([&client, s = std::move(s) ](
+          std::shared_ptr<RSocketClient> cl) mutable {
+        LOG(INFO) << "Connected";
+        client = std::move(cl);
```
somasun: I agree.  But I think .then().onError() makes the flow clearer.  And easy handling of error conditions.  But no strong preference either way.

rsocket/RSocket.cpp
```diff
@@ -1,12 +1,55 @@
...
+    std::unique_ptr<ConnectionFactory> connectionFactory,
+    SetupParameters setupParameters,
+    std::shared_ptr<RSocketResponder> responder,
+    std::unique_ptr<KeepaliveTimer> keepaliveTimer,
+    std::shared_ptr<RSocketStats> stats,
+    std::shared_ptr<RSocketNetworkStats> networkStats,
+    std::shared_ptr<ResumeManager> resumeManager,
+    std::shared_ptr<ColdResumeHandler> coldResumeHandler,
+    OnRSocketResume) {
+  auto c = std::shared_ptr<RSocketClient>(new RSocketClient(
```
somasun: make_shared doesn't seem to work well with friend classes

rsocket/RSocket.cpp
```diff
@@ -1,12 +1,55 @@
...
+
+  return c->connect().then([c]() mutable { return c; });
+}
+
+folly::Future<std::shared_ptr<RSocketClient>> RSocket::createResumedClient(
+    std::unique_ptr<ConnectionFactory> connectionFactory,
+    SetupParameters setupParameters,
+    std::shared_ptr<ResumeManager> resumeManager,
+    std::shared_ptr<ColdResumeHandler> coldResumeHandler,
+    OnRSocketResume,
```
somasun: Here the resumption arguments come first.  An application which wants a resumed-client cares about these arguments more than the other arguments (they could possibly leave it to default values)

rsocket/RSocket.cpp
```diff
@@ -1,12 +1,55 @@
...
+    std::shared_ptr<RSocketResponder> responder,
+    std::unique_ptr<KeepaliveTimer> keepaliveTimer,
+    std::shared_ptr<RSocketStats> stats,
+    std::shared_ptr<RSocketNetworkStats> networkStats) {
+  auto c = std::shared_ptr<RSocketClient>(new RSocketClient(
```
somasun: ditto

rsocket/RSocketClient.cpp
somasun: ConnectionManager has a nice way of closing the RSocket connection and cleaning the state.  Just reusing its logic here, instead of replicating the code.

rsocket/RSocketClient.h
```diff
@@ -4,66 +4,83 @@
...
-      std::shared_ptr<RSocketResponder> responder = std::shared_ptr<RSocketResponder>(),
-      std::unique_ptr<KeepaliveTimer> keepaliveTimer = std::unique_ptr<KeepaliveTimer>(),
-      std::shared_ptr<RSocketStats> stats = std::shared_ptr<RSocketStats>(),
-      std::shared_ptr<RSocketNetworkStats> networkStats = std::shared_ptr<RSocketNetworkStats>());
+  RSocketClient& operator=(RSocketClient&&) = default; // move
+
+  friend class RSocket;
+
+  // Returns the RSocketRequester associated with the RSocketClient.
+  std::shared_ptr<RSocketRequester> getRequester() const;
```
somasun: The plan is to make it such that, if the application has either of Requester or RSocketClient, it should be fine.  The code to make the Requester sufficient to keep the RSocket state is not in yet.

rsocket/internal/ClientResumeStatusCallback.h
```diff
@@ -6,22 +6,29 @@
 
 namespace rsocket {
 
+// Thrown when an ERROR frame with CONNECTION_ERROR is received during
+// resuming operation.
+// TODO: in this case we should get the DuplexConnection back to
+// create a new instance of RS with it
+class ResumptionException : public std::runtime_error {
```
somasun: Agreed.  I will make those changes in a separate PR.  The focus is now mainly on the user-facing API.

rsocket/RSocket.cpp
lexs: that's true, consider adding a create() method to RSocketClient that creates it. `make_shared()` saves an allocation.

benchmarks/StreamThroughput.cpp
lexs: try catch is even easier and the examples becomes much shorter :)
```cpp
std::shared_ptr<RSocketClient> client;
try {
  client = RSocket::createConnectedClient(...).get();
  client->getRequester()
      ->requestStream(Payload("BM_Stream"))
      ->subscribe(std::move(s));
} catch (const std::exception& ex) {
  LOG(INFO) << "Exception received " << ex;
}
```
Actually looking at this I don't think you should catch the exception, doesn't seem like the test will terminate then :)

rsocket/RSocket.cpp
lexs: I think that will only cause confusion. If you want these methods to be convenient I'd recommend switching to a builder so we can have sensible default. I really find that subtly changing argument order between two almost identical methods super confusing.

rsocket/RSocketClient.cpp
lexs: ¯\_(ツ)_/¯ `ConnectionManager` seems like a complicated way to call `onClose()`, is the `onClosed()` callback actually async? Why do we need a baton for shutdown?

rsocket/internal/ClientResumeStatusCallback.h
lexs: But this is the user-facing API no? You need to use these types when you resume.

rsocket/RSocket.cpp
somasun: create() has to be a static method.  You cant call a private constructor from a static method.

rsocket/RSocket.cpp
somasun: cold-resumption is going to be an extreme case.  Most usage will be of createConnectedClient().  
And when there is a usage of createResumedClient(), we need the application to pass in the necessary args like resumeManager, coldResumeHandler.  These args are mandatory for cold-resumed-client, but not necessary for an ordinary-client.  Hence the different ordering.

And since the normal-constructor is going to be used way more than the cold-resumed-constructor, I don't see much confusion for most cases.

rsocket/RSocketClient.cpp
somasun: The Baton is used to ensure all sockets are closed properly before destroying connectionManager.  The onClosed() depends on the implementation.  ConnectionManager discussion is out of scope of this client-API PR.

rsocket/RSocket.cpp
lexs: I guess you run into the same problem. But honestly I'm fine with having a public constructor.

rsocket/RSocket.cpp
lexs: > And since the normal-constructor is going to be used way more than the cold-resumed-constructor, I don't see much confusion for most cases.

I don't see why this is true, our only usage today will call them equally often. And even when you don't have cold state you'll have to pass all the same args to `createConnectedClient()` as you then want to create that state. Switching argument order between two very similar methods will cause confusion, see PHP for example.

rsocket/RSocketClient.cpp
lexs: Which implementation? Afaik `onClosed()` will be called in the same stack that calls `close()`, is this not the case?

examples/channel-hello-world/ChannelHelloWorld_Client.cpp
```diff
@@ -27,20 +27,24 @@ int main(int argc, char* argv[]) {
...
-  rs->requestChannel(Flowables::justN({"initialPayload", "Bob", "Jane"})
-                         ->map([](std::string v) {
-                           std::cout << "Sending: " << v << std::endl;
-                           return Payload(v);
-                         }))
-      ->subscribe([](Payload p) {
-        std::cout << "Received: " << p.moveDataToString() << std::endl;
+  std::shared_ptr<RSocketClient> client;
+
+  RSocket::createConnectedClient(
```
lexs: you can change this and all below to the same pattern with `get()` you used above

rsocket/Exception.h
```diff
@@ -0,0 +1,21 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#pragma once
+
+#include <stdexcept>
+
+// Thrown when an ERROR frame with CONNECTION_ERROR is received during
+// resuming operation.
+// TODO: in this case we should get the DuplexConnection back to
```
lexs: What does this mean? How would that work?

somasun: This was a copy of a old TODO.  removed it

rsocket/RSocket.cpp
somasun: True.  But this is designed keeping in mind the wider anticipated of RSocket (not just the existing use-case).  The assumption is most users would not care about cold-resumption.

PR #592, opened 2017-07-18
Use glog's DCHECK in Flowable.h, fix release build

alexmalyshev:
assert() generally causes "unused variable" errors because it gets completely
erased in optimized builds.  Switch to using DCHECK which won't cause a compiler
error (presumably because it does the equivalent of `(void)variable;`
internally).

### DIFF
--- a/yarpl/CMakeLists.txt
+++ b/yarpl/CMakeLists.txt
@@ -29,7 +29,6 @@ endif()
 # Using NDEBUG in Release builds.
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -DNDEBUG")
 
-# Find glog and libraries specifically.
 find_path(GLOG_INCLUDE_DIR glog/logging.h)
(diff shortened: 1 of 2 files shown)


yarpl/CMakeLists.txt
```diff
@@ -26,23 +26,18 @@ if(APPLE)
...
 
-# Configuration for Debug build mode.
-#set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fsanitize=address")
-set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}")
-
+# Using NDEBUG in Release builds.
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -DNDEBUG")
 
-# Find glog and gflags libraries specifically
+# Find glog and libraries specifically.
```
somasun: typo

PR #594, opened 2017-07-19
Drop use of OSS macro

alexmalyshev:
It was added so we could build older and newer versions of gflags (where the
namespaces differed between google:: and gflags::), but we don't need that if we
just stick with folly::init().


somasun:
Travis has some pkg install errors - "Couldn't find any package by glob 'clang-4.0'".  Else the PR looks good.

PR #595, opened 2017-07-19
Documenting and testing SetupResumeAcceptor

alexmalyshev:
Documenting and testing SetupResumeAcceptor

* Tweaked OnSetup and OnResume functions to be folly::Function to keep us
  honest about moving them instead of copying.

* Forced use of folly::EventBase::runInEventBaseThread() on all calls to close() in 
  SetupResumeAcceptor::processFrame().

* Fixed any places where we were constructing a folly::exception_wrapper without
  a std::exception_ptr.

* Added MockDuplexConnection ctor that runs lambdas on input and output
  subscribers.

### DIFF
--- a/rsocket/internal/SetupResumeAcceptor.cpp
+++ b/rsocket/internal/SetupResumeAcceptor.cpp
@@ -3,16 +3,24 @@
 #include "rsocket/internal/SetupResumeAcceptor.h"
 
 #include <folly/ExceptionWrapper.h>
-#include <folly/io/async/EventBaseManager.h>
+#include <folly/io/async/EventBase.h>
 
 #include "rsocket/DuplexConnection.h"
-#include "rsocket/RSocketStats.h"
 #include "rsocket/framing/Frame.h"
 #include "rsocket/framing/FrameProcessor.h"
 #include "rsocket/framing/FrameSerializer.h"
 #include "rsocket/framing/FrameTransport.h"
 
 namespace rsocket {
+
+namespace {
+
+folly::exception_wrapper error(folly::StringPiece message) {
+  std::runtime_error exn{message.str()};
+  auto eptr = std::make_exception_ptr(exn);
+  return folly::exception_wrapper{std::move(eptr), exn};
+}
+}
 
 class OneFrameProcessor : public FrameProcessor {
  public:
@@ -31,18 +39,20 @@ class OneFrameProcessor : public FrameProcessor {
   }
 
   void processFrame(std::unique_ptr<folly::IOBuf> buf) override {
-    acceptor_.processFrame(std::move(transport_), std::move(buf),
-                           std::move(onSetup_), std::move(onResume_));
-    // no more code here as the instance might be gone by now
+    acceptor_.processFrame(
+        std::move(transport_),
+        std::move(buf),
+        std::move(onSetup_),
+        std::move(onResume_));
+    // No more code here as the instance might be gone by now.
   }
 
-  void onTerminal(folly::exception_wrapper ex) override {
+  void onTerminal(folly::exception_wrapper ew) override {
     onSetup_ = nullptr;
     onResume_ = nullptr;
 
-    acceptor_.closeAndRemoveConnection(
-        std::move(transport_), std::move(ex));
-    // no more code here as the instance might be gone by now
+    acceptor_.close(std::move(transport_), std::move(ew));
+    // No more code here as the instance might be gone by now.
   }
 
  private:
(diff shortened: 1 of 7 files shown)


rsocket/internal/SetupResumeAcceptor.cpp
```diff
@@ -70,120 +71,94 @@ SetupResumeAcceptor::~SetupResumeAcceptor() {
...
-    case FrameType::METADATA_PUSH:
-    case FrameType::REQUEST_CHANNEL:
-    case FrameType::REQUEST_FNF:
-    case FrameType::REQUEST_N:
-    case FrameType::REQUEST_RESPONSE:
-    case FrameType::REQUEST_STREAM:
-    case FrameType::RESERVED:
-    case FrameType::PAYLOAD:
-    case FrameType::RESUME_OK:
-    case FrameType::EXT:
```
lehecka: I would suggest to keep these cases in, as its required by the compiler warning in -Wextra mode.
I found 2 bugs in RSocket in the past in this.

rsocket/internal/SetupResumeAcceptor.cpp
```diff
@@ -197,61 +172,59 @@ void SetupResumeAcceptor::accept(
...
+  // Transport can receive frames right away.
   transport->setFrameProcessor(std::move(processor));
 }
 
-std::shared_ptr<FrameSerializer>
-SetupResumeAcceptor::getOrAutodetectFrameSerializer(
-    const folly::IOBuf& firstFrame) {
-  if (defaultFrameSerializer_) {
-    return defaultFrameSerializer_;
+std::shared_ptr<FrameSerializer> SetupResumeAcceptor::frameSerializer(
```
lehecka: I would suggest to add the word "detect" in the name of the method. 
frameSerializer() doesn't suggest its an "expensive method" doing some heuristic to return FrameSerializer.

rsocket/internal/SetupResumeAcceptor.cpp
alexmalyshev: They shouldn't be necessary, as there's a `default` label that captures all of them.  I don't remember a warning included in `-Wextra` that would trigger that, are you thinking of `-Weverything` instead?

Also note that there are no warnings in the builds despite us using `-Wall -Wextra -Werror` for code in `rsocket/`.

rsocket/internal/SetupResumeAcceptor.cpp
alexmalyshev: Fair enough, I can see how `frameSerializer()` might be construed as being a simple getter.  Can I convince you on 'createSerializer()`?  I like keeping method names short, and the `create` part suggests that there's some expensive work being done.  Plus the method has a comment saying that it'll do the autodetect logic.

lexs:
Nice! There used to be tests for this class but they were deleted in https://github.com/rsocket/rsocket-cpp/pull/445. Last version was https://github.com/rsocket/rsocket-cpp/blob/f154f0c1fde388c660c0f7ee3724bb954ad14c2b/test/ServerConnectionAcceptorTest.cpp. Could you add the ones that make sense back? They tested for some very real bugs that we saw in prod.

alexmalyshev:
Sure, I can take a look through them and add back what I can.

rsocket/internal/SetupResumeAcceptor.cpp
lexs: Given that we really only care about `SETUP` and `RESUME` here I think `default` makes sense.

alexmalyshev:
Sorry, I'm intentionally delaying this as @somasun convinced me the code in https://github.com/rsocket/rsocket-cpp/blob/master/rsocket/internal/SetupResumeAcceptor.cpp#L149 was necessary for correctness.  I want to try exposing that crash in the tests first before landing.

rsocket/internal/SetupResumeAcceptor.cpp
```diff
@@ -197,61 +172,59 @@ void SetupResumeAcceptor::accept(
...
   }
 
-  VLOG(2) << "detected protocol version" << serializer->protocolVersion();
+  VLOG(3) << "Detected protocol version " << serializer->protocolVersion();
   return std::move(serializer);
 }
 
-void SetupResumeAcceptor::closeAndRemoveConnection(
-    const yarpl::Reference<FrameTransport>& transport,
+void SetupResumeAcceptor::close(
```
alexmalyshev: Too long, often clips past 80 columns.  I think reading `close(transport, exn)` is pretty unambiguous about closing `transport`.

PR #601, opened 2017-07-21
RSocket Server API

somasun:
This PR is to change RSocket Server API to include all the features necessary for resumption.

The main class being introduced is `RSocketServiceHandler`.  Refer `rsocket/RSocketServiceHandler.h` for detailed documentation.  The server application has to implement this interface.  Only one method is necessary.  There is a convenience constructor in RSocketServer to do this easily.

The server-application can use RSocketException to propagate connect/resume messages all the way to the client.   See `ConditionalRequestHandling_Server.cpp` and `ConditionalRequestHandling_Client.cpp` for an example.

The plumbing required to get warm-resumption working is also part of the PR.  `examples/warm-resumption` contains a sample warm-resumption server/client.

TODO
- Rename `RSocketNetworkStats` to `RSocketConnectionEvents`
- Add warm-resumption tests which break the underlying transport and verifies successful resumption.  This is in addition to warm-resumption server/client in this PR.
- The warm-resumption works only for 1 server thread.  To support multiple threads we need to make use of SwappableEventBase and ensure the StateMachine gets moved to the worker-thread where the client is resuming.


rsocket/internal/SetupResumeAcceptor.cpp
```diff
@@ -112,8 +113,25 @@ void SetupResumeAcceptor::processFrame(
...
-      removeConnection(transport);
-      onSetup(std::move(transport), std::move(setupPayload));
+      try {
+        onSetup(transport, std::move(setupPayload));
+      } catch (RSocketException ex) {
+        VLOG(1) << "Rejected connection " << ex.what();
+        transport->outputFrameOrEnqueue(frameSerializer->serializeOut(
+            Frame_ERROR::rejectedSetup(ex.what())));
+        // FrameTransport is above us in the call stack (and FrameTransport is
+        // holding a recursive_mutex).  After this call FrameTransport goes out
```
lexs: Is anything trying to fix this? I think this should be a priority.

rsocket/RSocketServiceHandler.h
```diff
@@ -0,0 +1,97 @@
...
+      ResumeIdentificationToken) {}
+
+  // This method gets called when a client tries to resume. The application
+  // should return the corresponding RSocketServerState for the given resumeId
+  // if it wants to resume.  The application should return a RSocketException
+  // if it does not want to resume.  If the application returns a
+  // RSocketException, then an ERROR frame with code REJECTED_RESUME is sent to
+  // the client.  The exception message is sent as payload.
+  virtual folly::Expected<std::unique_ptr<RSocketServerState>, RSocketException>
+  onResume(ResumeIdentificationToken) {
```
lexs: You have nothing else to go on other than the `ResumeIdentificationToken`? How would we pipe metadata like user id through to here?

rsocket/RSocketServiceHandler.h
```diff
@@ -0,0 +1,97 @@
...
+ public:
+  virtual ~RSocketServiceHandler() = default;
+
+  // This method gets called for each client that connects to the server.  The
+  // application has to implement this method.  If the application does not
+  // want to accept this connection, it should return a RSocketException.  If
+  // the application returns a RSocketException, then an ERROR frame with code
+  // REJECTED_SETUP is sent to the client.  The exception message is sent as
+  // payload.
+  virtual folly::Expected<RSocketConnectionParams, RSocketException> onNewSetup(
```
somasun: The server needs the responder (and construct `RSocketStateMachine`) to be able to process subsequent frames from the client.  So this call has to be synchronous.  In most cases the application just needs to maintain a global map.

rsocket/RSocketServiceHandler.h
```diff
@@ -0,0 +1,97 @@
...
+      const SetupParameters&) = 0;
```
somasun: Not sure I understand.  Do you want this to be passed by value ?

rsocket/RSocketServiceHandler.h
```diff
@@ -0,0 +1,97 @@
...
+  onResume(ResumeIdentificationToken) {
+    return nullptr;
+  }
+
+  // This method gives a fine-grained control to the application during
+  // resumption (provided it agreed to resume in the onResume() call above).
+  // It tells the application what streams will be resumed in clean/dirty
+  // state.  If the application is okay with resuming the given clean/dirty
+  // streams, it should return true.
+  virtual bool canResume(
```
somasun: To have local-scope, the application should use per connection ServiceHandler.  If not it has to use the ResumeId to tie it back to a ServerState.

lexs: Does the API support per-connection ServiceHandlers? I can't see that.

You don't pass the ResumeId here so you can't tie this back to anything.

rsocket/RSocketServiceHandler.h
lexs: As `SetupParameters` contains the Payload you want to be able to take ownership of it, currently you are forced to copy it out.

rsocket/RSocketServiceHandler.h
lexs: We talked 1:1 and agreed this can be synchronous. 

Is it possible to merge `onNewSetup()` and `onNewRSocketState()` so we avoid this two-stage setup? Something like this could work:

```cpp
struct Acceptor {
  std::shared_ptr<RSocketServerState> accept(RSocketConnectionParams params);
}

/**
 * Must call either acceptor.accept() or throw an exception. Returning without calling
 * accept() is treated the same as throwing an exception.
 */
void onSetup(Acceptor& acceptor, SetupParameters&& params) throws RSocketException {
  auto state = acceptor.accept(createConnectionParams());
  saveState(params.token, std::move(state));
}
```

rsocket/internal/SetupResumeAcceptor.cpp
```diff
@@ -143,24 +164,23 @@ void SetupResumeAcceptor::processFrame(
         break;
       }
 
-      if (!onResume(transport, std::move(resumeParams))) {
+      try {
+        onResume(transport, std::move(resumeParams));
+      } catch (RSocketException ex) {
         transport->outputFrameOrEnqueue(frameSerializer->serializeOut(
-            Frame_ERROR::rejectedResume("Resumption Failure")));
+            Frame_ERROR::rejectedResume(ex.what())));
```
lexs: Not sure about this, this has the potential to leak information. Could we make a specific type that has a "user-friendly" message and use that?

rsocket/transports/tcp/TcpDuplexConnection.cpp
```diff
@@ -192,10 +192,15 @@ class TcpOutputSubscriber
...
+  void onError(std::exception_ptr) noexcept override {
     CHECK(tcpReaderWriter_);
     auto tcpReaderWriter = std::move(tcpReaderWriter_);
-    tcpReaderWriter->closeErr(folly::exception_wrapper(std::move(ex)));
+    // FIXME: See https://github.com/rsocket/rsocket-cpp/issues/589
+    // The expensive alternative is to throw and the get the exception
+    // reference.  Since the message is internal only (not sent over wire),
+    // discarding the message (ex) as a short term fix to get things working in
+    // a debug build.  We should move to one type soon.
+    tcpReaderWriter->closeErr(folly::exception_wrapper());
```
lexs: Commenting here again, I would prefer us to throw for now instead of changing behavior.

rsocket/RSocketServiceHandler.h
somasun: I think you are looking at a older version. The latest changes have a ResumeId

rsocket/RSocketServiceHandler.h
somasun: The ownership of SetupParams still resides with the protocol-implementation.  I did not want to move it away to the application.  The RSocketStateMachine::connectServer() for example, takes the SetupParams as parameter (though it does not use the initialPayload).  My reasoning around this has been, though initialPayload is not currently being used by the current protocol-implementation, moving in the direction where we lose this information could leave us handicapped later when we want to use it somewhere.

rsocket/internal/SetupResumeAcceptor.cpp
somasun: Discussed about this offline. The current design allows the server-application to pass any error message to the client in the ERROR msg payload (though not in a structured process).  This can be re-visited later independent of the API changes in the PR.

rsocket/RSocketServiceHandler.h
somasun: What you are proposing is similar to what we had previously (the application called `RSocketSetup.createRSocketStateMachine()`).  We had extensive discussion on this and decided to go with this approach.  The drawback with the old approach is that the application has to remember to call the right method and it will be repeated across all applications.  Making the application call a protocol-method also seemed like back-and-forth.  The old approach is optimized more for the case where the application wants to cold-resume.

But we expect most applications using RSocket to be simple (without leveraging cold-resumption).  In which case they have to worry only about the `onNewSetup()` method.  The current design is optimized to make the most common use-case simpler.

lexs: But wouldn't the simple case just use `OnNewSetupFn` which can keep that behavior? I don't think making setup awkward for complicated applications (which is the only thing we currently have) needs to happen just to make simpler applications easier.

I think we have all the bits (`OnNewSetupFn`) to make the simple case happen on top of the more advanced low-level API (`RSocketServiceHandler`).

If you want something inbetween you could easily have a default implementation of `RSocketServiceHandler`.

rsocket/RSocketServiceHandler.h
lexs: Why would you use the initial payload in the implementation? It's purely the applications concern no? connectServer() uses some of the metadata (resume token, resume flag etc).

Could we split this up into two classes in that case? One which is still a `const&` and one that is `&&`. I want us to avoid making copies for possibly large initial payloads.

PR #602, opened 2017-07-24
add IPAdress to TcpConnectionAcceptor::options

ragansa:
T19240097

I can push an IP in, and I see the server trying to use it.  I'm not sure what a good use case/test case demonstrating this would be.

### DIFF
--- a/rsocket/transports/tcp/TcpConnectionAcceptor.h
+++ b/rsocket/transports/tcp/TcpConnectionAcceptor.h
@@ -21,14 +21,11 @@ class TcpConnectionAcceptor : public ConnectionAcceptor {
  public:
   struct Options {
     explicit Options(uint16_t port_ = 8080, size_t threads_ = 2,
-                     int backlog_ = 10) : port(port_), threads(threads_),
+                     int backlog_ = 10) : socketAddress("::", port_), threads(threads_),
                                           backlog(backlog_) {}
 
-    /// IP address to listen on for TCP requests.
-    folly::IPAddress ipAddress{folly::IPAddress("::")};
-
-    /// Port to listen on for TCP requests.
-    uint16_t port;
+    /// Address to listen on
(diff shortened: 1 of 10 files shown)


alexmalyshev:
Instead of separate `folly::IPAddress ipAddress;` and `uint16_t port;` fields, you can just have a single `folly::SocketAddress address;` field.

ragansa:
Oh yeah, and that would avoid having to create it later as well.  (we build a folly::SocketAddress when we start the server).  I was trying to avoid making a change that would affect all the use cases currently out there.  

Most look like:
```
TcpConnectionAcceptor::Options opts;
  opts.port = FLAGS_port;
  opts.threads = 2;
```

Do we want to change how the options are put together?  I can go through and update the places that would be affected if we do.

alexmalyshev:
It shouldn't be too hard to go through all places where we do `opts.port = port;` and replace them with `opts.address = folly::SocketAddress{foo, port};`, I think it's worth it.

ragansa:
Cool, I'll update

rsocket/transports/tcp/TcpConnectionAcceptor.h
```diff
@@ -21,11 +21,11 @@ class TcpConnectionAcceptor : public ConnectionAcceptor {
...
   struct Options {
     explicit Options(uint16_t port_ = 8080, size_t threads_ = 2,
-                     int backlog_ = 10) : port(port_), threads(threads_),
+                     int backlog_ = 10) : socketAddress("::", port_), threads(threads_),
                                           backlog(backlog_) {}
 
-    /// Port to listen on for TCP requests.
-    uint16_t port;
+    /// Address to listen on
+    folly::SocketAddress socketAddress;
```
alexmalyshev: I think `address` is a better field name.

PR #604, opened 2017-07-26
Fix race in ~TcpConnectionAcceptor

alexmalyshev:
We need to destroy the "on accept" callback after the worker threads, as
the workers might still have outstanding "accept" events to process.


alexmalyshev:
Example crash in https://travis-ci.org/rsocket/rsocket-cpp/jobs/255781167

PR #607, opened 2017-07-28
Add warm-resumption unittest

somasun:
Added a simple Warm Resumption Unittest.  

It also helped in reproducing a heap-after-use in RSocketRequester.  If the TcpConnectionFactory/TcpConnectionAcceptor go out of scope (and hence their evbs), RSocketRequester will be holding on to invalid evbs and use them during ~RSocketRequester.

Notes:
- This does not prevent the case where RSocketRequester makes a request (which uses evb) after RSocketClient/RSocketServer go out of scope.  In the current state of the code, the application has to hold on to RSocketClient/RSocketServer instances to be able to use RSocketRequester.
- The TestSubscriber used in the tests is not thread-safe - https://github.com/rsocket/rsocket-cpp/issues/606

### DIFF
--- a/test/WarmResumptionTest.cpp
+++ b/test/WarmResumptionTest.cpp
@@ -1,0 +1,64 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#include <thread>
+
+#include "RSocketTests.h"
+
+#include "rsocket/RSocketServiceHandler.h"
+#include "yarpl/flowable/TestSubscriber.h"
+
+using namespace rsocket;
+using namespace rsocket::tests::client_server;
+using namespace yarpl::flowable;
+
+namespace {
+
+class HelloServiceHandler : public RSocketServiceHandler {
+ public:
+  folly::Expected<RSocketConnectionParams, RSocketException> onNewSetup(
+      const SetupParameters&) override {
+    return RSocketConnectionParams(
+        std::make_shared<rsocket::tests::HelloStreamRequestHandler>());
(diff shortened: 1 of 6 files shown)


test/WarmResumptionTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include "rsocket/RSocketServiceHandler.h"
+#include "rsocket/transports/tcp/TcpConnectionAcceptor.h"
+#include "rsocket/transports/tcp/TcpConnectionFactory.h"
+#include "yarpl/flowable/TestSubscriber.h"
+
+using namespace rsocket;
+using namespace yarpl::flowable;
+
+namespace {
+class HelloStreamRequestResponder : public RSocketResponder {
```
alexmalyshev: Can you just do `#include "test/handlers/HelloStreamRequestResponder.h"`?

test/WarmResumptionTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+  };
+
+ private:
+  folly::Synchronized<
+      std::map<ResumeIdentificationToken, std::shared_ptr<RSocketServerState>>,
+      std::mutex>
+      store_;
+};
+
+std::shared_ptr<RSocketClient> makeResumableClient(uint16_t port) {
```
alexmalyshev: Do you think this should be in `test/RSocketTests.h`instead?  I could see it being used from more than just this test file.

PR #613, opened 2017-07-31
Rewrite benchmarks/StreamThroughput.cpp as a folly benchmark

alexmalyshev:
Instead of google benchmark.  Using folly benchmark will simplify our build and
make it easier to build and run the benchmark binaries internally.

### DIFF
--- a/benchmarks/StreamThroughput.cpp
+++ b/benchmarks/StreamThroughput.cpp
@@ -1,6 +1,7 @@
 // Copyright 2004-present Facebook. All Rights Reserved.
 
-#include <benchmark/benchmark.h>
+#include <folly/Benchmark.h>
+#include <folly/init/Init.h>
 #include <folly/io/async/ScopedEventBaseThread.h>
 
 #include <gflags/gflags.h>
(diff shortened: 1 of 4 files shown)


benchmarks/StreamThroughput.cpp
```diff
@@ -152,69 +153,69 @@ class BM_Subscriber : public yarpl::flowable::Subscriber<Payload> {
...
 
-BENCHMARK_DEFINE_F(BM_RsFixture, BM_Stream_Throughput)
-(benchmark::State& state) {
-  folly::SocketAddress address;
-  address.setFromHostPort(host_, port_);
+  return server;
+}
 
-  auto s = yarpl::make_ref<BM_Subscriber>(state.range(0));
+std::shared_ptr<RSocketClient> makeClient(folly::SocketAddress address) {
```
somasun: the `makeServer` and `makeClient` methods are very similar to the ones in `test/RSocketTests.h`.  Worth reusing the code ?

alexmalyshev: Yeah they should.  I'll cover it in a follow-up PR.

PR #616, opened 2017-08-01
add coverage support for clang, fix coverage remove globs

dymk:
`lcov` wasn't stripping coverage information for gmock (on osx, at least), as the file paths for gmock were under a prefix; same for folly and various includes from XCode. New globs are more forgiving. 

adds support for code coverage when compiling with clang on macOS
adds support for TCK tests in macOS

also removes an errant duplicated FrameType.{h,cpp}

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,22 +100,23 @@ cache:
 before_script:
   # Install lcov to coveralls conversion + upload tool.
   - gem install coveralls-lcov
-  - lcov --directory build --zerocounters
+  - lcov --version
 
 script:
-  - mkdir build &&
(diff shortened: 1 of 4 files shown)


dymk:
Would also be nice if we could get @coveralls bot commenting code coverage changes

https://coveralls.io/features (ctrl+f "GITHUB AND CI INTEGRATION")

dymk:
Alright lcov just looks super brittle, this might take a few tries

dymk:
whew that took too many attempts; TIL don't attempt to use clang and gnu lcov in a linux env

PR #619, opened 2017-08-04
Add sane defaults for StateMachine parameters

somasun:
Some of these defaults got removed with API changes.
This is also to ensure the dependent code continues working.

### DIFF
--- a/rsocket/RSocketClient.cpp
+++ b/rsocket/RSocketClient.cpp
@@ -150,14 +150,23 @@ void RSocketClient::createState(folly::EventBase& eventBase) {
   // Creation of state is permitted only once for each RSocketClient.
   // When evb is removed from RSocketStateMachine, the state can be
   // created in constructor
-  CHECK(!stateMachine_);
+  CHECK(!stateMachine_) << "A stateMachine has already been created";
+
+  if (!keepaliveTimer_) {
+    keepaliveTimer_ = std::make_unique<FollyKeepaliveTimer>(
+        eventBase, std::chrono::seconds(5));
+  }
+
+  if (!responder_) {
+    responder_ = std::make_shared<RSocketResponder>();
+  }
 
   stateMachine_ = std::make_shared<RSocketStateMachine>(
       eventBase,
-      responder_,
+      std::move(responder_),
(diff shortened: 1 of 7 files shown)


rsocket/RSocketClient.cpp
```diff
@@ -150,14 +150,23 @@ void RSocketClient::createState(folly::EventBase& eventBase) {
   // Creation of state is permitted only once for each RSocketClient.
   // When evb is removed from RSocketStateMachine, the state can be
   // created in constructor
-  CHECK(!stateMachine_);
+  CHECK(!stateMachine_) << "A stateMachine has already been created";
+
+  if (!keepaliveTimer_) {
+    keepaliveTimer_ = std::make_unique<FollyKeepaliveTimer>(
+        eventBase, std::chrono::milliseconds(5000));
```
alexmalyshev: `std::chrono::seconds{5}`

rsocket/RSocketServer.cpp
```diff
@@ -128,7 +128,10 @@ void RSocketServer::onRSocketSetup(
     throw result.error();
   }
   auto connectionParams = result.value();
-  CHECK(connectionParams.responder);
+  if (!connectionParams.responder) {
+    LOG(ERROR) << "Received invalid Responder. Dropping connection";
```
alexmalyshev: Maybe `VLOG`?

somasun: If the server-application does not want to accept the connection, it should ideally return a RSocketException.  This case is for erroneous applications which return a NULL responder by mistake.  I did not want to terminate the server in those cases.  We can reconsider this later if it turns to be used as a feature (I hope not).

alexmalyshev: I wasn't commenting on the exception (which I think is fine), I just meant maybe replacing the `LOG(INFO)` with `VLOG(n)`.

somasun: Right, I wanted to treat the cases where the server-application returns a NULL responder as a error (which I saw in one of the tests).  And wanted to surface the error prominently in the logs.  It instead should be using RSocketException.

PR #621, opened 2017-08-07
Fix Warnings

somasun:


rsocket/RSocketClient.cpp
```diff
@@ -124,7 +124,7 @@ void RSocketClient::disconnect(folly::exception_wrapper ex) {
   CHECK(stateMachine_);
   evb_->runInEventBaseThread([ this, ex = std::move(ex) ] {
     VLOG(2) << "Disconnecting RSocketStateMachine on EventBase";
-    stateMachine_->disconnect(std::move(ex));
+    stateMachine_->disconnect(ex);
```
alexmalyshev: Leave the `std::move`, instead mark the lambda as `mutable`.

yarpl/test/Observable_test.cpp
```diff
@@ -122,7 +122,7 @@ TEST(Observable, OnError) {
   });
 
   a->subscribe(Observers::create<int>(
-      [](int value) { /* do nothing */ },
+      [](int /* value */) { /* do nothing */ },
```
alexmalyshev: You can just drop the comments too (i.e. `[](int) {},`

PR #624, opened 2017-08-07
Fix DuplexConnection comments, fix TCP impl closing connection on out…

alexmalyshev:
…put errors


alexmalyshev:
Yes, that's correct.

PR #625, opened 2017-08-07
Inline StreamState field into RSocketStateMachine

alexmalyshev:
No real reason for it to be a shared_ptr.

Initialize the stats_ field early, so that we won't accidentally pass a nullptr
to other fields that use it.

Replace asserts with DCHECKs to get rid of code like `(void)result;`.

### DIFF
--- a/rsocket/statemachine/RSocketStateMachine.cpp
+++ b/rsocket/statemachine/RSocketStateMachine.cpp
@@ -32,23 +32,18 @@ RSocketStateMachine::RSocketStateMachine(
     std::shared_ptr<RSocketStats> stats,
     std::shared_ptr<RSocketConnectionEvents> connectionEvents)
     : mode_(mode),
-      resumeCache_(std::make_shared<ResumeCache>(stats)),
-      streamState_(std::make_shared<StreamState>(*stats)),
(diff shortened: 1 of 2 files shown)


rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -32,23 +32,18 @@ RSocketStateMachine::RSocketStateMachine(
     std::shared_ptr<RSocketStats> stats,
     std::shared_ptr<RSocketConnectionEvents> connectionEvents)
     : mode_(mode),
-      resumeCache_(std::make_shared<ResumeCache>(stats)),
-      streamState_(std::make_shared<StreamState>(*stats)),
+      stats_(stats ? stats : RSocketStats::noop()),
+      resumeCache_(stats_),
```
somasun: With cold-resumption, resumeCache will be constructed and passed in by the client-application.  Keeping it a smart-pointer will make it convenient on the application part (if the application also wants access to it)

alexmalyshev: Ah okay, I'll drop the `ResumeCache` changes then.

PR #626, opened 2017-08-08
ResumeManager and InMemResumeManager for resumption.

somasun:
Introduce a ResumeManager interface.  Move the existing ResumeCache to be an implementation of ResumeManager and retitle it as InMemResumeManager.

This PR contains only the functions of ResumeManager which the InMemResumeManager cares about.

I will follow up with 
- other methods of ResumeManager
- ColdResumeHandler and its default implementation.
- Cold resumption plumbing in RSocketStateMachine
- tests


rsocket/ResumeManager.h
```diff
@@ -2,7 +2,73 @@
...
+class ResumeManager {
+ public:
+  virtual ~ResumeManager(){};
+
+  // The following methods will be called for each frame which is being
+  // sent/received on the wire.  The application should implement a way to
+  // store the sent and received frames in persistent storage.
+  virtual void trackReceivedFrame(
+      const folly::IOBuf& serializedFrame,
+      const FrameType frameType,
```
alexmalyshev: Drop the `const` for `FrameType`, `StreamId`, and `folly::Optional<>`.  They're value types and aren't affected by it.

rsocket/ResumeManager.h
```diff
@@ -2,7 +2,73 @@
...
+  // indexed by position.  This API gives the flexibility to the application to
+  // store the frames in any way it wants (randomly accessed or sequentially
+  // accessed).
+  virtual void sendFramesFromPosition(
+      ResumePosition position,
+      FrameTransport& transport) const = 0;
+
+  // This should return the first (oldest) available position in the send
+  // buffer.
+  virtual ResumePosition firstSentPosition() const = 0;
```
alexmalyshev: Thanks for renaming these.

rsocket/internal/InMemResumeManager.h
```diff
@@ -91,9 +69,9 @@ class ResumeCache {
   std::shared_ptr<RSocketStats> stats_;
 
   // End position of the send buffer queue
-  ResumePosition position_{0};
+  ResumePosition lastSentPosition_{0};
```
alexmalyshev: Nitpicking, but maybe swap the order of these two so `firstSentPosition_` is the first field?

rsocket/ResumeManager.h
```diff
@@ -2,7 +2,73 @@
...
+      const FrameType frameType,
+      const StreamId streamId) = 0;
+
+  virtual void trackSentFrame(
+      const folly::IOBuf& serializedFrame,
+      const FrameType frameType,
+      const folly::Optional<StreamId> streamIdPtr) = 0;
+
+  // We have received acknowledgement from the remote-side that it has frames
+  // upto "position".  We can discard all frames before that.  This information
```
alexmalyshev: Nit: `up to`

rsocket/ResumeManager.h
```diff
@@ -2,7 +2,73 @@
...
+  // is periodically received from remote-side through KeepAlive frames.
+  virtual void resetUpToPosition(ResumePosition position) = 0;
+
+  // The application should check its persistent storage and respond whether it
+  // has frames starting from "position" in send buffer.
```
alexmalyshev: Maybe mention that `position` needs to explicitly be an element in the send buffer or the end, i.e. if the send buffer is [2,4,6,8], position=5 isn't valid even though it's part of the send buffer's bytes.

PR #629, opened 2017-08-09
fixing failure path for RSocketClient creation

lehecka:
The core of this PR is removing TcpConnectionFactory::worker_ of type folly::ScopedEventBaseThread. Instead for the RScketClient the EventBase instance is passed from the outside.

The reasons for this change:
1. The problem was that the EventBase was destroyed before some of the RSocketClient callbacks were called. There was no clean solution to work around this. We need the EventBase to outlive the RSocket::* methods calls.
2. on mobile platform the EventBase will be provided to the RSocketClient from the outside.

I had to change the ConnecitonFactory::connect method to return Future in order to be able to call onError path for the RSocket::createConnectedClient method (solving the "unrealized promise" exception).


somasun:
The approach looks good to me.  The API gets a little clunky with the ConnectionFactory having to take an EventBase.  But conceptually it seems right.  It is better if the application controls the evb which runs the client state machine.  RSocket client library can get out of the business of spawning threads/evbs.  On the con side, if the application schedules a never-ending task on the evb or terminates the loop, then it might often appear as a RSocket bug.  But seems like a okay trade-off.

test/RSocketClientTest.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include <folly/io/async/ScopedEventBaseThread.h>
+#include <gtest/gtest.h>
+
+#include "rsocket/transports/tcp/TcpConnectionFactory.h"
+
+using namespace rsocket;
+using namespace rsocket::tests;
+using namespace rsocket::tests::client_server;
+
+TEST(RSocketClient, ConnectFails) {
```
somasun: This seems to be causing an ASAN failure.

rsocket/ConnectionFactory.h
```diff
@@ -42,6 +39,11 @@ class ConnectionFactory {
    *
    * Resource creation depends on the particular implementation.
    */
-  virtual void connect(OnDuplexConnectionConnect onConnect) = 0;
+  struct ConnectedDuplexConnection {
```
alexmalyshev: Nit: Maybe move this above the comment block for `connect()`?

rsocket/ConnectionFactory.h
```diff
@@ -42,6 +39,11 @@ class ConnectionFactory {
...
+    std::unique_ptr<rsocket::DuplexConnection> connection;
+    folly::EventBase &eventBase;
```
alexmalyshev: Nit: `folly::EventBase& eventBase;`

rsocket/RSocketClient.h
```diff
@@ -93,7 +93,7 @@ class RSocketClient {
 
   // Remember the evb on which the client was created.  Ensure warme-resume()
   // operations are done on the same evb.
-  folly::EventBase* evb_;
+  folly::EventBase* evb_{nullptr};
```
alexmalyshev: 👍

test/RSocketClientTest.cpp
```diff
@@ -0,0 +1,29 @@
...
+  folly::ScopedEventBaseThread worker;
+
+  folly::SocketAddress address;
+  address.setFromHostPort("localhost", 9876);
```
alexmalyshev: Nit: I think port==1 is better for an "obvious can't connect failure" type situation.  Wasn't immediately obvious that 9876 was expected to have nothing bound to it.

test/RSocketClientTest.cpp
```diff
@@ -0,0 +1,33 @@
...
+                                             std::move(address)));
+
+  folly::Baton<> baton;
+  client.then([&](auto&) {
+    FAIL() << "the test needs to fail";
+    baton.post();
+  }).onError([&](const std::exception&) {
+    LOG(INFO) << "connection failed as expected";
+    baton.post();
+  });
```
alexmalyshev: Can't you do `.get()` on the future instead of using a manual `folly::Baton`?  Or was that hitting the ASAN issues?  (I've hit them before too and they're weird...)

lehecka: I tried wait() which didn't work. I will try get().

PR #636, opened 2017-08-13
Remove output buffering in FrameTransport

lehecka:
This is a the first diff in the series towards removing FrameTransport class. 

* removed flow control for the output frames stream in FrameTransport. This is possible because of the previous changes we made in all implementations of DuplexConnection interface. The transport layer is expected to process all output frames without backpressure. In the theoretical case the transport layer cannot process the output frame, it should error out the stream. It might be reasonable to change the DuplexConnection::getOutput to return Observer instead of Subscriber.

* removed buffering of the output frames in FrameTransport. This was done as a pure code cleanup by cleaning up how we used FrameTransport class, we don't need to have another layer of buffering. There is one existing place in RSocketStateMachine where we buffer output frames in the case we are disconnected.

I renamed FrameTransport::outputFrame* to FrameTransport::outputFrameOrDrop to make it clear that for the closed connection we will drop the frame. This is ok since the RSocket is either definitely closed or the frame was duplicated in the resumption cache (so not really lost).

I tested the change via all unit tests and all examples with I manually ran.

In the next step I will remove the input buffering in FrameTransport and then the whole class all together.


rsocket/framing/FrameTransport.cpp
```diff
@@ -161,44 +160,25 @@ void FrameTransport::onError(std::exception_ptr eptr) {
...
 }
 
 void FrameTransport::cancel() {
   VLOG(6) << "cancel";
   terminateProcessor(folly::exception_wrapper());
 }
 
-void FrameTransport::outputFrameOrEnqueue(std::unique_ptr<folly::IOBuf> frame) {
+void FrameTransport::outputFrameOrDrop(std::unique_ptr<folly::IOBuf> frame) {
   Lock lock(mutex_);
```
phoad: Do we still need to lock for writes?

test/framing/FrameTransportTest.cpp
```diff
@@ -48,28 +48,6 @@ TEST(FrameTransport, CloseWithError) {
   transport->closeWithError(std::runtime_error("Uh oh"));
 }
 
-TEST(FrameTransport, SimpleEnqueue) {
```
phoad: This test might be still useful that if no FrameProcessor is set, we can check that the frames are successfully dropped and no crash etc happens.

rsocket/transports/tcp/TcpDuplexConnection.cpp
```diff
@@ -210,7 +211,7 @@ class TcpInputSubscription : public Subscription {
 
   void request(int64_t n) noexcept override {
     DCHECK(tcpReaderWriter_);
-    DCHECK(n == kMaxRequestN)
+    DCHECK_EQ(n, std::numeric_limits<int64_t>::max())
```
phoad: Frame_Request_N's constructor checks that the N value is to be smaller than INT_MAX. Would it be a problem to pass int64_max in here?

rsocket/framing/FrameTransport.cpp
lehecka: The lock is not needed now at all (in the whole class). We don't support resumption from multiple threads now so we never access the member variables from a different thread.
In this diff I don't want to remove more than its necessary.

rsocket/transports/tcp/TcpDuplexConnection.cpp
lehecka: The short answer is no, its not a problem.
We have 2 max values to use. The yarpl library is using int64_t max value and the rsocket protocol is using int32_t max value.
In this case we are using yarpl interfaces and the value is never sent over the network via rsocket.

test/framing/FrameTransportTest.cpp
lehecka: The whole FrameTransport class is going away so I would argue we don't have to worry about the FrameTransportTests which will go away too.

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -76,25 +76,27 @@ void RSocketStateMachine::setResumable(bool resumable) {
...
+void RSocketStateMachine::connectServer(
     yarpl::Reference<FrameTransport> frameTransport,
     const SetupParameters& setupParams) {
   setResumable(setupParams.resumable);
-  return connect(std::move(frameTransport), true, setupParams.protocolVersion);
+  CHECK(connect(std::move(frameTransport), setupParams.protocolVersion));
+  sendingPendingFrames();
 }
 
+//TODO: ensure the return value is not ignored!
```
alexmalyshev: What's the point of this TODO?  How are we going to ensure the return value isn't ignored here?  With C++17's `[[nodiscard]]`?

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -136,24 +138,23 @@ bool RSocketStateMachine::connect(
...
-    auto outputFrames = streamState_.moveOutputPendingFrames();
-    for (auto& frame : outputFrames) {
-      outputFrameOrEnqueue(std::move(frame));
-    }
-
-    // TODO: turn on only after setup frame was received
-    if (keepaliveTimer_) {
-      keepaliveTimer_->start(shared_from_this());
-    }
+void RSocketStateMachine::sendingPendingFrames() {
```
alexmalyshev: Nit: Seems like this should be `send` instead of `sending`.

rsocket/statemachine/RSocketStateMachine.cpp
lehecka: no, I was going to address this in the following diff. I didn't know about [[nodiscard]]. thanks! ;)

PR #637, opened 2017-08-14
Improvements on the stream benchmark

alexmalyshev:
* Use a gflag to control the number of items to send.  Much easier to profile,
  and configurable so we can run it as a test as part of Travis.

* Get rid of the yield logic.  Have the subscriber cancel the subscription once
  it hits the number of requested items.

* Replace mutex+condvar+bool with folly::Baton.

* Delete BM_Subscription, it was unused.

### DIFF
--- a/benchmarks/StreamThroughput.cpp
+++ b/benchmarks/StreamThroughput.cpp
@@ -1,5 +1,6 @@
 // Copyright 2004-present Facebook. All Rights Reserved.
 
+#include <folly/Baton.h>
 #include <folly/Benchmark.h>
 #include <folly/init/Init.h>
 #include <folly/io/async/ScopedEventBaseThread.h>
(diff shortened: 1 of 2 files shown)


benchmarks/StreamThroughput.cpp
```diff
@@ -22,39 +23,7 @@ using namespace rsocket;
...
-    VLOG(3) << "cancellation received";
-    cancelled_ = true;
-  }
-
-  yarpl::Reference<yarpl::flowable::Subscriber<Payload>> subscriber_;
-  std::string data_;
-  size_t currentElem_ = 0;
-  std::atomic_bool cancelled_;
-};
+DEFINE_int32(items, 1000000, "number of items in stream");
```
phoad: Why was BENCHMARKPARAM not working for Travis?
It is nice to have various numbers for this parameter to observe the complexity of the functionalities, like if we double the size or quadruple it how much the spent time increases, linear or not, or what's the multiplier for liner O(k.n) etc.

benchmarks/StreamThroughput.cpp
```diff
@@ -72,85 +41,51 @@ class BM_Subscriber : public yarpl::flowable::Subscriber<Payload> {
...
-    if (--requested_ == thresholdForRequest_) {
-      int toRequest = (initialRequest_ - thresholdForRequest_);
-      VLOG(2) << "BM_Subscriber " << this << " requesting " << toRequest
-              << " more items";
-      requested_ += toRequest;
-      subscription_->request(toRequest);
-    };
-
-    if (cancel_) {
+    if (received_.fetch_add(1) == requested_ - 1) {
```
phoad: If requested_ is 0, this will never end.

phoad: nit though.

benchmarks/StreamThroughput.cpp
alexmalyshev: Because we compiled with runs for 1000, 100000, and 1000000.  Running all of that takes too long (in my opinion), and it can't be configured at runtime, hence I didn't want to enable it for Travis.

Replacing that with a gflag lets the user control the number of items at runtime, and if they want multiple values then they can run the binary multiple times.

benchmarks/StreamThroughput.cpp
alexmalyshev: True, though I'm tempted to be lazy and say that's not valid as `request(0)` isn't valid, hence we don't need to account for it :smile:

PR #638, opened 2017-08-14
Porting baseline benchmark tests to folly benchmarks

lehecka:
Also added AsyncSocket baseline benchmarks.


benchmarks/BaselinesTcp.cpp
```diff
@@ -0,0 +1,178 @@
...
+}
+
+BENCHMARK_PARAM(BM_Baseline_TCP_Throughput_100MB_s40B_r1024B, 1);
+BENCHMARK_PARAM(BM_Baseline_TCP_Throughput_100MB_s40B_r4096B, 1);
+BENCHMARK_PARAM(BM_Baseline_TCP_Throughput_100MB_s80B_r4096B, 1);
+BENCHMARK_PARAM(BM_Baseline_TCP_Throughput_100MB_s4096B_r4096B, 1);
+
+void BM_Baseline_TCP_Latency_1M_msgs_32B(unsigned, unsigned) {
+  constexpr size_t messageSizeB = 32;
+  constexpr size_t loadSizeB = 1000000 * messageSizeB;
```
phoad: nit: we can put point in the integer values for easy read.
1000000 => 1.000.000

benchmarks/BaselinesTcp.cpp
```diff
@@ -0,0 +1,178 @@
...
+  std::atomic<bool> accepted{false};
+
+  std::thread t([&]() {
+    int serverSock = socket(AF_INET, SOCK_STREAM, 0);
+    int sock = -1;
+    struct sockaddr_in addr;
+    socklen_t addrlen = sizeof(addr);
+    char message[MAX_MESSAGE_LENGTH];
+
+    std::memset(message, 0, sizeof(message));
```
phoad: nit: Do we always send 0's? seems no need to clear its content.

benchmarks/BaselinesAsyncSocket.cpp
```diff
@@ -0,0 +1,248 @@
...
+void BM_Baseline_AsyncSocket_Throughput_100MB_s4096B_r4096B(
+    unsigned,
+    unsigned) {
+  constexpr size_t loadSizeB = 100 * 1024 * 1024;
+  constexpr size_t sendSizeB = 4096;
+  constexpr size_t receiveSizeB = 4096;
+  BM_Baseline_AsyncSocket_SendReceive(loadSizeB, sendSizeB, receiveSizeB);
+}
+
+BENCHMARK_PARAM(BM_Baseline_AsyncSocket_Throughput_100MB_s40B_r1024B, 1);
```
alexmalyshev: If you're always going to pass 1, then you should be using `BENCHMARK()` instead of BENCHMARK_PARAM()`.

benchmarks/BaselinesAsyncSocket.cpp
```diff
@@ -0,0 +1,248 @@
...
+static void BM_Baseline_AsyncSocket_SendReceive(
+    size_t loadSize,
+    size_t msgLength,
+    size_t recvLength) {
+  EventBase serverEventBase;
+  auto serverSocket = AsyncServerSocket::newSocket(&serverEventBase);
+
+  ServerAcceptCallback serverCallback(serverEventBase, loadSize, recvLength);
+
+  SocketAddress addr("::", PORT);
```
alexmalyshev: I think we should be using port 0 instead of a hardcoded port.  We can get the list of listened-on addresses from the `AsyncServerSocket` and pass the resolved ephemeral port to the client.

benchmarks/StreamThroughput.cpp
```diff
@@ -178,8 +178,6 @@ void streamThroughput(unsigned, size_t items) {
   yarpl::Reference<BM_Subscriber> subscriber;
 
   BENCHMARK_SUSPEND {
-    LOG(INFO) << "  Running with " << items << " items";
```
alexmalyshev: Given the recent changes in master, I think it's fine to leave this in.  Most users are going to use a high number for `items` so there won't be any more spammy lines of "Running with 1000 items".

benchmarks/BaselinesTcp.cpp
lehecka: we have seen odd behavior when the buffer was uninitialized. I never got into trying to understand why so and am just keeping it as it.

benchmarks/BaselinesTcp.cpp
lehecka: in C++?
isn't 1.000 just a float(1)?

benchmarks/BaselinesAsyncSocket.cpp
lehecka: Yeah, nice to have. I will try to do it in the next diff.

benchmarks/BaselinesTcp.cpp
alexmalyshev: I think he mean't `1'000'000`, which was added in C++14 [http://en.cppreference.com/w/cpp/language/integer_literal]

benchmarks/BaselinesAsyncSocket.cpp
lehecka: yeah, I tried that, but the compiler generates this error:
error: function 'BM_Baseline_AsyncSocket_Latency_1M_msgs_32B' has internal linkage but is not defined [-Werror,-Wundefined-internal]
BENCHMARK(BM_Baseline_AsyncSocket_Latency_1M_msgs_32B);

alexmalyshev: That's not how it's used, you put the function definition inside the `BENCHMARK` macro.

see https://github.com/rsocket/rsocket-cpp/blob/master/benchmarks/StreamThroughput.cpp#L110

lehecka: got it, will do.

PR #639, opened 2017-08-15
Templatize SingleOperator over Operator

dymk:
Removes unneeded static_cast in MapOperator

### DIFF
--- a/yarpl/include/yarpl/single/SingleOperator.h
+++ b/yarpl/include/yarpl/single/SingleOperator.h
@@ -32,11 +32,12 @@ class SingleOperator : public Single<D> {
   /// against Operators.  Each operator subscription has two functions: as a
   /// observer for the previous stage; as a subscription for the next one,
   /// the user-supplied observer being the last of the pipeline stages.
+  template <typename Operator>
   class Subscription : public ::yarpl::single::SingleSubscription,
                        public SingleObserver<U> {
    protected:
     Subscription(
-        Reference<Single<D>> single,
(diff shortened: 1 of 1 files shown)


alexmalyshev:
I'm not convinced this makes the world better... While I'll agree the code is more type-safe and that's a good thing, it's much harder to follow along with what's going on.

dymk:
I could templatize the Subscriber in SingleOperator over the Operator instead of SingleOperator itself - so you don't have inheritance that looks like `<U, D, F> class MapOperator : public SingleOperator<U, D, MapOperator<U, D, F>>`; just `... : public SingleOperator<U, D>`

fwiw, in the original code I had to squint at why usage of getObservableAs wasn't entirely broken, and that the Single<D> was actually the MapOperator itself (this version makes that explicit)

yarpl/include/yarpl/single/SingleOperator.h
```diff
@@ -97,33 +96,33 @@ template <
     typename D,
     typename F,
     typename = typename std::enable_if<std::is_callable<F(U), D>::value>::type>
-class MapOperator : public SingleOperator<U, D> {
+class MapOperator : public SingleOperator<U, D, MapOperator<U, D, F>> {
+  using ThisMap = MapOperator<U, D, F>;
```
lehecka: please consider renaming ThisMap to ThisOperatorT
MSuper to Super
and removing SSuper or renaming SSuper to ThisOperatorSubscriptionT

just improving readability...

PR #641, opened 2017-08-16
Fix extraneous semi-colon in ResumeManager

alexmalyshev:
-Wextra-semi should be supported by gcc and clang, so it's fine to set for all
 builds.

### DIFF
--- a/rsocket/ResumeManager.h
+++ b/rsocket/ResumeManager.h
@@ -27,7 +27,7 @@ namespace rsocket {
 // - lastSentPosition() would return 350
 class ResumeManager {
  public:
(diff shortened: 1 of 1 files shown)


alexmalyshev:
Context here is that we have some internal builds with this warning enabled, so we want to match the warning in the OSS build.

alexmalyshev:
Huh https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html lied to me, there's no -Wextra-semi in gcc.  Well alrighty then, I guess I'll just limit this PR to killing the semi and not touch the build.

PR #642, opened 2017-08-16
Fuzzer harness for rsocket server

dymk:
A harness for end-to-end testing rsocketserver (aside from any TCP specific stuff); implements a dummy duplex connection fed with input from afl-fuzz (although other fuzzers could be hooked up pretty easily). Reads input fuzzing/test cases from stdin. 

Rsocket/the harness is built like:   
 - `env CC=~/afl-2.49b/afl-gcc CXX=~/afl-2.49b/afl-g++ cmake -DCMAKE_BUILD_TYPE=Debug -DRSOCKET_ASAN=Off ......`
 - `env AFL_HARDEN=1 make -j8 frame_fuzzer`

I highly recommend setting up a Linux virtual machine rather than running this on macOS, AFL runs incredibly slow under macOS.

It's found a few cases so far, mostly segfaults; still triaging to figure out if they're actually interesting or not.

### DIFF
--- a/test/fuzzers/frame_fuzzer.cpp
+++ b/test/fuzzers/frame_fuzzer.cpp
@@ -1,0 +1,120 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+#include <folly/String.h>
+#include <folly/io/async/EventBaseManager.h>
+#include <iostream>
+
+#include <folly/init/Init.h>
+#include <folly/portability/GFlags.h>
+#include <glog/logging.h>
+
+#include "rsocket/ConnectionAcceptor.h"
+#include "rsocket/DuplexConnection.h"
+#include "rsocket/RSocketServer.h"
+
+struct FuzzerConnectionAcceptor : rsocket::ConnectionAcceptor {
+  void start(rsocket::OnDuplexConnectionAccept func_) override {
+    VLOG(1) << "FuzzerConnectionAcceptor::start()" << std::endl;
+    func = func_;
+  }
+
+  void stop() override {
+    VLOG(1) << "FuzzerConnectionAcceptor::stop()" << std::endl;
+  }
+
+  folly::Optional<uint16_t> listeningPort() const override {
+    return 0;
+  }
+
+  rsocket::OnDuplexConnectionAccept func;
+};
+
+struct FuzzerDuplexConnection : rsocket::DuplexConnection {
+  using Subscriber = rsocket::DuplexConnection::Subscriber;
+
+  struct SinkSubscriber : Subscriber {
+    std::vector<std::unique_ptr<folly::IOBuf>> sent_buffers;
+
+    void onNext(std::unique_ptr<folly::IOBuf> buf) {
+      VLOG(1) << "SinkSubscriber::onNext(\""
+              << folly::humanify(buf->cloneAsValue().moveToFbString()) << "\")"
+              << std::endl;
+      sent_buffers.push_back(std::move(buf));
+    }
+  };
+
+  FuzzerDuplexConnection() : output_sub(yarpl::make_ref<SinkSubscriber>()) {}
+
+  virtual void setInput(yarpl::Reference<Subscriber> sub) {
+    VLOG(1) << "FuzzerDuplexConnection::setInput()" << std::endl;
+    input_sub = sub;
+  }
+  virtual yarpl::Reference<Subscriber> getOutput() {
+    VLOG(1) << "FuzzerDuplexConnection::getOutput()" << std::endl;
+    return output_sub;
+  }
+
+  yarpl::Reference<Subscriber> input_sub;
+  yarpl::Reference<Subscriber> output_sub;
+};
+
+struct NoopSubscription : yarpl::flowable::Subscription {
+  void request(int64_t n) override {
+    VLOG(1) << "NoopSubscription::request(" << n << ")";
+  }
+  void cancel() override {
+    VLOG(1) << "NoopSubscription::cancel()";
+  }
+};
+
+struct NoopResponder : rsocket::RSocketResponder {};
+
+std::string get_stdin() {
+  std::cin >> std::noskipws;
+  std::istream_iterator<char> it(std::cin);
+  std::istream_iterator<char> end;
+  std::string input(it, end);
+  return input;
+}
+
+int main(int argc, char* argv[]) {
+  folly::init(&argc, &argv);
+  FLAGS_logtostderr = 1;
+
+  folly::EventBase evb;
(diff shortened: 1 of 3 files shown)


test/fuzzers/frame_fuzzer.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+  CHECK(acceptor_func_ptr);
+  acceptor_func_ptr(std::move(feed_conn), evb);
+  evb.loopOnce();
+
+  CHECK(input_sub);
+  CHECK(output_sub);
+  auto input_subscription = yarpl::make_ref<FuzzerSubscription>();
+  input_sub->onSubscribe(input_subscription);
+  input_sub->onNext(std::move(buf));
```
dymk: There's probably a better way of doing this, and I'd love to know how. For now, this imitates a frame coming in off the wire.

test/fuzzers/frame_fuzzer.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+  void stop() override {
+    VLOG(1) << "FuzzerConnectionAcceptor::stop()" << std::endl;
+  }
+
+  folly::Optional<uint16_t> listeningPort() const override {
+    return {0};
+  }
+
+  rsocket::OnDuplexConnectionAccept func_;
```
alexmalyshev: You should drop the trailing `_`, those are generally used for private fields.

test/fuzzers/frame_fuzzer.cpp
```diff
@@ -0,0 +1,119 @@
...
+    VLOG(1) << "FuzzerConnectionAcceptor::start()" << std::endl;
+    func_ = func;
+  }
+
+  void stop() override {
+    VLOG(1) << "FuzzerConnectionAcceptor::stop()" << std::endl;
+  }
+
+  folly::Optional<uint16_t> listeningPort() const override {
+    return {0};
```
alexmalyshev: `return 0;` or `return {};` is even shorter.

test/fuzzers/frame_fuzzer.cpp
```diff
@@ -0,0 +1,119 @@
...
+  virtual void setInput(yarpl::Reference<Subscriber> sub) {
+    VLOG(1) << "FuzzerDuplexConnection::setInput()" << std::endl;
+    input_sub_ = sub;
+  }
+  virtual yarpl::Reference<Subscriber> getOutput() {
+    VLOG(1) << "FuzzerDuplexConnection::getOutput()" << std::endl;
+    return output_sub_;
+  }
+
+  yarpl::Reference<Subscriber> input_sub_;
```
alexmalyshev: Nit: Ditto about trailing `_`.  I also think these should be camelCase to match the rest of the codebase.

test/fuzzers/frame_fuzzer.cpp
```diff
@@ -0,0 +1,119 @@
...
+  yarpl::Reference<Subscriber> output_sub_;
+};
+
+struct FuzzerSubscription : public yarpl::flowable::Subscription {
+  void request(int64_t n) noexcept override {
```
alexmalyshev: Drop the `noexcept`s on these two methods.

test/fuzzers/frame_fuzzer.cpp
```diff
@@ -0,0 +1,119 @@
...
+      VLOG(1) << "SinkSubscriber::onNext(\""
+              << folly::humanify(buf->cloneAsValue().moveToFbString()) << "\")"
+              << std::endl;
+      sunk_buffers_.push_back(std::move(buf));
+    }
+  };
+
+  FuzzerDuplexConnection() : output_sub_(yarpl::make_ref<SinkSubscriber>()) {}
+
+  virtual void setInput(yarpl::Reference<Subscriber> sub) {
```
alexmalyshev: `override` instead of `virtual` for both methods.

test/fuzzers/frame_fuzzer.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+#include <folly/init/Init.h>
+#include <folly/portability/GFlags.h>
+#include <glog/logging.h>
+
+#include "rsocket/ConnectionAcceptor.h"
+#include "rsocket/DuplexConnection.h"
+#include "rsocket/RSocketServer.h"
+
+struct FuzzerConnectionAcceptor : public rsocket::ConnectionAcceptor {
```
alexmalyshev: If you're using `struct`, then you don't need the `public` keyword for the super class.  Here and all following classes.

test/fuzzers/frame_fuzzer.cpp
```diff
@@ -0,0 +1,119 @@
...
+  std::istream_iterator<char> end;
+  std::string input(it, end);
+  return input;
+}
+
+int main(int argc, char* argv[]) {
+  folly::init(&argc, &argv);
+  FLAGS_logtostderr = 1;
+
+  std::string fuzz_input = get_stdin();
```
alexmalyshev: Both vars here should be `auto`.

test/fuzzers/frame_fuzzer.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+  CHECK(input_sub);
+  CHECK(output_sub);
+  auto input_subscription = yarpl::make_ref<FuzzerSubscription>();
+  input_sub->onSubscribe(input_subscription);
+  input_sub->onNext(std::move(buf));
+  evb.loopOnce();
+
+  return 0;
+}
```
alexmalyshev: Can you configure your editor to write a trailing newline at the end of files?  These are going to make the FB internal linter very unhappy.

test/fuzzers/frame_fuzzer_net.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "rsocket/RSocket.h"
+#include "rsocket/transports/tcp/TcpConnectionAcceptor.h"
+#include "yarpl/Flowable.h"
+
+using namespace rsocket;
+using namespace yarpl::flowable;
+
+DEFINE_int32(port, 9898, "port to connect to");
+
+struct FrameFuzzerResponder : public rsocket::RSocketResponder {
```
alexmalyshev: This doesn't seem to have anything regarding fuzzing here, it's just a server with a default responder.

I'd drop the mentions to fuzzing and rename things along the lines of `EmptyServer.cpp` or `DefaultServer.cpp`

test/fuzzers/frame_fuzzer.cpp
dymk: (y)

PR #647, opened 2017-08-17
Update request-response benchmark to folly benchmark

alexmalyshev:
It was entirely commented out, so make it work.  Sends 1,000,000
request-response requests to a server, by default.

Needed to fix bugs in Single where we weren't calling onSubscribe() on the
SingleObserver in Singles::create().


alexmalyshev:
Wait, running this in debug mode breaks.  Seems like we're not calling `SingleObserver::onSubscribe()`?

PR #649, opened 2017-08-17
Remove manual refcounting in FlowableOperator

dymk:
Changes from #648 remove the need for manual refcounting to be done in FlowableOperator

### DIFF
--- a/yarpl/include/yarpl/flowable/FlowableOperator.h
+++ b/yarpl/include/yarpl/flowable/FlowableOperator.h
@@ -43,11 +43,6 @@ class FlowableOperator : public Flowable<D> {
         : flowable_(std::move(flowable)), subscriber_(std::move(subscriber)) {
       assert(flowable_);
       assert(subscriber_);
-
-      // We expect to be heap-allocated; until this subscription finishes (is
-      // canceled; completes; error's out), hold a reference so we are not
-      // deallocated (by the subscriber).
-      Refcounted::incRef(*this);
     }
 
     Reference<Operator> getFlowableOperator() {
@@ -151,8 +146,6 @@ class FlowableOperator : public Flowable<D> {
           }
         }
       }
-
-      Refcounted::decRef(*this);
     }
 
     /// The Flowable has the lambda, and other creation parameters.


yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -43,11 +43,6 @@ class FlowableOperator : public Flowable<D> {
         : flowable_(std::move(flowable)), subscriber_(std::move(subscriber)) {
       assert(flowable_);
       assert(subscriber_);
-
-      // We expect to be heap-allocated; until this subscription finishes (is
-      // canceled; completes; error's out), hold a reference so we are not
-      // deallocated (by the subscriber).
-      Refcounted::incRef(*this);
```
phoad: Might this mean `delayed destruction` is wanted?
So we delay the destruction up till terminateImpl is being called. Otherwise, if ref count drops to zero, then this object will be deleted immediately. We might be wanting to prevent that to happen.

alexmalyshev:
I *think* this is fine, especially since we're running ASAN+LSAN tests on Travis and those all passed...

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: With these refcounted types, if another object has a handle to it, it should never be destroyed; passing around a Reference rather than a raw pointer fixes that. Having to manually inc/dec the refcount is a bug.

phoad: Can you add that fix to this PR too? The line that we were passing `*this` to a constructor, using the Subscriber&. Should we pass yarpl::Reference<Subscriber> in that Emitter class?

I could not be able to see the benefit of manually calling inc_ref, dec_ref other than delayed destruction, for which we already have a comment stating that the original implementor wants to have it. I am not sure if that comment still has validity. Can you sync up with @lehecka ?

lehecka: I believe this class was historically written as a copy/paste of SynchronousSubscription which has this refcounting logic and its necessary. In this case, since the OperatorSubscription is always held by the reference, I agree that it is unnecessary.

PR #650, opened 2017-08-17
Remove manual refcounting in ObservableOperator

dymk:
Doesn't need typesafe fix, just works without additional manual refcounting incs/decs

### DIFF
--- a/yarpl/include/yarpl/observable/ObservableOperator.h
+++ b/yarpl/include/yarpl/observable/ObservableOperator.h
@@ -41,11 +41,6 @@ class ObservableOperator : public Observable<D> {
         : observable_(std::move(observable)), observer_(std::move(observer)) {
       assert(observable_);
       assert(observer_);
-
-      // We expect to be heap-allocated; until this subscription finishes (is
-      // canceled; completes; error's out), hold a reference so we are not
-      // deallocated (by the subscriber).
-      Refcounted::incRef(*this);
     }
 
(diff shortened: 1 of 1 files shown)


yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -41,11 +41,6 @@ class ObservableOperator : public Observable<D> {
         : observable_(std::move(observable)), observer_(std::move(observer)) {
       assert(observable_);
       assert(observer_);
-
-      // We expect to be heap-allocated; until this subscription finishes (is
-      // canceled; completes; error's out), hold a reference so we are not
-      // deallocated (by the subscriber).
-      Refcounted::incRef(*this);
```
phoad: Might this mean `delayed destruction` is wanted?
So we delay the destruction up till terminateImpl is being called. Otherwise, if ref count drops to zero, then this object will be deleted immediately. We might be wanting to prevent that to happen.

dymk: Ideally, in a refcounted world, the Reference object saves us from any need to delay destruction. If nobody has a reference to us, we just by definition unreachable. I think this incRef/decRef was put here to be consistent with the other inc/decRefs (but still not needed). 

@lehecka looks good?

lehecka: looks good.

PR #652, opened 2017-08-17
Pass function objects by value and move them

phoad:
I have had crashes when I tried to use Single in multithreaded environment. With debugging and testing, I saw that if we don't do perfect forwarding properly we lose objects (the function pointer objects.

So I have replaced this code in Single.h
```
  void subscribe(Success&& next) {
    subscribe(SingleObservers::create<T>(next));
  }
```

with this
```
  void subscribe(Success next) {
    subscribe(SingleObservers::create<T>(std::move(next)));
  }
```
and it fixed the crash!

So I have applied the same to all locations where std::forward should have been used.

### DIFF
--- a/yarpl/include/yarpl/observable/Observable.h
+++ b/yarpl/include/yarpl/observable/Observable.h
@@ -40,8 +40,8 @@ class Observable : public virtual Refcounted {
       typename Next,
       typename =
           typename std::enable_if<std::is_callable<Next(T), void>::value>::type>
-  void subscribe(Next&& next) {
-    subscribe(Observers::create<T>(next));
+  void subscribe(Next next) {
+    subscribe(Observers::create<T>(std::move(next)));
   }
 
   /**
@@ -53,8 +53,9 @@ class Observable : public virtual Refcounted {
       typename = typename std::enable_if<
           std::is_callable<Next(T), void>::value &&
           std::is_callable<Error(std::exception_ptr), void>::value>::type>
-  void subscribe(Next&& next, Error&& error) {
-    subscribe(Observers::create<T>(next, error));
+  void subscribe(Next next, Error error) {
+    subscribe(Observers::create<T>(
+        std::move(next), std::move(error)));
   }
 
   /**
@@ -68,21 +69,24 @@ class Observable : public virtual Refcounted {
           std::is_callable<Next(T), void>::value &&
           std::is_callable<Error(std::exception_ptr), void>::value &&
           std::is_callable<Complete(), void>::value>::type>
-  void subscribe(Next&& next, Error&& error, Complete&& complete) {
-    subscribe(Observers::create<T>(next, error, complete));
+  void subscribe(Next next, Error error, Complete complete) {
+    subscribe(Observers::create<T>(
+        std::move(next),
+        std::move(error),
+        std::move(complete)));
   }
 
   template <typename OnSubscribe>
-  static auto create(OnSubscribe&&);
+  static auto create(OnSubscribe);
 
   template <typename Function>
-  auto map(Function&& function);
+  auto map(Function function);
 
   template <typename Function>
-  auto filter(Function&& function);
+  auto filter(Function function);
 
   template <typename Function>
-  auto reduce(Function&& function);
+  auto reduce(Function function);
 
   auto take(int64_t);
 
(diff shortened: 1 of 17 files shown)


yarpl/include/yarpl/observable/Observable.h
```diff
@@ -69,7 +69,7 @@ class Observable : public virtual Refcounted {
           std::is_callable<Error(std::exception_ptr), void>::value &&
           std::is_callable<Complete(), void>::value>::type>
   void subscribe(Next&& next, Error&& error, Complete&& complete) {
-    subscribe(Observers::create<T>(next, error, complete));
+    subscribe(Observers::create<T>(std::forward<Next>(next), std::forward<Error>(error), std::forward<Complete>(complete)));
```
dymk: clang-fmt?

phoad: yep :)

alexmalyshev:
While I agree this PR is an improvement, I wonder if an even better course of action is to take arguments by value and move them instead (i.e. `Next next;  flowable->subscribe(std::move(next));`).

We can totally get in trouble by accidentally taking something by reference we don't mean to with perfect forwarding, these methods feel like they should be taking ownership of whatever callable we're passing in.

yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -466,7 +466,7 @@ template <typename T, typename OnSubscribe>
 class FromPublisherOperator : public Observable<T> {
  public:
   explicit FromPublisherOperator(OnSubscribe&& function)
-      : function_(std::move(function)) {}
+      : function_(std::forward<OnSubscribe>(function)) {}
```
lehecka: this is not the case of perfect forwarding. Its moving of rvalue reference.

yarpl/include/yarpl/single/Single.h
```diff
@@ -103,7 +104,7 @@ class Single<void> : public virtual Refcounted {
   void subscribe(Success&& s) {
     class SuccessSingleObserver : public SingleObserver<void> {
      public:
-      SuccessSingleObserver(Success&& s) : success_{std::move(s)} {}
+      SuccessSingleObserver(Success&& s) : success_{std::forward<Success>(s)} {}
```
lehecka: this is not the case of perfect forwarding. Its moving of rvalue reference.

yarpl/include/yarpl/single/Single.h
```diff
@@ -144,7 +145,7 @@ template <typename T, typename OnSubscribe>
 class FromPublisherOperator : public Single<T> {
  public:
   explicit FromPublisherOperator(OnSubscribe&& function)
-      : function_(std::move(function)) {}
+      : function_(std::forward<OnSubscribe>(function)) {}
```
lehecka: this is not the case of perfect forwarding. Its moving of rvalue reference.

yarpl/include/yarpl/single/Single.h
```diff
@@ -158,7 +159,7 @@ template <typename OnSubscribe>
 class SingleVoidFromPublisherOperator : public Single<void> {
  public:
   explicit SingleVoidFromPublisherOperator(OnSubscribe&& function)
-      : function_(std::move(function)) {}
+      : function_(std::forward<OnSubscribe>(function)) {}
```
lehecka: this is not the case of perfect forwarding. Its moving of rvalue reference.

yarpl/include/yarpl/single/SingleOperator.h
```diff
@@ -134,7 +134,7 @@ template <typename T, typename OnSubscribe>
 class FromPublisherOperator : public Single<T> {
  public:
   explicit FromPublisherOperator(OnSubscribe&& function)
-      : function_(std::move(function)) {}
+      : function_(std::forward<OnSubscribe>(function)) {}
```
lehecka: this is not the case of perfect forwarding. Its moving of rvalue reference.

yarpl/src/yarpl/observable/Subscriptions.cpp
```diff
@@ -22,7 +22,7 @@ bool AtomicBoolSubscription::isCancelled() const {
  * Implementation that gets a callback when cancellation occurs.
  */
 CallbackSubscription::CallbackSubscription(std::function<void()>&& onCancel)
-    : onCancel_(std::move(onCancel)) {}
+    : onCancel_(std::forward<std::function<void()>>(onCancel)) {}
```
lehecka: this is not the case of perfect forwarding. Its moving of rvalue reference.

phoad:
I will revert the moving of rvalue reference lines ;)

PR #654, opened 2017-08-18
`M-x clang-format-buffer` on a few files

alexmalyshev:
Specifically ones that have been changed recently, and are now firing lint
errors on the FB internal linter.


lehecka:
we need to redo this since sources changed and merging is blocked now

alexmalyshev:
I don't think so?  At least GitHub isn't saying anything.  Pretty sure I'm rebased on current master right now.

PR #656, opened 2017-08-21
Add test for producing/consuming from a flowable in different threads

dymk:
Uses the ugly Flowable::create for now; we should figure out what it means to push into a subscriber from a different thread/how do we signal that the lambda passed to create should be invoked on a particular thread.

### DIFF
--- a/yarpl/test/FlowableTest.cpp
+++ b/yarpl/test/FlowableTest.cpp
@@ -1,8 +1,11 @@
 // Copyright 2004-present Facebook. All Rights Reserved.
 
+#include <gtest/gtest.h>
+#include <thread>
 #include <type_traits>
 #include <vector>
-#include <gtest/gtest.h>
+
+#include <folly/Baton.h>
 
 #include "yarpl/Flowable.h"
 #include "yarpl/flowable/TestSubscriber.h"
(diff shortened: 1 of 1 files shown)


yarpl/test/FlowableTest.cpp
```diff
@@ -1,8 +1,11 @@
 // Copyright 2004-present Facebook. All Rights Reserved.
 
+#include <gtest/gtest.h>
+#include <thread>
 #include <type_traits>
 #include <vector>
-#include <gtest/gtest.h>
+
+#include <folly/fibers/Baton.h>
```
alexmalyshev: Since there are no fibers involved, you can use `folly/Baton.h` instead.

yarpl/test/FlowableTest.cpp
```diff
@@ -400,5 +403,56 @@ TEST(FlowableTest, FlowableCompleteInTheMiddle) {
...
+  void onComplete() override {
+    Subscriber<int32_t>::onComplete();
+    EXPECT_EQ(total_, current_);
+    on_complete_.post();
+  }
+
+ private:
+  int32_t current_{0};
+  int32_t total_;
+  folly::fibers::Baton& on_complete_;
```
alexmalyshev: Nit: Please use camelCase for fields.

yarpl/test/FlowableTest.cpp
```diff
@@ -400,5 +403,56 @@ TEST(FlowableTest, FlowableCompleteInTheMiddle) {
...
+};
+
+TEST(FlowableTest, FlowableFromDifferentThreads) {
+  auto flowable = Flowable<int32_t>::create([&](auto& subscriber, int64_t req) {
+    EXPECT_GE(req, 10000);
+    auto t1 = std::thread([&] {
```
phoad: nit: I believe we use ScopedEventBaseThread in our examples. Would you mind updating the test to use that also?

yarpl/test/FlowableTest.cpp
```diff
@@ -400,5 +403,56 @@ TEST(FlowableTest, FlowableCompleteInTheMiddle) {
...
+      for (int32_t i = 0; i < req; i++) {
+        subscriber.onNext(i);
+      }
+
+      subscriber.onComplete();
```
phoad: Let's also handle the onError case as we have written this unit test. Add one more unit test that errors before completing the expected count of items.

yarpl/test/FlowableTest.cpp
```diff
@@ -400,5 +403,56 @@ TEST(FlowableTest, FlowableCompleteInTheMiddle) {
...
+    });
+    t1.join();
+    return std::make_tuple(req, true);
+  });
+
+  auto t2 = std::thread([&] {
+    folly::fibers::Baton on_flowable_complete;
+    flowable->subscribe(
+        yarpl::make_ref<RangeCheckingSubscriber>(10000, on_flowable_complete));
```
phoad: Can we also have one test that utilizes usage of Subscription?
The subscription->request(N) messages along with subscriber->onNext(..), to operate in coordination in multi-threaded environment.

yarpl/test/FlowableTest.cpp
alexmalyshev: Since there's no `EventBase`, I'd say `ScopedEventBaseThread` is unnecessary.

yarpl/test/FlowableTest.cpp
dymk: I think let's hold off on that until we've got a better constructor for flowables - it'll do us good to migrate away from the current `Flowable::create` ctor, which is sort of confusing and not very general.

yarpl/test/FlowableTest.cpp
dymk: (y)

yarpl/test/FlowableTest.cpp
```diff
@@ -400,5 +403,136 @@ TEST(FlowableTest, FlowableCompleteInTheMiddle) {
...
+    subscription->request(total_);
+  }
+
+  void onNext(int32_t val) override {
+    EXPECT_EQ(val, current_);
+    current_++;
+  }
+
+  void onError(std::exception_ptr) override {
+    EXPECT_TRUE(false) << "shouldn't call onError";
```
alexmalyshev: You can do `FAIL() << "message";`

yarpl/test/FlowableTest.cpp
```diff
@@ -400,5 +403,136 @@ TEST(FlowableTest, FlowableCompleteInTheMiddle) {
...
+    EXPECT_EQ(
+      yarpl::exceptionStr(err),
+      yarpl::exceptionStr(expectedErr_)
+    );
+    onError_.post();
+  }
+
+  void onComplete() override {
+    Subscriber<int32_t>::onComplete();
+    EXPECT_TRUE(false) << "shouldn't ever onComplete";
```
alexmalyshev: Ditto.

yarpl/test/FlowableTest.cpp
```diff
@@ -400,5 +403,136 @@ TEST(FlowableTest, FlowableCompleteInTheMiddle) {
...
+    });
+    t1.join();
+    return std::make_tuple<int64_t, bool>(expect, true);
+  });
+
+  auto t2 = std::thread([&] {
+    folly::Baton<> on_flowable_error;
+    flowable->subscribe(
+        yarpl::make_ref<ErrorRangeCheckingSubscriber>(expect, request, on_flowable_error, the_ex));
+    on_flowable_error.wait();
```
phoad: Will we have indefinite wait if we don't get any error?

yarpl/test/FlowableTest.cpp
```diff
@@ -400,5 +403,136 @@ TEST(FlowableTest, FlowableCompleteInTheMiddle) {
...
+    });
+    t1.join();
+    return std::make_tuple(req, true);
+  });
+
+  auto t2 = std::thread([&] {
+    folly::Baton<> on_flowable_complete;
+    flowable->subscribe(
+        yarpl::make_ref<RangeCheckingSubscriber>(count, on_flowable_complete));
+    on_flowable_complete.wait();
```
phoad: Add timed wait? In case the test fails, we should not wait indefinitely.

dymk:
alright GCC 4.9 is just being a pain now...

dymk:
there's a sort of ugly fix for gcc being such a pain; looks like name resolution is broken somehow

PR #657, opened 2017-08-21
Allowance persistence for cold resumption

somasun:
This PR changes how allowance is persisted for cold-resumption.  It currently changes only the consumer side allowance.  If folks are okay with the approach, I will make similar changes to the publisher side allowance.

- The allowance information for cold-resumption is now explicitly obtained from `StreamStateMachineBase` (instead of inferring it based on in/out frames).   For REQUEST_RESPONSE, the Requester will have a consumerAllowance of 1 if it is in a requested state, else 0.  For REQUEST_STREAM and REQUEST_CHANNEL, the consumerAllowance will the requestN sent minus the requestN fulfilled the from remote side.

- The allowance information and the in/out frames are now persisted with a single call to `ResumeManager`.  The persistence implementation could save both atomically.

- Earlier, the frames were persisted (event 1) before the application consumed the frame (event 2).  This could lead to application permanently losing the frame if there is a crash between the two events.  I have changed the design to persist frames after the application has consumed them.  This could potentially lead to application consuming some frames twice (but I felt this is better than application missing the frame).  Ideally the events should be in a single transaction.

Notes:
- There is a publisherAllowance and consumerAllowance for each stream since REQUEST_CHANNEL would need both and there should be a distinction.
- We could enforce all sent frames to have a valid StreamId (I don't know of a case where a frame can be without a StreamId).  Then we could get rid of the folly::Optional in trackSentFrame signature.

### DIFF
--- a/rsocket/ResumeManager.h
+++ b/rsocket/ResumeManager.h
@@ -33,14 +33,16 @@ class ResumeManager {
   // sent/received on the wire.  The application should implement a way to
   // store the sent and received frames in persistent storage.
   virtual void trackReceivedFrame(
-      const folly::IOBuf& serializedFrame,
+      size_t frameLength,
       FrameType frameType,
-      StreamId streamId) = 0;
+      StreamId streamId,
+      size_t consumerAllowance) = 0;
 
   virtual void trackSentFrame(
       const folly::IOBuf& serializedFrame,
       FrameType frameType,
-      folly::Optional<StreamId> streamIdPtr) = 0;
+      folly::Optional<StreamId> streamIdPtr,
+      size_t consumerAllowance) = 0;
 
   // We have received acknowledgement from the remote-side that it has frames
   // up to "position".  We can discard all frames before that.  This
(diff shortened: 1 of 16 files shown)


rsocket/ResumeManager.h
```diff
@@ -33,14 +33,16 @@ class ResumeManager {
   // sent/received on the wire.  The application should implement a way to
   // store the sent and received frames in persistent storage.
   virtual void trackReceivedFrame(
-      const folly::IOBuf& serializedFrame,
+      uint64_t frameLength,
```
alexmalyshev: Nit: Since this is a length, I think it should be `size_t`.

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -387,26 +387,29 @@ void RSocketStateMachine::processFrameImpl(
...
+    // during the time when we are resuming we are can't receive any other
+    // than connection level frames which drives the resumption
+    // TODO(lehecka): this assertion should be handled more elegantly using
+    // different state machine
     constexpr folly::StringPiece message{
         "Received stream frame while resuming"};
     LOG(ERROR) << message;
     closeWithError(Frame_ERROR::connectionError(message.str()));
-    return;
+  } else {
```
alexmalyshev: Don't we want to `return` after `closeWithError()`?

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -852,8 +854,16 @@ void RSocketStateMachine::outputFrame(std::unique_ptr<folly::IOBuf> frame) {
   stats_->frameWritten(frameType);
 
   if (isResumable_) {
+    size_t consumerAllowance = 0;
     auto streamIdPtr = frameSerializer_->peekStreamId(*frame);
-    resumeManager_->trackSentFrame(*frame, frameType, streamIdPtr);
+    if (streamIdPtr) {
+      auto it = streamState_.streams_.find(*streamIdPtr);
```
alexmalyshev: I've seen this pattern a few times now.  Maybe put a helper on `StreamState`?

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -1046,4 +1045,14 @@ bool RSocketStateMachine::ensureOrAutodetectFrameSerializer(
   frameSerializer_ = std::move(serializer);
   return true;
 }
+
+size_t RSocketStateMachine::getConsumerAllowance(StreamId streamId) const {
+  size_t consumerAllowance = 0;
+  auto it = streamState_.streams_.find(streamId);
+  if (it != streamState_.streams_.end()) {
```
alexmalyshev: Nit: `return it != streamState_.streams_.end() ? it->second->getConsumerAllowance() : 0;`

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -853,7 +851,10 @@ void RSocketStateMachine::outputFrame(std::unique_ptr<folly::IOBuf> frame) {
 
   if (isResumable_) {
     auto streamIdPtr = frameSerializer_->peekStreamId(*frame);
-    resumeManager_->trackSentFrame(*frame, frameType, streamIdPtr);
+    size_t consumerAllowance =
+        (streamIdPtr) ? getConsumerAllowance(*streamIdPtr) : 0;
```
alexmalyshev: Nit: Don't need the parens around `streamIdPtr`.

rsocket/statemachine/RequestResponseRequester.cpp
```diff
@@ -116,15 +116,7 @@ void RequestResponseRequester::handlePayload(
...
-//  }
-//}
-//
-// void RequestResponseRequester::resumeStream(RequestHandler& requestHandler) {
-//  if (consumingSubscriber_) {
-//    requestHandler.onSubscriberResumed(consumingSubscriber_);
-//  }
-//}
+size_t RequestResponseRequester::getConsumerAllowance() const {
+  return (state_ == State::REQUESTED) ? 1 : 0;
```
alexmalyshev: Nit: You can do `return state_ == State::REQUESTED`;

PR #662, opened 2017-08-22
Replace std::exception_ptr with folly::exception_wrapper

phoad:
Previously we didn't want to depend on folly and use the standard exception_ptr. 
This requirement is not valid anymore, 
so it should be ok to replace std::exception_ptr with folly::exception_wrapper.


lehecka:
thanks for this big change. Let's bring this to fbcode asap.

PR #663, opened 2017-08-22
Returning ConnectionException for failed setup

lehecka:


rsocket/RSocketServiceHandler.cpp
```diff
@@ -27,7 +27,12 @@ std::shared_ptr<RSocketServiceHandler> RSocketServiceHandler::create(
     explicit ServiceHandler(OnNewSetupFn fn) : onNewSetupFn_(std::move(fn)) {}
     folly::Expected<RSocketConnectionParams, RSocketException> onNewSetup(
         const SetupParameters& setupParameters) override {
-      return RSocketConnectionParams(onNewSetupFn_(setupParameters));
+      try {
+        return RSocketConnectionParams(onNewSetupFn_(setupParameters));
+      } catch(const std::exception& e) {
```
phoad: nit: let's catch everything? ```...```

alexmalyshev: It's usually fine to go with catching `std::exception` as the "catch-all" clause.  Anyone throwing integers is just asking for trouble.

PR #664, opened 2017-08-22
Remove lock and buffering of input in FrameTransport

lehecka:
removing the last bits of extra logic from FrameTransport before I remove the class all together.

* removed mutex from the class. Locking is not necessary, we support single threaded access to/from duplex connection. The RSocketStateMachine and DuplexConnection are running on the same thread.
* removed input buffering. Since everything is synchronous, there is no need to provide buffering for situations when FrameTransport is connected and the FrameProcessor is not attached. This used to happen because we were potentially resuming on another thread and there could a race there. There is no race anymore.

I tested the change with all unit tests and I also ran all of our examples + resumption.

### DIFF
--- a/rsocket/framing/FrameTransport.cpp
+++ b/rsocket/framing/FrameTransport.cpp
@@ -23,9 +23,7 @@ FrameTransport::~FrameTransport() {
 }
 
 void FrameTransport::connect() {
-  Lock lock(mutex_);
-
-  DCHECK(connection_);
+  CHECK(connection_);
 
   if (connectionOutput_) {
     // Already connected.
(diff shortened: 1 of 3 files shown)


rsocket/framing/FrameTransport.cpp
```diff
@@ -102,8 +94,6 @@ void FrameTransport::closeImpl(folly::exception_wrapper ew) {
 }
 
 void FrameTransport::onSubscribe(yarpl::Reference<Subscription> subscription) {
-  Lock lock(mutex_);
-
   if (!connection_) {
```
phoad: Should we cancel the subscription in this case?

rsocket/framing/FrameTransport.cpp
```diff
@@ -164,8 +143,6 @@ void FrameTransport::cancel() {
 }
 
 void FrameTransport::outputFrameOrDrop(std::unique_ptr<folly::IOBuf> frame) {
```
phoad: This class doesn't know if dropping the frame is a critical error or not, but a higher logic might wish to take action if frames get dropped. 
Might it be beneficial to return a bool, stating if the frame is dropped or processed?

lehecka: Agree with you. boolean as a return value would be good. However this class is going away so the caller will have to implement the logic if it needs to.

rsocket/framing/FrameTransport.cpp
lehecka: good point! we should. But I am going to remove this class instead. ;)

PR #665, opened 2017-08-22
removing EventBase from RSocketStateMachine

lehecka:
RSocketStateMachine no longer runs on a different EventBase than its DuplexConnection.
There is no need to have EventBase in the RSocketStateMachine. In fact it was a perf issue since we were scheduling processing of every frame on the EventBase.

I removed the eventBase.

The PR is not ready to land until https://github.com/rsocket/rsocket-cpp/pull/664 is landed.


lehecka:
I tested the change with all of our examples + warm resumption example.
The cold resumption is not working because of the TODO left in the code. Mani is working in this area now.

rsocket/RSocketServer.cpp
```diff
@@ -135,7 +135,6 @@ void RSocketServer::onRSocketSetup(
   auto responder = std::make_shared<ScheduledRSocketResponder>(
       std::move(connectionParams.responder), *eventBase);
   auto rs = std::make_shared<RSocketStateMachine>(
-      *eventBase,
```
phoad: nit: SetupResumeAcceptor holds an instance of `eventBase`. 
Instead of getting the `eventbase` using the `getExistingEventBase` function, we might make `SetupResumeAcceptor` to call `onSetup` function with the `eventBase` that it holds. 
It might remove one more ambiguity of which event base do we get from the EventBaseManager, which is clear that it is the event base of the io thread, but..

rsocket/RSocketServer.cpp
```diff
@@ -162,7 +161,7 @@ void RSocketServer::onRSocketResume(
   }
   auto serverState = std::move(result.value());
   CHECK(serverState);
-  if (!serverState->rSocketStateMachine_->isInEventBaseThread()) {
+  if (!serverState->eventBase_.isInEventBaseThread()) {
     throw RSocketException("Trying to RESUME in a different worker thread");
```
phoad: nit: LOG(ERROR)?

s/`worker thread`/`io thread`/  ?

rsocket/RSocketServer.cpp
lehecka: yep, that is a good suggestion. I saw that too. I will address that in another diff.

rsocket/RSocketServer.cpp
lehecka: this is a temporary exception. it will be removed.

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -495,9 +465,12 @@ void RSocketStateMachine::handleConnectionFrame(
           auto streamToken = resumeManager_->getStreamToken(streamId);
           auto subscriber = coldResumeHandler_->handleRequesterResumeStream(
               streamToken, consumerAllowance);
+          //TODO(somatsun): ensure that subscription is called from the correct
```
somasun: Can you use `EventBaseManager::get()->getCurrentEventBase()` and use it here directly to create `ScheduledSubscriptionSubscriber` instead ?

dymk:
squash commits?

PR #666, opened 2017-08-22
remove manual refcounting in EmitterWrapper

dymk:


yarpl/include/yarpl/flowable/EmitterFlowable.h
```diff
@@ -36,17 +36,9 @@ class EmiterSubscription : private Subscription, private Subscriber<T> {
...
       : emiter_(std::move(emiter)), subscriber_(std::move(subscriber)) {
-    // We expect to be heap-allocated; until this subscription finishes
-    // (is canceled; completes; error's out), hold a reference so we are
-    // not deallocated (by the subscriber).
-    Refcounted::incRef(*this);
     subscriber_->onSubscribe(get_ref<Subscription>(this));
   }
 
-  virtual ~EmiterSubscription() {
-    subscriber_.reset();
```
dymk: removed as this is redundant code; subscriber_ is reset when ~Reference is called

yarpl/include/yarpl/flowable/EmitterFlowable.h
```diff
@@ -164,8 +161,12 @@ class EmiterSubscription : private Subscription, private Subscriber<T> {
       int64_t emitted;
       bool done;
 
-      std::tie(emitted, done) =
-          emiter_->emit(get_ref<Subscriber<T>>(this), current);
+      std::tie(emitted, done) = emiter_->emit(this_subscriber, current);
+
+      if(this->count() == 1) {
```
phoad: We have one more while/true loop below. We might put this check to there? Are we sure that we will not stuck in that loop?

yarpl/include/yarpl/flowable/EmitterFlowable.h
```diff
@@ -208,7 +208,7 @@ class EmitterWrapper : public EmiterBase<T>, public Flowable<T> {
       : emitter_(std::move(emitter)) {}
 
   void subscribe(Reference <Subscriber<T>> subscriber) override {
-    new EmiterSubscription<T>(get_ref(this), std::move(subscriber));
+    make_ref<EmiterSubscription<T>>(get_ref(this), std::move(subscriber));
```
phoad: nit: do we need to create a reference? It will be immediately deleted already.

dymk: The code here is functionally identical; the reason EmitterSubscription doesn't get immediately destroyed is because we call `subscriber->onSubscribe(make_ref(this))` in `EmitterSubscription`'s ctor (so the subscriber keeps a reference to it)

yarpl/include/yarpl/flowable/EmitterFlowable.h
dymk: The `while` loop is just there to satisfy a race around the compare&swap with the `requested_` member, if multiple threads have a handle to us and are requesting (eg count() > 1). So this is safe to do here rather than the while loop, as if count() == 1, we'll only iterate through the loop once (also, we should be done/canceled if nobody has a reference to us, else it's a bug).

yarpl/include/yarpl/flowable/EmitterFlowable.h
```diff
@@ -164,8 +161,12 @@ class EmiterSubscription : private Subscription, private Subscriber<T> {
       int64_t emitted;
       bool done;
 
-      std::tie(emitted, done) =
-          emiter_->emit(get_ref<Subscriber<T>>(this), current);
+      std::tie(emitted, done) = emiter_->emit(this_subscriber, current);
+
+      if(this->count() == 1) {
+        assert(requested_.load(std::memory_order_relaxed) == kCanceled);
```
alexmalyshev: Can these be `DCHECK_EQ` and `DCHECK`?  I dislike using `assert` directly because it tends to result in unused variables and broken opt builds.

yarpl/include/yarpl/flowable/EmitterFlowable.h
```diff
@@ -26,25 +26,20 @@ class EmiterBase : public virtual Refcounted {
  * This is synchronous: the emit calls are triggered within the context
  * of a request(n) call.
  */
-template<typename T>
-class EmiterSubscription : private Subscription, private Subscriber<T> {
+template <typename T>
+class EmiterSubscription : public virtual Subscription,
```
phoad: It doesn't seem like we need `virtual` keyword in here. https://en.wikipedia.org/wiki/Virtual_inheritance

yarpl/include/yarpl/flowable/EmitterFlowable.h
```diff
@@ -26,25 +26,20 @@ class EmiterBase : public virtual Refcounted {
...
-    subscriber_->onSubscribe(get_ref<Subscription>(this));
-  }
+      Reference<EmiterBase<T>> emiter,
+      Reference<Subscriber<T>> subscriber)
+      : emiter_(std::move(emiter)), subscriber_(std::move(subscriber)) {}
 
-  virtual ~EmiterSubscription() {
-    subscriber_.reset();
+  void init() {
+    subscriber_->onSubscribe(get_ref<Subscription>(this));
```
phoad: what's the crash if you place this line into the constructor?

dymk: Refcounting breaks for still-to-be-determined reasons; it seems bumping the refcount of ourselves in the ctor leads to us destroying ourselves in the ctor when the reference created there is destroyed (even if someone else holds a reference to us by the time we reach the end of the ctor).

lehecka: Dylan, can you elaborate a little more? I don't understand your comment. You are saying that by calling incRef in the ctor leads to calling dtor at the end of the ctor? That doesn't make sense to me. Why we didn't see this in the current code where we call onSubscribe from the ctor?

dymk: Oh I see what's happening - similar problem to what was going on in `process`. There's a refcount of 0 on entry to the ctor, then we pass a reference to ourselves to `subscriber_->onSubscribe(get_ref(this))`, which doesn't keep a reference to `this` in some cases past the `onSubscribe` call, and once the temporary `get_ref` passed to `onSubscribe` gets destroyed, we call our dtor (while we're still in the ctor). 

Visually, if the onSubscribe is kept in the ctor: 
```cpp
  EmiterSubscription(
      Reference<EmiterBase<T>> emiter,
      Reference<Subscriber<T>> subscriber)
      : emiter_(std::move(emiter)), subscriber_(std::move(subscriber)) {
        // count() == 0
        auto ref = get_ref<Subscription>(this);
        // count() == 1
        subscriber_->onSubscribe(ref);
        // count() == 0, dtor called while we're in the ctor
      }
```

So the solution here is to defer calling `onSubscribe` until we're out of the ctor for EmitterSubscription (by moving it into an `init()` method).

phoad: The problem might be that, which is similar to shared_ptr's.
http://en.cppreference.com/w/cpp/memory/enable_shared_from_this
Check this link. In there it says that, if we call `shared_from_this`, from an instance which is not a shared_ptr, it causes undefined behavior.

So I need to ask, if we call get_ref<T>(this), but if there is no Reference<T> outside of `this`, then if we call get_ref - increase count to 1, when we lose this ref, count goes to 0, which deletes the object.

So the problem is somewhere we create an instance of the Emitter without a reference! Which is the constructor of the Emitter already. Namely even if we have Reference<Emitter>, at the constructor of the Emitter, there is nothing that is incrementing the count of the reference. 

I believe the inc_count and dec_count in the code in other places were specifically for this purpose. So we need inc_cnt / dec_cnt in here too. Perhaps we need to revert the old update that has removed these two calls from another class.

phoad: It looks like either we increment count in the constructor and decrement it in the release function, or we first create an instance of Reference<T> and later pass this instance to another object. Passing an instance of itself, get_ref(this), in the constructor and passing it around causes the object to get deleted. To prevent this they were calling inc_ref.

This is an improvement on that at least and I believe an improvement is good to check in :)

https://github.com/rsocket/rsocket-cpp/issues/671

dymk:
Changes to Reference itself won't fix this; they'll just make the location of the dangerous operation obvious. A change that might work is, inside of `make_ref`, create a new object with a refcount of 1 rather than 0, wait for the ctor to finish, then manually decrement the refcount and move the pointer into a Reference object and return. But this means that constructors now behave differently than normal methods on a class, which might also result in an object deleting itself (we can't provide automatic protection in those cases by modifying Reference).

PR #667, opened 2017-08-22
Decrease the connection count in ConnectManyAsync test

phoad:
Opening 1000 connections is high number which might consume system resources and cause failure. Even when a socket gets closed, it might take ~3 secs for system to mark that port is closed, so we might get too many files in use error.

So I have decreased it to 100 and to increase the parallelism I have made 10 threads only that are created at the start, not per connection request.

### DIFF
--- a/test/RSocketClientServerTest.cpp
+++ b/test/RSocketClientServerTest.cpp
@@ -3,6 +3,7 @@
 #include "RSocketTests.h"
 
 #include <folly/io/async/ScopedEventBaseThread.h>
+#include <folly/Random.h>
 #include <gtest/gtest.h>
 #include "test/handlers/HelloStreamRequestHandler.h"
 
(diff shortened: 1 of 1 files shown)


test/RSocketClientServerTest.cpp
```diff
@@ -36,15 +37,16 @@ TEST(RSocketClientServer, ConnectManySync) {
...
-  std::vector<std::unique_ptr<folly::ScopedEventBaseThread>> workers(count);
+  size_t count = 100;
+  size_t workerCount = 10;
+  std::vector<folly::ScopedEventBaseThread> workers(workerCount);
   std::vector<folly::Future<std::shared_ptr<RSocketClient>>> clients;
 
   std::atomic<int> executed{0};
   for (size_t i = 0; i < count; ++i) {
-    workers[i] = std::make_unique<folly::ScopedEventBaseThread>();
+    int workerId = folly::Random::rand32(workerCount);
```
dymk: it'll only use about 63% of the scoped event bases for moderatly large count (probability of selecting every id exactly once is pretty slim)
probably not a good idea to have nondeterminism in tests in general, too

dymk:
LGTM

test/RSocketClientServerTest.cpp
```diff
@@ -36,15 +37,16 @@ TEST(RSocketClientServer, ConnectManySync) {
 TEST(RSocketClientServer, ConnectManyAsync) {
   auto server = makeServer(std::make_shared<HelloStreamRequestHandler>());
 
-  size_t count = 1000;
-  std::vector<std::unique_ptr<folly::ScopedEventBaseThread>> workers(count);
+  size_t count = 100;
```
alexmalyshev: Nit: `constexpr` or `const` for both `size_t` vars.  I'd rename `count` to `connectionCount` or something like it to make things clearer.

PR #668, opened 2017-08-23
Test for cold resumption

somasun:
This tests client-originated REQUEST_STREAMs for cold-resumption.

Three streams are created from the client and two cold-resumptions happen in-between.

The test verifies that the client gets all the frames in the right order across these two resumptions.  It remembers the last received frame for each stream before resumption and verifies that the incoming frames after resumption are in expected ones.

### DIFF
--- a/examples/resumption/ColdResumption_Client.cpp
+++ b/examples/resumption/ColdResumption_Client.cpp
@@ -25,10 +25,10 @@ namespace {
 class HelloSubscriber : public virtual Refcounted, public Subscriber<Payload> {
  public:
   void request(int n) {
-    while (!subscription_) {
+    while (!Subscriber<Payload>::subscription()) {
       std::this_thread::yield();
     }
-    subscription_->request(n);
+    Subscriber<Payload>::subscription()->request(n);
   }
 
   int rcvdCount() const {
@@ -37,7 +37,7 @@ class HelloSubscriber : public virtual Refcounted, public Subscriber<Payload> {
 
  protected:
   void onSubscribe(Reference<Subscription> subscription) noexcept override {
-    subscription_ = subscription;
+    Subscriber<rsocket::Payload>::onSubscribe(subscription);
   }
 
   void onNext(Payload) noexcept override {
@@ -45,7 +45,6 @@ class HelloSubscriber : public virtual Refcounted, public Subscriber<Payload> {
   }
 
  private:
-  Reference<Subscription> subscription_{nullptr};
   std::atomic<int> count_{0};
 };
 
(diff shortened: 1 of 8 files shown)


examples/resumption/ColdResumption_Client.cpp
```diff
@@ -44,6 +44,14 @@ class HelloSubscriber : public virtual Refcounted, public Subscriber<Payload> {
     count_++;
   }
 
+  void onError(folly::exception_wrapper) override {
```
alexmalyshev: Not this diff but, if you call `Subscriber<Payload>::onSubscribe(sub)` inside of `HelloSubscriber::onSubscribe(sub)`, then you won't need these overloads, and you'll be able to access `sub` via `Subscriber<Payload>::subscription()`.

test/ColdResumptionTest.cpp
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  void onNext(Payload p) noexcept override {
+    auto currValue = folly::to<size_t>(p.data->moveToFbString().toStdString());
+    EXPECT_EQ(latestValue_, currValue - 1);
+    latestValue_ = currValue;
+    count_++;
+  }
+
+  void onError(folly::exception_wrapper) override {
```
alexmalyshev: Same comment as on `HelloSubscriber`.

test/RSocketTests.cpp
```diff
@@ -3,12 +3,19 @@
...
 
 namespace rsocket {
 namespace tests {
 namespace client_server {
 
+std::unique_ptr<TcpConnectionFactory> getConnFactory(
+    folly::EventBase* eventBase,
+    uint16_t port) {
+  folly::SocketAddress address;
+  address.setFromHostPort("localhost", port);
```
alexmalyshev: You can do `folly::SocketAddress address{"::1", port};`

test/ColdResumptionTest.cpp
```diff
@@ -0,0 +1,220 @@
...
+      std::this_thread::yield();
+    }
+    Subscriber<Payload>::subscription()->request(n);
+  }
+
+  void awaitLatestValue(size_t value) {
+    auto count = 3;
+    while (value != latestValue_ && count > 0) {
+      VLOG(1) << "Wait for " << count << " seconds for latest value";
+      sleep(1);
```
alexmalyshev: This seems like it can be done with a `folly::Baton<>` and timed wait...

examples/resumption/ColdResumption_Client.cpp
```diff
@@ -25,10 +25,10 @@ namespace {
 class HelloSubscriber : public virtual Refcounted, public Subscriber<Payload> {
  public:
   void request(int n) {
-    while (!subscription_) {
+    while (!Subscriber<Payload>::subscription()) {
```
alexmalyshev: This can be `folly::Baton<>::wait()`, and `onSubscribe()` can do `folly::Baton<>::post()`.

PR #672, opened 2017-08-23
`get_ref` should never be called for an instance of reference count of zero

phoad:
https://github.com/rsocket/rsocket-cpp/issues/671

This will prevent buggy code to happen.

Next, we should also properly implement make_ref to let calling get_ref to be possible at the inside of the constructor.

### DIFF
--- a/yarpl/include/yarpl/Refcounted.h
+++ b/yarpl/include/yarpl/Refcounted.h
@@ -273,13 +273,26 @@ Reference<T> make_ref(Args&&... args) {
   return Reference<T>(new T(std::forward<Args>(args)...));
 }
 
-template <typename T>
+template <
+    typename T,
+    typename =
+    std::enable_if_t<std::is_base_of<Refcounted, std::decay_t<T>>::value>>
 Reference<T> get_ref(T& object) {
+  assert(
+      object.count() > 0 &&
+          "get_ref should not be called from a non referenced object");
+
   return Reference<T>(&object);
 }
 
-template <typename T>
+template <
+    typename T,
+    typename =
+    std::enable_if_t<std::is_base_of<Refcounted, std::decay_t<T>>::value>>
 Reference<T> get_ref(T* object) {
+  assert(
+      object->count() > 0 &&
+          "get_ref should not be called from a non referenced object");
   return Reference<T>(object);
 }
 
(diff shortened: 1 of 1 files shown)


yarpl/include/yarpl/Refcounted.h
```diff
@@ -273,13 +273,26 @@ Reference<T> make_ref(Args&&... args) {
   return Reference<T>(new T(std::forward<Args>(args)...));
 }
 
-template <typename T>
+template <
+    typename T,
+    typename =
+    std::enable_if_t<std::is_base_of<Refcounted, std::decay_t<T>>::value>>
 Reference<T> get_ref(T& object) {
+  assert(
```
alexmalyshev: `DCHECK_GT(object.count(), 0) << "get_ref should not...";`

yarpl/include/yarpl/Refcounted.h
```diff
@@ -273,13 +273,26 @@ Reference<T> make_ref(Args&&... args) {
   return Reference<T>(new T(std::forward<Args>(args)...));
 }
 
-template <typename T>
+template <
+    typename T,
+    typename =
+    std::enable_if_t<std::is_base_of<Refcounted, std::decay_t<T>>::value>>
```
alexmalyshev: I'd do this as a `static_assert()` inside the method personally.

yarpl/include/yarpl/Refcounted.h
phoad: DCHECK_GT depends on including gtest related header. I did not want to include that in this header as this is a low level header.
We all use `assert` in this file.

yarpl/include/yarpl/Refcounted.h
phoad: Why static_assert ?

alexmalyshev: Cuts down on unnecessary template params.  They can get quite verbose when looking at stack traces 😄 .

yarpl/include/yarpl/Refcounted.h
alexmalyshev: Technically it's glog and not gtest, so it's not as heavyweight.

I generally dislike using `assert()` as it tends to break opt builds with that have `-Wno-unused-variable` set.

yarpl/include/yarpl/Refcounted.h
phoad: Oo, you would like to also prevent any overloaded one to have the error too.

PR #673, opened 2017-08-23
check that new references are also held somewhere else

dymk:
Eventually disallow `Reference<T>(new T())`

### DIFF
--- a/yarpl/include/yarpl/Refcounted.h
+++ b/yarpl/include/yarpl/Refcounted.h
@@ -11,6 +11,11 @@
 
 namespace yarpl {
 
+namespace detail {
+struct skip_initial_refcount_check {};
+struct do_initial_refcount_check {};
+}
+
 /// Base of refcounted objects.  The intention is the same as that
 /// of boost::intrusive_ptr<>, except that we have virtual methods
 /// anyway, and want to avoid argument-dependent lookup.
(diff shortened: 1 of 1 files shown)


yarpl/include/yarpl/Refcounted.h
```diff
@@ -9,8 +9,15 @@
 #include <type_traits>
 #include <utility>
 
+#include <glog/logging.h>
```
phoad: Let's do pure assert.

yarpl/include/yarpl/Refcounted.h
```diff
@@ -270,17 +306,26 @@ bool operator>=(std::nullptr_t, const Reference<T>& rhs) noexcept {
...
+  return Reference<T>(
+    new T(std::forward<Args>(args)...),
+    detail::skip_initial_refcount_check{}
+  );
 }
 
 template <typename T>
 Reference<T> get_ref(T& object) {
-  return Reference<T>(&object);
+  return Reference<T>(
```
phoad: Merge the `enable_if` part in the other diff to here. 
We should not be calling `get_ref` on a type T, if T is not a Refcounted object.

phoad:
```Reference<T>(new T())```
If we move the instance of T into the Reference object then this should be ok, isn't it?
So the constructor of the Reference<T> should move the object not reference it.


I have thought a bit about the issue and I have a proposal about the solution:
- Setting the count to 1 at the constructor of Refcounted and setting it to 0 at the destructor.
- Moving the T object at the constructor of the Reference object
- Already the assignment operator or copy constructor of the Reference is incrementing the count.

Why:

At the constructor of Refcounted, we can set the initial count to 1. At the destructor of it we can set the counter to 0, as it is getting destroyed.
So if we don't reference an object
```
{
T t1(3);
t1->foo();
}
```
it will be pure construction and destruction.

If we create reference to it, we will move it:
```
{
Reference<T> t1(new T());
t1->foo();
}
```
as we move it, the count will still be 1.

But if we create another reference from the old reference, then count should increase
```
Reference<T> t2 = t1;
delete t1; // t2 still references
```

- If we call get_ref, even if we are at the constructor, the count was 1, it will be incremented to 2. Ok!
- If we call make_ref(T*), we should still move the T, which will keep the count value still at 1.

File: yarpl/include/yarpl/Refcounted.h [hunk 3]
```diff
@@ -66,9 +71,40 @@ class Reference {
...
+        // if do_something_with(r) doens't keep a reference to r somewhere, then
+        // count() == 0
+        // and we call ~MyClass() within the constructor, which is a Bad Thing
+       }
+     };
+
+     * the check below prevents (at runtime) taking a reference in situations
+     * like this
+     */
+    assert(
```
phoad: If the count is 0, there should be no such object already.

yarpl/include/yarpl/Refcounted.h
```diff
@@ -11,6 +11,11 @@
 
 namespace yarpl {
 
+namespace detail {
```
phoad: I don't see any needs for this struct.

PR #675, opened 2017-08-24
Fix hang when destroying RSocketConnectionManager off of the client IO thread

alexmalyshev:
Or really, any IO thread that's managing one of the state machines inside the
connection manager.

To do so, the thread running ~RSocketConnectionManager needs to close the state
machines using it synchronously.  For that, it has to relinquish the
connections map lock.  We'll move out all connections from the map first, then
close them all (either sync or async).  Then we'll use the moral equivalent of a
latch (atomic_int + baton) to determine when all of them have actually closed.

### DIFF
--- a/rsocket/internal/RSocketConnectionManager.cpp
+++ b/rsocket/internal/RSocketConnectionManager.cpp
@@ -2,9 +2,9 @@
 
 #include "rsocket/internal/RSocketConnectionManager.h"
 
-#include <folly/io/async/EventBase.h>
-#include <folly/ScopeGuard.h>
 #include <folly/ExceptionWrapper.h>
+#include <folly/ScopeGuard.h>
+#include <folly/io/async/EventBase.h>
 
 #include "rsocket/RSocketConnectionEvents.h"
 #include "rsocket/statemachine/RSocketStateMachine.h"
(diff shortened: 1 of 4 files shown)


rsocket/internal/RSocketConnectionManager.cpp
```diff
@@ -2,42 +2,71 @@
...
       return;
     }
 
-    shutdown_.emplace();
+    shutdownBaton_.emplace();
+    shutdownCounter_ = locked->size();
+
+    VLOG(2) << "Need to close " << shutdownCounter_.load() << " connections";
+
+    map.swap(*locked);
```
somasun: It took some time to wrap my head around all the changes.  Have you considered just making a copy of the map (instead of clearing `locked`).  Then you wouldn't need the complexity in `removeConnection()`.

rsocket/internal/RSocketConnectionManager.cpp
```diff
@@ -108,12 +132,19 @@ void RSocketConnectionManager::manageConnection(
 void RSocketConnectionManager::removeConnection(
     const std::shared_ptr<RSocketStateMachine>& socket) {
   auto locked = sockets_.lock();
-  locked->erase(socket);
+  auto const result = locked->erase(socket);
+  DCHECK_LE(result, 1);
```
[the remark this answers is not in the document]
alexmalyshev: It should never fail.  This is me asserting that the value returned from `erase` will always be 0 or 1.

rsocket/internal/RSocketConnectionManager.cpp
alexmalyshev: I don't like making a copy of the map when I can do a move instead.

I don't think `removeConnection` is that complex.  I could make it even simpler if we had a Latch implementation in the `rsocket/` directory (I have one in `benchmarks/` but I meant it for testing only).

PR #676, opened 2017-08-24
No more `new`

dymk:
Changes instances of `new Refcounted` to `make_ref<Refcounted>`, to prepare for removing the `Reference(T*)` public API. With this we can start reducing the surface area for raw pointers held by `Reference`s to be leaked.

### DIFF
--- a/yarpl/include/yarpl/Refcounted.h
+++ b/yarpl/include/yarpl/Refcounted.h
@@ -11,6 +11,11 @@
 
 namespace yarpl {
 
+namespace detail {
+struct skip_initial_refcount_check {};
+struct do_initial_refcount_check {};
+}
+
 /// Base of refcounted objects.  The intention is the same as that
 /// of boost::intrusive_ptr<>, except that we have virtual methods
 /// anyway, and want to avoid argument-dependent lookup.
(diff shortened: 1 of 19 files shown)


yarpl/include/yarpl/Refcounted.h
```diff
@@ -66,9 +71,40 @@ class Reference {
   Reference() = default;
   inline /* implicit */ Reference(std::nullptr_t) {}
 
+  // TODO: Move check for count() into this constructor
+  // TODO: Remove public ctor for Reference(T*)
   explicit Reference(T* pointer) : pointer_(pointer) {
     inc();
   }
+  explicit Reference(T* pointer, detail::skip_initial_refcount_check)
```
alexmalyshev: Are these ctors meant to be private?  If so then I approve of the `detail::` namespace.  But if they're meant to be private to `Reference` then I think we shouldn't use the namespace.

yarpl/include/yarpl/Refcounted.h
```diff
@@ -66,9 +71,40 @@ class Reference {
...
+     * consider the following:
+     *
+     class MyClass : Refcounted {
+       MyClass() {
+        // count() == 0
+        auto r = get_ref<MyClass>(this)
+        // count() == 1
+        do_something_with(r);
+        // if do_something_with(r) doens't keep a reference to r somewhere, then
+        // count() == 0
```
alexmalyshev: Nit: This is a bit confusing, as this point `count() == 1` because `r` is still in scope.  I get what you mean though, `r` will go out of scope at the end of the ctor, decrement the count to 0, and then run the dtor.

phoad: I still believe that we need to start the count of Refcounted from 1 and don't do any increment in here.

phoad:
It doesn't 100 percent match to what I was imagined, though it doesn't mean we should not do that way. I let @alexmalyshev  and you to decide what to do in here.

yarpl/include/yarpl/Refcounted.h
dymk: Yep, private ctors

PR #677, opened 2017-08-24
Enable calling `get_ref` inside of constructors without fear of deallocation

dymk:
What's on the tin

### DIFF
--- a/yarpl/include/yarpl/flowable/EmitterFlowable.h
+++ b/yarpl/include/yarpl/flowable/EmitterFlowable.h
@@ -25,7 +25,7 @@ class EmiterBase : public virtual Refcounted {
  * of a request(n) call.
  */
 template <typename T>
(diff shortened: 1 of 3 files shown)


yarpl/include/yarpl/flowable/EmitterFlowable.h
```diff
@@ -203,7 +203,7 @@ class EmitterWrapper : public EmiterBase<T>, public Flowable<T> {
   explicit EmitterWrapper(Emitter emitter) : emitter_(std::move(emitter)) {}
 
   void subscribe(Reference<Subscriber<T>> subscriber) override {
-    new EmiterSubscription<T>(get_ref(this), std::move(subscriber));
+    make_ref<EmiterSubscription<T>>(get_ref(this), std::move(subscriber));
```
dymk: this leaks memory otherwise, which is why this change is here

PR #679, opened 2017-08-25
Allow cold-resume even if the old transport closure is delayed

somasun:
If the client cold-resumes quickly (before the old connection to the server is properly terminated), then the cold-resume fails `CHECK(isClosedOrDisconnected())` in `resumeServer()`.  This change proactively closes the old transport (if it is still open) on cold-resumption at the server.

### DIFF
--- a/rsocket/statemachine/RSocketStateMachine.cpp
+++ b/rsocket/statemachine/RSocketStateMachine.cpp
@@ -88,6 +88,10 @@ void RSocketStateMachine::connectServer(
 bool RSocketStateMachine::resumeServer(
     yarpl::Reference<FrameTransport> frameTransport,
     const ResumeParameters& resumeParams) {
+  if (!isDisconnectedOrClosed()) {
+    disconnect(folly::exception_wrapper(
+        std::runtime_error("A new transport is resuming the session")));
+  }
   CHECK(connect(
       std::move(frameTransport), resumeParams.protocolVersion));
   return resumeFromPositionOrClose(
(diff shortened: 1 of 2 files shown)


rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -88,6 +88,10 @@ void RSocketStateMachine::connectServer(
 bool RSocketStateMachine::resumeServer(
     yarpl::Reference<FrameTransport> frameTransport,
     const ResumeParameters& resumeParams) {
+  if (!isDisconnectedOrClosed()) {
```
alexmalyshev: How do we hit this?  Shouldn't `resumeServer()` be scheduled after the termination, and on the same thread?

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -88,6 +88,10 @@ void RSocketStateMachine::connectServer(
...
+    disconnect(folly::exception_wrapper(
```
alexmalyshev: Nit: You don't need the `folly::exception_wrapper()` here, the ctor is implicit.

rsocket/statemachine/RSocketStateMachine.cpp
somasun: This could happen if remote side did not terminate gracefully (did not send a FYN in TCP's case) and tried to cold-resume before the local-side's network stack closed the old connection (in TCP's case the connection can remain half-open forever).  In this case we want to forcefully terminate the old connection/transport and use the new one being used for resumption.

This could be exploited if a client is already connected and another client tries to resume with the same ResumeID (but that is very unlikely)

PR #681, opened 2017-08-25
Add DCHECKs on base Subscriber classes to ensure we're using them cor…

alexmalyshev:
…rectly

Stuff like making sure we call onSubscribe() before on{Next,Complete,Error}().
Also be sure to cancel subscriptions in onSubscribe() if we've already got a
subscription running.

### DIFF
--- a/yarpl/include/yarpl/flowable/Subscriber.h
+++ b/yarpl/include/yarpl/flowable/Subscriber.h
@@ -2,11 +2,12 @@
 
 #pragma once
 
-#include <folly/ExceptionWrapper.h>
-#include <stdexcept>
-
 #include "yarpl/Refcounted.h"
 #include "yarpl/flowable/Subscription.h"
+
+#include <folly/ExceptionWrapper.h>
+
+#include <glog/logging.h>
 
 namespace yarpl {
 namespace flowable {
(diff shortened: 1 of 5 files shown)


yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -2,31 +2,41 @@
...
-  // Note: if any of the following methods is overridden in a subclass,
-  // the new methods SHOULD ensure that these are invoked as well.
+  // Note: If any of the following methods is overridden in a subclass, the new
+  // methods SHOULD ensure that these are invoked as well.
   virtual void onSubscribe(Reference<Subscription> subscription) {
-    subscription_ = subscription;
+    DCHECK(subscription);
+
+    if (subscription_) {
+      subscription->cancel();
```
dymk: We probably want to complain loudly that we called onSubscribe twice

dymk:
Eg: 

```cpp
class Subscriber<T> {
protected:
    virtual onSubscribeImpl(Subscription) = 0;
    virtual onCompleteImpl() = 0;
    virtual onErrorImpl(exception) = 0;

public:
    void onSubscribe(Subscription s) {
        // do checks with 's'
        onSubscribeImpl(s);
    }

    void onComplete() {
        // do checks that we have a subscription
        onCompleteImpl();
    }

    void onError(ex) {
        // do checks we have a subscription
        onErrorImpl();
    }
};
```

Modifications to the codebase are straightforward; just rename methods in subclasses on Subscriber mechanically, the change in name for the virtual methods mean we can't accidentally miss a case.

yarpl/include/yarpl/flowable/Subscriber.h
alexmalyshev: I believe this actually fine, as there are some operators that re-subscribe (Rx.repeat I think), and that's okay.

alexmalyshev:
@dymk Yeah actually did have that `onFooImpl()` setup with the older `reactivesocket-cpp` codebase.  If we were to go back to the `onFooImpl` route, I think it shouldn't be on the top-level `Subscriber` class.  I think `Subscriber` should be completely abstract, and then have something like a `SubscriberImpl` or whatever subclass that does as you suggest.

That way we give the user a choice on what they want to do, they can opt into the convenient `impl` methods, or they can do their own thing.

PR #683, opened 2017-08-25
typesafe ObservableOperator

dymk:
same transformation as FlowableOperator and SingleOperator

### DIFF
--- a/yarpl/include/yarpl/observable/ObservableOperator.h
+++ b/yarpl/include/yarpl/observable/ObservableOperator.h
@@ -18,7 +18,7 @@ namespace observable {
  * pipelines
  * can be built: a Observable heading a sequence of Operators.
  */
-template <typename U, typename D>
+template <typename U, typename D, typename ThisOperatorT>
 class ObservableOperator : public Observable<D> {
  public:
   explicit ObservableOperator(Reference<Observable<U>> upstream)
@@ -36,16 +36,18 @@ class ObservableOperator : public Observable<D> {
                        public Observer<U> {
    protected:
     Subscription(
-        Reference<Observable<D>> observable,
+        Reference<ThisOperatorT> observable,
         Reference<Observer<D>> observer)
         : observable_(std::move(observable)), observer_(std::move(observer)) {
       assert(observable_);
       assert(observer_);
     }
 
-    template <typename TOperator>
-    TOperator* getObservableAs() {
-      return static_cast<TOperator*>(observable_.get());
+    Reference<ThisOperatorT>& getObservableOperator() {
+      static_assert(
+          std::is_base_of<Observable<D>, ThisOperatorT>::value,
+          "Operator must be a subclass of Observable<D>");
+      return observable_;
     }
 
     void observerOnNext(D value) {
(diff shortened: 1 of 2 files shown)


yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -18,7 +18,7 @@ namespace observable {
  * pipelines
  * can be built: a Observable heading a sequence of Operators.
  */
-template <typename U, typename D>
+template <typename U, typename D, typename ThisOperatorT>
```
phoad: I suggest renaming the template name from `ThisOperatorT` to smt shorter. All other templates are like U, D, F..

yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -165,31 +167,31 @@ template <
     typename D,
     typename F,
     typename = typename std::enable_if<std::is_callable<F(U), D>::value>::type>
-class MapOperator : public ObservableOperator<U, D> {
+class MapOperator : public ObservableOperator<U, D, MapOperator<U, D, F>> {
+  using ThisOperatorT = MapOperator<U, D, F>;
```
phoad: Do we need to write these two lines again and again in each sub class.
Is it possible to define them in the ObservableOperator class? 
If we do so, can we still be able to use them in the child classes?

yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -165,31 +167,31 @@ template <
...
+  using Super = ObservableOperator<U, D, ThisOperatorT>;
```
phoad: nit: As we already shorten some long declarations with `using` keyword, we might utilize that for members of the Super class also. `Super::` is repeating too many times. Put this to 
```
using Super::upstream_;
```

yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -368,69 +375,70 @@ class SkipOperator : public ObservableOperator<T, T> {
...
+        : Super::Subscription(std::move(observable), std::move(observer)),
           worker_(std::move(worker)) {}
 
     void cancel() override {
       worker_->schedule([this] { this->callSuperCancel(); });
     }
 
     void onNext(T value) override {
-      auto* observer = ObservableOperator<T, T>::Subscription::observer_.get();
+      auto observer = Super::Subscription::observer_;
```
phoad: Can we use `using` keyword instead of this line? 
```
 class Subscription : public Super::Subscription {
     public:
        using Super::Subscription::observer_;
     ...
```

yarpl/include/yarpl/observable/ObservableOperator.h
dymk: Unqualified names aren't looked up in the parent class, so you'd end up prefixing everything with `ObservableOperator<child class template params>::<the type>` each time

yarpl/include/yarpl/observable/ObservableOperator.h
dymk: We could, I'm not convinced this makes it more apparent what the code is doing or where `observer_` comes from though

PR #684, opened 2017-08-25
NonAtomicManualRefcounted

dymk:
Helper for cases like `AsyncSocket` or ProxyGen

```cpp
class InteractsWithExternalApi : public ReadDoneHandler, public NonAtomicManualReference {
  InteractsWithExternalApi(AsyncThing some_async_thing) : NonAtomicManualReference() {
    // some other place might have a Reference to us, maybe not
    give_something_a_handle(get_ref(this));

    for(int i = 0; i < 10; i++) {
      incRef();
      some_async_thing->queue_read(this); // queue up a call to read_done()
    }
  }

  void read_done() override {
    decRef();
  }
}
```

### DIFF
--- a/yarpl/include/yarpl/Refcounted.h
+++ b/yarpl/include/yarpl/Refcounted.h
@@ -333,6 +333,33 @@ Reference<T> get_ref(T* object) {
   return Reference<T>(object, detail::do_initial_refcount_check{});
 }
 
+////////////////////////////////////////////////////////////////////////////////
+
+// helper for retaining a reference to a Refcounted object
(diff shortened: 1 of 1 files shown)


dymk:
(needs tests written, don't merge yet, just wanting to get feedback on this)

#665 is a usecase for this, which otherwise needs to call `{inc,dec}Ref`

lehecka:
I don't think we need this class. boost::intrusive_ptr combined with intrusive_ptr_add_ref and intrusive_ptr_release functions seems to cover the use case just fine.
Thanks for offering a solution thou!

PR #686, opened 2017-08-27
Adding support for cancellation to Observables

lehecka:
* Merged AtomicBoolSubscription and Subscription. The functionality is needed in the base class and by the reactive streams implementation guide, the Subscription methods should be thread safe. 

Adding similar behaviors as we have in Observables in Java.
* added support to call Observable<T>->subscribe(...)->cancel();
* added support for isCancelled

added new unit tests.

### DIFF
--- a/yarpl/include/yarpl/observable/ObservableOperator.h
+++ b/yarpl/include/yarpl/observable/ObservableOperator.h
@@ -468,7 +468,7 @@ class FromPublisherOperator : public Observable<T> {
       Observer<T>::onSubscribe(std::move(subscription));
     }
 
-    void onSubscribe(Reference<Subscription> subscription) override {
+    void onSubscribe(Reference<Subscription>) override {
       DLOG(ERROR) << "not allowed to call";
     }
 
--- a/yarpl/include/yarpl/observable/ObservableOperator.h
+++ b/yarpl/include/yarpl/observable/ObservableOperator.h
@@ -6,7 +6,7 @@
 
 #include "yarpl/Observable.h"
 #include "yarpl/observable/Observer.h"
-#include "yarpl/observable/Subscription.h"
+#include "yarpl/observable/Subscriptions.h"
 
 namespace yarpl {
 namespace observable {
@@ -32,10 +32,10 @@ class ObservableOperator : public Observable<D> {
   /// against Operators.  Each operator subscription has two functions: as a
   /// subscriber for the previous stage; as a subscription for the next one,
   /// the user-supplied subscriber being the last of the pipeline stages.
-  class Subscription : public ::yarpl::observable::Subscription,
+  class OperatorSubscription : public ::yarpl::observable::Subscription,
                        public Observer<U> {
    protected:
-    Subscription(
+    OperatorSubscription(
         Reference<Observable<D>> observable,
         Reference<Observer<D>> observer)
         : observable_(std::move(observable)), observer_(std::move(observer)) {
@@ -67,6 +67,7 @@ class ObservableOperator : public Observable<D> {
     // Subscription.
 
     void cancel() override {
+      Subscription::cancel();
       terminateImpl(TerminateState::Up());
     }
 
@@ -75,10 +76,10 @@ class ObservableOperator : public Observable<D> {
     void onSubscribe(
         Reference<yarpl::observable::Subscription> subscription) override {
       if (upstream_) {
+        DLOG(ERROR) << "attempt to subscribe twice";
         subscription->cancel();
         return;
       }
-
       upstream_ = std::move(subscription);
       observer_->onSubscribe(get_ref(this));
     }
@@ -154,6 +155,7 @@ class ObservableOperator : public Observable<D> {
     /// calls should be forwarded upstream.  Note that `this` is also a
     /// observer for the upstream stage: thus, there are cycles; all of
     /// the objects drop their references at cancel/complete.
+    //TODO(lehecka): this is extra field... base class has this member so remove it
     Reference<::yarpl::observable::Subscription> upstream_;
   };
 
@@ -171,18 +173,20 @@ class MapOperator : public ObservableOperator<U, D> {
       : ObservableOperator<U, D>(std::move(upstream)),
         function_(std::move(function)) {}
 
-  void subscribe(Reference<Observer<D>> observer) override {
+  Reference<Subscription> subscribe(Reference<Observer<D>> observer) override {
+    auto subscription = make_ref<MapSubscription>(get_ref(this), std::move(observer));
     ObservableOperator<U, D>::upstream_->subscribe(
         // Note: implicit cast to a reference to a observer.
-        make_ref<Subscription>(get_ref(this), std::move(observer)));
+        subscription);
+    return subscription;
   }
 
  private:
-  class Subscription : public ObservableOperator<U, D>::Subscription {
-    using Super = typename ObservableOperator<U, D>::Subscription;
+  class MapSubscription : public ObservableOperator<U, D>::OperatorSubscription {
+    using Super = typename ObservableOperator<U, D>::OperatorSubscription;
 
    public:
-    Subscription(
+    MapSubscription(
         Reference<Observable<D>> observable,
         Reference<Observer<D>> observer)
         : Super(std::move(observable), std::move(observer)) {}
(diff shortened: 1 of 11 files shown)


yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -171,18 +173,20 @@ class MapOperator : public ObservableOperator<U, D> {
       : ObservableOperator<U, D>(std::move(upstream)),
         function_(std::move(function)) {}
 
-  void subscribe(Reference<Observer<D>> observer) override {
+  Reference<Subscription> subscribe(Reference<Observer<D>> observer) override {
+    auto subscription = make_ref<MapSubscription>(get_ref(this), std::move(observer));
     ObservableOperator<U, D>::upstream_->subscribe(
```
phoad: Why don't we directly return the result of this call? Perhaps the ObservableOperator can augment the given subscription and return the augmented one. So, might be better to return that one.

```
return ObservableOperator<U, D>::upstream_->subscribe(make_ref<MapSubscription>(get_ref(this), std::move(observer));
```

yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -207,18 +211,20 @@ class FilterOperator : public ObservableOperator<U, U> {
       : ObservableOperator<U, U>(std::move(upstream)),
         function_(std::move(function)) {}
 
-  void subscribe(Reference<Observer<U>> observer) override {
+  Reference<Subscription> subscribe(Reference<Observer<U>> observer) override {
+    auto subscription = make_ref<FilterSubscription>(get_ref(this), std::move(observer));
     ObservableOperator<U, U>::upstream_->subscribe(
```
phoad: ditto

yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -247,18 +253,20 @@ class ReduceOperator : public ObservableOperator<U, D> {
       : ObservableOperator<U, D>(std::move(upstream)),
         function_(std::move(function)) {}
 
-  void subscribe(Reference<Observer<D>> subscriber) override {
+  Reference<Subscription> subscribe(Reference<Observer<D>> subscriber) override {
+    auto subscription = make_ref<ReduceSubscription>(get_ref(this), std::move(subscriber));
     ObservableOperator<U, D>::upstream_->subscribe(
```
phoad: ditto

yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -445,8 +461,44 @@ class FromPublisherOperator : public Observable<T> {
   explicit FromPublisherOperator(OnSubscribe function)
       : function_(std::move(function)) {}
 
-  void subscribe(Reference<Observer<T>> observer) override {
-    function_(std::move(observer));
+ private:
+  class PublisherObserver : public Observer<T> {
```
phoad: I could not understand why did we need this class. Is it for just to eliminate calling onSubscribe?

yarpl/include/yarpl/observable/Observer.h
```diff
@@ -42,11 +43,21 @@ class Observer : public virtual Refcounted {
 
   virtual void onNext(T) = 0;
 
+  bool isUnsubscribed() const {
+    CHECK(subscription_);
```
phoad: Why do we call CHECK(), is subscription_ is null, shouldn't we just return false?

yarpl/perf/Observable_perf.cpp
```diff
@@ -20,7 +19,6 @@ BENCHMARK(Observable_OnNextOne_ConstructOnly);
 
 static void Observable_OnNextOne_SubscribeOnly(benchmark::State& state) {
   auto a = Observable<int>::create([](yarpl::Reference<Observer<int>> obs) {
-    obs->onSubscribe(Subscriptions::empty());
```
phoad: nice!

yarpl/test/Observable_test.cpp
```diff
@@ -153,39 +143,34 @@ TEST(Observable, ItemsCollectedSynchronously) {
  * in a Vector which could then be consumed on another thread.
  */
 TEST(DISABLED_Observable, ItemsCollectedAsynchronously) {
```
phoad: Is it possible to enable these now :)

yarpl/test/Observable_test.cpp
```diff
@@ -470,3 +468,85 @@ TEST(Observable, CancelReleasesObjects) {
...
+      checkpoint.Call();
+    });
+  });
+  auto infinite2 = infinite1->skip(1)->skip(1);
+  auto test1 = make_ref<InfiniteAsyncTestOperator>(infinite2, checkpoint2);
+  auto test2 = make_ref<InfiniteAsyncTestOperator>(test1->skip(1), checkpoint3);
+  auto skip = test2->skip(8);
+
+  auto subscription = skip->subscribe([](int){});
+  while(emitted < 1000);
```
phoad: nit: better to have a timed wait, in case of failure, we can early notice instead of waiting infinitely.

yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -75,10 +76,10 @@ class ObservableOperator : public Observable<D> {
     void onSubscribe(
         Reference<yarpl::observable::Subscription> subscription) override {
       if (upstream_) {
+        DLOG(ERROR) << "attempt to subscribe twice";
```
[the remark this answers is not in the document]
lehecka: I believe this error is just to avoid misusing the API. Using cancelled subscription for onSubscribe is as wrong as re-subscribing with already closed Subscriber. I would ideally just std::terminate for this case, but we should avoid crashing if possible.

yarpl/include/yarpl/observable/ObservableOperator.h
lehecka: Funny you are asking something which I tried during working on this change. I added a unit test which shows we need to properly pass the cancel signal thru all chained operator subscriptions. The object dependencies between observable-subscriber-subscription is already so complicated that adding option to "augment" subscription feels to me like an overkill, not mentioning that its hard to ensure we have the right terminate signal chain.

yarpl/include/yarpl/observable/ObservableOperator.h
lehecka: yeah, the above comment applies to all operators. Please note that all the operator subscription class (eq. FilterSubscription) is implementing both Subscriber and Subscription which needs to be properly chained to the upstream subscribers. 
(its easy to get lost in the complexity when debugging)

yarpl/include/yarpl/observable/ObservableOperator.h
lehecka: Exactly. As I am following the Java RX, we also don't want the application code to call onSubscribe in the lambdas. I believe they choose this behavior because it adds more states the subscriber can be in when you get it as parameter for your lambda:
Observable::create([](Subscriber s) { /*its already subscribed, so I don't have to worry about onSubscribe*/ });

PublisherObserver just swallows the onSubscribe in release mode, but it errors out in the debug mode.

yarpl/include/yarpl/observable/Observer.h
lehecka: Good suggestion, but thinking about the meaning of isUnsubscribed when the subscriber hasn't subscribed yet (subscription_ == null), it feels wrong.
Reactive streams protocol says when we call onComplete or onError, the subscription object is no longer valid so I would argue that it also makes no sense to call isUnsubscribed.
Would you agree with that?

yarpl/test/Observable_test.cpp
lehecka: it should. Let me do it in the next diff.
thanks for the suggestion

yarpl/test/Observable_test.cpp
lehecka: yeah, good idea, I will fix this.

phoad:
In file included from /home/travis/build/rsocket/rsocket-cpp/yarpl/include/yarpl/observable/Observable.h:113:0,
                 from /home/travis/build/rsocket/rsocket-cpp/yarpl/include/yarpl/Observable.h:6,
                 from /home/travis/build/rsocket/rsocket-cpp/yarpl/test/Observable_test.cpp:9:
/home/travis/build/rsocket/rsocket-cpp/yarpl/include/yarpl/observable/ObservableOperator.h: In lambda function:
/home/travis/build/rsocket/rsocket-cpp/yarpl/include/yarpl/observable/ObservableOperator.h:51:10: error: ‘void yarpl::observable::ObservableOperator<U, D>::OperatorSubscription::observerOnNext(D) [with U = int; D = int]’ is protected
     void observerOnNext(D value) {
          ^
/home/travis/build/rsocket/rsocket-cpp/yarpl/test/Observable_test.cpp:516:34: error: within this context
           Super::observerOnNext(1);

yarpl/include/yarpl/observable/Observer.h
```diff
@@ -21,6 +21,7 @@ class Observer : public virtual Refcounted {
     DCHECK(subscription);
 
     if (subscription_) {
+      DLOG(ERROR) << "attempt to double subscribe";
       subscription->cancel();
```
dymk: Do we want to fail loudly/throw an exception here?

yarpl/include/yarpl/observable/Observer.h
dymk: Can we call this `isSubscribed` so we're writing `if(obj->isSubscribed())` rather than the double-negative `if(!obj->isUnsubscribed())`?

phoad:
@dymk has rebased your diff and resolved the conflicts and we checked that in @lehecka ;)
should be ok to close this.

PR #687, opened 2017-08-27
Run tests binary with current_version=0.1

alexmalyshev:
Making sure nothing gets broken.  Some of the SetupResumeAcceptor tests were
failing as they very much assumed things to be 1.0.

### DIFF
--- a/test/internal/SetupResumeAcceptorTest.cpp
+++ b/test/internal/SetupResumeAcceptorTest.cpp
@@ -19,10 +19,12 @@ namespace {
  * Make a legitimate-looking SETUP frame.
  */
 Frame_SETUP makeSetup() {
+  auto version = FrameSerializer::getCurrentProtocolVersion();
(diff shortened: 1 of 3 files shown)


test/internal/SetupResumeAcceptorTest.cpp
```diff
@@ -138,56 +135,9 @@ TEST(SetupResumeAcceptor, SingleSetup) {
   EXPECT_TRUE(setupCalled);
 }
 
-TEST(SetupResumeAcceptor, SetupAndFnf) {
```
phoad: Why did you delete this one?

alexmalyshev: This test is invalid, actually.  It sends a SETUP and a REQUEST_FNF frame to a `OneFrameProcessor`, which isn't supposed to happen.

PR #688, opened 2017-08-28
Further cleaning of ResumeManager API

somasun:
This PR has the following changes.
- `InMemResumeManager` has been divided into two classes - `WarmResumeManager` and `ColdResumeManager`.  `WarmResumeManager` will be the default ResumeManager used in RSocket.  `ColdResumeManager` is an in-memory implementation of ResumeManger for cold-resumption (for testing and prototyping of cold-resumption).  It extends `WarmResumeManager` and implements additional methods relevant for cold resumption.
- The cold resumption specific API has been simplified to involve only one call - `getStreamInfo()` to get all stream related information during cold-resumption.

I guess with this PR, all changes needed for basic cold-resumption of client-originated REQUEST_STREAMs are in. 

Notes:
- I will follow up by writing a disk-based ColdResumeManager
- The cmake targets needs to be simplified so that common targets can be reused by both tests and examples.


rsocket/ResumeManager.h
```diff
@@ -15,9 +15,17 @@ class IOBuf;
 
 namespace rsocket {
 
+struct StreamResumeInfo {
+  StreamType streamType;
+  Requester requester;
+  std::string streamToken{""};
```
alexmalyshev: Nit: Don't need the `{""}` bit, default ctor of `std::string` is an empty string.

rsocket/internal/Common.h
```diff
@@ -59,6 +59,11 @@ enum class StreamType {
   FNF,
 };
 
+enum class Requester {
```
alexmalyshev: I can see this getting conflated with `RSocketRequester`...  Is there a different name we can use?

rsocket/internal/WarmResumeManager.h
```diff
@@ -0,0 +1,98 @@
...
+    return impliedPosition_;
+  }
+
+  void onStreamClosed(StreamId) override{
+      // No action to perform for WarmResumeManager
+  };
+
+  const std::unordered_map<StreamId, StreamResumeInfo>& getStreamResumeInfo()
+      override {
+    LOG(FATAL) << "Not Implemented for Warm Resumption";
```
alexmalyshev: Should this crash?  It seems like we should have softer failures for someone trying to do cold resumption with `WarmResumeManager`.

somasun: I think crash is better here.  Users should not be attempting cold resumption with `WarmResumeManager`.   This is mostly during development.  It is unlikely that a user will suddenly hit this code path and FATAL in production.

rsocket/ResumeManager.h
```diff
@@ -15,9 +15,42 @@ class IOBuf;
 
 namespace rsocket {
 
+// Struct to hold information relevant per stream.
+struct StreamResumeInfo {
+  StreamResumeInfo() = delete;
+  StreamResumeInfo(
+      StreamType _streamType,
```
alexmalyshev: Nit: Don't name variables `_foo`, all underscore-prefixed identifiers are "reserved by the system" in C++.

rsocket/framing/Frame.cpp
```diff
@@ -249,4 +249,25 @@ std::ostream& operator<<(std::ostream& os, const Frame_ERROR& frame) {
...
+    if (frameType == FrameType::REQUEST_STREAM) {
+      return StreamType::STREAM;
+    } else if (frameType == FrameType::REQUEST_CHANNEL) {
+      return StreamType::CHANNEL;
+    } else if (frameType == FrameType::REQUEST_RESPONSE) {
+      return StreamType::REQUEST_RESPONSE;
+    } else if (frameType == FrameType::REQUEST_FNF) {
+      return StreamType::FNF;
+    } else {
+      CHECK(false) << "Unknown open stream frame : " << frameType;
```
alexmalyshev: Nit: Use `LOG(FATAL)` instead.  It's marked as `[[noreturn]]` so there won't be any compilation errors.

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -461,28 +461,27 @@ void RSocketStateMachine::handleConnectionFrame(
...
-                  std::move(subscriber),
-                  *folly::EventBaseManager::get()->getEventBase()),
-              streamId,
-              consumerAllowance);
-        }
-
         streamsFactory().setNextStreamId(
             resumeManager_->getLargestUsedStreamId());
-
+        for (auto it : resumeManager_->getStreamResumeInfos()) {
```
alexmalyshev: You want `auto&`, `it` here is a `std::pair<StreamId, StreamResumeInfo>` which you don't want to copy.

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -620,74 +619,67 @@ void RSocketStateMachine::handleUnknownStream(
...
-      auto stateMachine =
-          streamsFactory_.createChannelResponder(frame.requestN_, streamId);
-      auto requestSink = requestResponder_->handleRequestChannelCore(
-          std::move(frame.payload_), streamId, stateMachine);
-      stateMachine->subscribe(requestSink);
-      break;
+  if (!isNewStreamFrame(frameType)) {
+    std::stringstream msg;
+    msg << "Unexpected frame " << frameType << " for stream " << streamId;
+    LOG(ERROR) << msg.str();
```
alexmalyshev: I think this should be a `VLOG` or deleted.  I can see this getting spammy with rogue clients, and it's better to just stay away from that.

test/test_utils/ColdResumeManager.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+void ColdResumeManager::trackSentFrame(
+    const folly::IOBuf& serializedFrame,
+    FrameType frameType,
+    folly::Optional<StreamId> streamIdPtr,
+    size_t consumerAllowance) {
+  if (!shouldTrackFrame(frameType)) {
+    return;
+  }
+  const StreamId streamId = *streamIdPtr;
```
alexmalyshev: Why don't we need to check if `streamIdPtr` is `folly::none`?

somasun: All frames should have StreamID.  So I want to get rid of the `folly::Optional` part for the StreamID in another PR.  For now, will add the check.

PR #689, opened 2017-08-28
Adding support for cancellation to Observables

dymk:
rebase #686 on top of master to use typesafe changes to ObservableOperator

@lehecka @phoad


dymk:
I wish Travis had a configuration option for allowing failures from any of one of our gcc configs, or auto-restarting ones caused by killed processes. It feels like at least half of these builds OOM one of the GCCs.

PR #690, opened 2017-08-28
Call Subscriber's onError when client is closed

dymk:
Previously we just dropped those requests on the floor, this now calls the application provided Subscriber's `onError` method, so the application can tell if its request even went through or not.

Adds tests for requestResponse, requestStream, and requestChannel

### DIFF
--- a/rsocket/RSocketRequester.cpp
+++ b/rsocket/RSocketRequester.cpp
@@ -53,9 +53,14 @@ RSocketRequester::requestChannel(
               std::move(subscriber), *eb));
       // responseSink is wrapped with thread scheduling
       // so all emissions happen on the right thread
-      requestStream->subscribe(
-          yarpl::make_ref<ScheduledSubscriber<Payload>>(std::move(responseSink),
-                                                        *eb));
+
+      // if we don't get a responseSink back, that means that
+      // the requesting peer wasn't connected (or similar error)
+      // and the Flowable it gets back will immediately call onError
+      if (responseSink) {
+        requestStream->subscribe(yarpl::make_ref<ScheduledSubscriber<Payload>>(
+            std::move(responseSink), *eb));
+      }
     });
   });
 }
--- a/rsocket/statemachine/StreamsFactory.cpp
+++ b/rsocket/statemachine/StreamsFactory.cpp
@@ -10,6 +10,9 @@
 #include "rsocket/statemachine/StreamRequester.h"
 #include "rsocket/statemachine/StreamResponder.h"
 
+#include "yarpl/flowable/Flowable.h"
+#include "yarpl/single/Singles.h"
+
 namespace rsocket {
 
 using namespace yarpl;
@@ -25,9 +28,30 @@ StreamsFactory::StreamsFactory(
               : 2 /*streams initiated by the server MUST use
                     even-numbered stream identifiers*/) {}
 
+void subscribeToErrorFlowable(
+    Reference<yarpl::flowable::Subscriber<Payload>> responseSink) {
+  yarpl::flowable::Flowables::error<Payload>(
+      std::runtime_error(
+          "state machine is disconnected/closed"))
+      ->subscribe(std::move(responseSink));
+}
+
+void subscribeToErrorSingle(
+    Reference<yarpl::single::SingleObserver<Payload>> responseSink) {
+  yarpl::single::Singles::error<Payload>(
+      std::runtime_error(
+          "state machine is disconnected/closed"))
+      ->subscribe(std::move(responseSink));
+}
+
 Reference<yarpl::flowable::Subscriber<Payload>>
 StreamsFactory::createChannelRequester(
     Reference<yarpl::flowable::Subscriber<Payload>> responseSink) {
+  if (connection_.isDisconnectedOrClosed()) {
+    subscribeToErrorFlowable(std::move(responseSink));
+    return nullptr;
+  }
+
   ChannelRequester::Parameters params(
       connection_.shared_from_this(), getNextStreamId());
   auto stateMachine = yarpl::make_ref<ChannelRequester>(params);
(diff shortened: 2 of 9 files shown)


rsocket/statemachine/StreamsFactory.cpp
```diff
@@ -25,9 +28,30 @@ StreamsFactory::StreamsFactory(
               : 2 /*streams initiated by the server MUST use
                     even-numbered stream identifiers*/) {}
 
+void subscribeToErrorFlowable(
+    Reference<yarpl::flowable::Subscriber<Payload>> responseSink) {
+  yarpl::flowable::Flowables::error<Payload>(
+      folly::make_exception_wrapper<std::runtime_error>(
```
alexmalyshev: Nit: `folly::make_exception_wrapper<Exn>` is unnecessary.  `exception_wrapper` can be explicitly constructed from any subclass of `std::exception`.

test/RSocketTests.cpp
```diff
@@ -53,6 +54,58 @@ std::shared_ptr<RSocketClient> makeClient(
   return makeClientAsync(eventBase, port).get();
 }
 
+namespace {
+struct DisconnectedRequestResponseHandler : public rsocket::RSocketResponder {
+  explicit DisconnectedRequestResponseHandler() {}
```
alexmalyshev: Nit: Don't use `explicit` for zero-arg ctors.

test/test_utils/GenericRequestResponseHandler.h
```diff
@@ -64,20 +64,20 @@ struct GenericRequestResponseHandler : public rsocket::RSocketResponder {
...
 }
 
-Response payload_response(std::string const& a, std::string const& b) {
+inline Response payload_response(std::string const& a, std::string const& b) {
   return payload_response({a, b});
 }
 
 template <typename T>
-Response error_response(T const& err) {
+inline Response error_response(T const& err) {
```
alexmalyshev: Nit: Don't need `inline` on template functions.

test/RequestStreamTest.cpp
```diff
@@ -93,3 +93,30 @@ TEST(RequestStreamTest, HelloAsync) {
...
+            // onNext shouldn't be called
+            ASSERT(false);
+          },
+          [&](folly::exception_wrapper) {
+            did_call_on_error = true;
+            wait_for_on_error.post();
+          },
+          []() {
+            // onComplete shouldn't be called
+            ASSERT(false);
```
alexmalyshev: Use `FAIL()` instead of `ASSERT(false)`.

test/RSocketTests.cpp
```diff
@@ -53,6 +54,58 @@ std::shared_ptr<RSocketClient> makeClient(
...
+std::shared_ptr<RSocketClient> makeDisconnectedClient(
+    folly::EventBase* eventBase) {
+  auto server =
+      makeServer(std::make_shared<DisconnectedRequestResponseHandler>());
+
+  auto client = makeClient(eventBase, *server->listeningPort());
+
+  client->disconnect();
+  folly::Baton<> wait_for_disconnect;
+  eventBase->runInEventBaseThread([&] { wait_for_disconnect.post(); });
```
alexmalyshev: Seems like we should make `RSocketClient::disconnect()` return a future.  This /should/ work, but I don't want to see this done for every single disconnect test case.

rsocket/RSocketClient.cpp
```diff
@@ -129,6 +129,10 @@ void RSocketClient::disconnect(folly::exception_wrapper ex) {
   });
 }
 
+bool RSocketClient::isDisconnectedOrClosed() const {
```
lehecka: This is not a thread safe method. Do you need it in the RSocketClient class?
The StreamsFactory is using the RSocketStateMachine:: isDisconnectedOrClosed, is that correct?

rsocket/RSocketRequester.cpp
```diff
@@ -53,9 +53,14 @@ RSocketRequester::requestChannel(
               std::move(subscriber), *eb));
       // responseSink is wrapped with thread scheduling
       // so all emissions happen on the right thread
-      requestStream->subscribe(
-          yarpl::make_ref<ScheduledSubscriber<Payload>>(std::move(responseSink),
-                                                        *eb));
+
+      // if we don't get a responseSink back, that means that
+      // the requesting peer wasn't connected (or similar error)
+      // and the Flowable it gets back will immediatly call onError
```
alexmalyshev: Nit: `s|immediatly|immediately|g`

test/RequestResponseTest.cpp
```diff
@@ -135,6 +135,28 @@ TEST(RequestResponseTest, FailureInResponse) {
...
+  requester->requestResponse(Payload("foo", "bar"))
+      ->subscribe(
+          [](auto) {
+            // should not call onSuccess
+            FAIL();
+          },
+          [&](folly::exception_wrapper) {
+            did_call_on_error = true;
+            wait_for_on_error.post();
+          });
```
alexmalyshev: Did you want to do an onComplete lambda that also does `FAIL()` (mimicking `RequestStreamTest.RequestOnDisconnectedClient`)

test/RSocketTests.cpp
```diff
@@ -53,6 +54,58 @@ std::shared_ptr<RSocketClient> makeClient(
   return makeClientAsync(eventBase, port).get();
 }
 
+namespace {
+struct DisconnectedRequestResponseHandler : public rsocket::RSocketResponder {
+  DisconnectedRequestResponseHandler() {}
+
+  yarpl::Reference<yarpl::single::Single<rsocket::Payload>>
+  handleRequestResponse(rsocket::Payload, rsocket::StreamId) override {
+    assert(false);
```
alexmalyshev: Nit: Shouldn't this be named `DisconnectedResponder`?  It's doing `assert(false)` on all possible requests, not just req-response.

Nit: I'd replace `assert(false)` with `CHECK` tbh, I think it makes sense to fail loudly in all build modes for this test-only construct.

test/RequestResponseTest.cpp
dymk: It's requestResponse, so the subscriber only has onSuccess and onError

test/RSocketTests.cpp
dymk: That's a better name yeah; current name is a relic from when it was just for R/R tests.

test/RequestResponseTest.cpp
alexmalyshev: Oh right, I always forget that...

PR #695, opened 2017-08-29
adding support for Observer::addSubscription

lehecka:
Similarly to Java API, we will be able to add more callbacks for the on-unsubscribe signal.
This makes it easy for the lambdas in observables to hook into the un-subscribe logic.
See the unit test for the usage.


yarpl/src/yarpl/observable/Subscriptions.cpp
```diff
@@ -12,12 +13,20 @@ namespace observable {
...
+  }
 }
 
 bool Subscription::isCancelled() const {
   return cancelled_;
 }
 
+void Subscription::tieSubscription(Reference<Subscription> subscription) {
+  CHECK(subscription);
+  tiedSubscriptions_->push_back(std::move(subscription));
```
dymk: What should happen if you call `tieSubscription` after `cancel`? This seems racey.

lehecka: What would you like to see happening? I would suggest that after the subscription is canceled, any tied subscriptions are ignored.

dymk: Because it's now nondeterministic if a subscription that you tie into a subscription that's about to be canceled will ever receive a cancel. There's no paired `onSubscribe` to signal if a subscription was added before a cancel (and if the Subscription *will* have its cancel called), so the application developer won't know if they're throwing a `Subscription` into the ether or not.

lehecka: Agree. I will change it so that is the case the original subscription is cancelled, I will cancel all tied subscriptions right away.

PR #696, opened 2017-08-29
fix concurrency bug in OperatorSubscription::observerOnNext

lehecka:
Fixing concurrency problem detected by the unit test:
https://travis-ci.org/rsocket/rsocket-cpp/jobs/269724602

I believe we have more concurrency bugs in the ObservableOperator implementation. Namely, the cancel method nullifies some of the member variables (references). This happens non atomically and the other methods are still accessing the the member variables (non-atomically). As a consequence we can get into corrupted pointers issues when the instance is used in the concurrent environment.

This is more a bandate than a proper fix.


lehecka:
I am not very happy with this fix. Let's hold off the merge of this for a bit.

PR #700, opened 2017-08-30
Fix use-after-free in RSocketStateMachine::processFrame()

alexmalyshev:
processFrame() might be processing the CANCEL frame for the last stream state
machine in the RSocketStateMachine.  This ends up destroying the entire
RSocketStateMachine object, but we still try to use the frame afterwards
(tracking it for resumption).

Repros locally on Linux with gcc-7.1.1, running
benchmarks/StreamThroughputTcp.cpp under valgrind.

Also fix VLOG printing of CANCEL frames in RSocketStateMachine.  They were using
the default ctor, which doesn't set the FrameType to CANCEL, printing it as
RESERVED instead.


dymk:
Ha, we've seen this exact same type of bug manifested with Reference. It's a bummer we can't statically check for instances of this bug.

PR #701, opened 2017-08-30
BackpressureStrategy { BUFFER, DROP, ERROR, LATEST, MISSING };

lehecka:
implementing all standard strategies.
Added unit test for all of them except for MISSING since that one is supposed to be used with some custom strategy operator, which we don't have yet.


yarpl/test/Observable_test.cpp
```diff
@@ -250,30 +252,32 @@ TEST(Observable, CancelFromDifferentThread) {
...
 
-  std::vector<int> v;
-  f->subscribe(yarpl::flowable::Subscribers::create<int>(
-      [&v](const int& value) { v.push_back(value); }));
+  std::vector<int64_t> v;
 
-  EXPECT_EQ(v.at(0), 1);
+  auto subscriber = make_ref<testing::StrictMock<yarpl::flowable::MockSubscriber<int64_t>>>(5);
+
+  EXPECT_CALL(*subscriber, onSubscribe_(_));
```
dymk:
```cpp
{
  ::testing::InSequence ensure_order;
  EXPECT_CALL(...);
  EXPECT_CALL(...);
}
``` 
to verify that these calls take place in the order you `EXPECT_CALL` them in (unless StrictMock does this already? looks like it only treats unEXPECT'd calls as errors).

yarpl/test/Observable_test.cpp
```diff
@@ -285,6 +289,73 @@ TEST(Observable, toFlowableWithCancel) {
...
+      .WillRepeatedly(Invoke([&](int64_t value) {
+          v.push_back(value);
+      }));
+  EXPECT_CALL(*subscriber, onComplete_());
+
+  f->subscribe(subscriber);
+  EXPECT_EQ(v, std::vector<int64_t>({1, 2, 3, 4, 5}));
+
+  subscriber->subscription()->request(5);
+  EXPECT_EQ(v, std::vector<int64_t>({1, 2, 3, 4, 5, 9}));
```
dymk: So LATEST is like DROP but we buffer only the last element?

yarpl/include/yarpl/observable/Observer.h
```diff
@@ -48,6 +48,10 @@ class Observer : public virtual Refcounted {
     return subscription_->isCancelled();
   }
 
+  bool isUnsubscribedOrTerminated() const {
+    return !subscription_ || subscription_->isCancelled();
+  }
```
dymk: Races with calls to `onComplete/onError` which may reset `subscription_`

yarpl/include/yarpl/flowable/Flowable_FromObservable.h
```diff
@@ -40,58 +49,200 @@ class FlowableFromObservableSubscription
...
+      Super::onComplete();
+    }
+  }
+
+  //
+  // onError signal is delivered immediately by design
+  //
+
+  void onNextWithoutCredits(T t) override {
+    storesLatest_ = true;
```
dymk: `latest_` being protected but `storesLatest_`  not seems racy

yarpl/include/yarpl/flowable/Flowable_FromObservable.h
```diff
@@ -40,58 +49,200 @@ class FlowableFromObservableSubscription
...
+    *latest_.wlock() = std::move(t);
+  }
+
+  void onCreditsAvailable(int64_t credits) override {
+    DCHECK(credits > 0);
+    if(storesLatest_) {
```
dymk: looks like if `storesLatest_ == true` and T1 calls `onCreditsAvailable`, we context switch here, and T2 calls `onCreditsAvailable`, we race and call onNext will be called twice, once with a reset/empty due to the move `T`.

dymk:
I'm very hesitant about most of the thread-safety of Flowable and Observable; we're mixing mutexes and atomics in not thread safe ways. I think we should relax the restriction that Flowable/Observable are threadsafe (making the case where a F/O is used in a single threaded environment fast and lock/atomic free), and provide a thin synchronization wrapper over them (making a multithreaded environment correct and the wrapper testable), rather than having to worry about synchronization and races in the meat of F/O.

yarpl/test/Observable_test.cpp
lehecka: yeah, I know about InSequence. I would argue that in this test, I am not testing the sequence of calls. That is tested somewhere else.

yarpl/test/Observable_test.cpp
lehecka: that is correct: http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/BackpressureStrategy.html

yarpl/include/yarpl/observable/Observer.h
lehecka: that is true. This is not thread safe now and we need to address that. See my later comment.

yarpl/include/yarpl/flowable/Flowable_FromObservable.h
lehecka: storesLatest_ is atmoic... but I do agree that the model in general is broken..

yarpl/include/yarpl/flowable/Flowable_FromObservable.h
lehecka: T1 and T2 will never execute the methods concurrently. The execution is always serial. That is the contract for Flowables/Observables.

yarpl/test/Mocks.h
```diff
@@ -0,0 +1,160 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
```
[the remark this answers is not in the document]
lehecka: thanks for sending the PR to move the files. I will rebase it on top of it.

lehecka:
I agree with you that the concurrency model is broken for this new code as well as for the rest of the yarpl types. We need to work out a clear design how we want to achieve thread safety. Please note, that the contract is to allow being called from any thread, but there will never be concurrent calls. The implementation should take advantage of the fact that this is not full thread safety requirement (for instance we should favor using memory barriers / atomics over mutexes because we don't need exclusive access protection).
We can consider this implementation not thread safe, but in the single threaded environment its working fine. And since there are no users of this code yet, we do have a little time to address the concurrency. Actually we need to address it for the whole yarpl library.

yarpl/include/yarpl/flowable/Flowable_FromObservable.h
dymk: (y) that would be really great

yarpl/test/Observable_test.cpp
dymk: Fair enough

dymk:
Alright, in that case, let's discuss concurrency stuff later, and get this merged in. Test for a single threaded usecase are fine if we're going to build a generalized synchronization solution anyways.

PR #704, opened 2017-08-30
unittests for calling subscribe() on a flowable multiple times

dymk:
and a fix for EmitterSubscriber, which incorrectly checks that onComplete/onError aren't called after onCancel (which it is allowed to race with).

### DIFF
--- a/yarpl/test/FlowableTest.cpp
+++ b/yarpl/test/FlowableTest.cpp
@@ -7,6 +7,8 @@
 
 #include <folly/Baton.h>
 
+#include "yarpl/test_utils/Mocks.h"
+
 #include "yarpl/Flowable.h"
 #include "yarpl/flowable/TestSubscriber.h"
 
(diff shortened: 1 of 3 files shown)


yarpl/test/FlowableTest.cpp
```diff
@@ -538,5 +540,58 @@ TEST(FlowableTest, FlowableFromDifferentThreadsWithError) {
...
+  });
+
+  auto setup_mock = [](auto request_num, auto& resps) {
+    auto mock = make_ref<StrictMockSubscriber>(request_num);
+
+    Sequence seq;
+    EXPECT_CALL(*mock, onSubscribe_(_)).InSequence(seq);
+    EXPECT_CALL(*mock, onNext_(_))
+      .InSequence(seq)
+      .WillRepeatedly(Invoke([r = &resps](int64_t value) {
```
alexmalyshev: Nit: Why not just `[&resps]`?

yarpl/test/FlowableTest.cpp
```diff
@@ -538,5 +540,58 @@ TEST(FlowableTest, FlowableFromDifferentThreadsWithError) {
...
+  auto mock2 = setup_mock(5, results[1]);
+  auto mock3 = setup_mock(5, results[2]);
+  auto mock4 = setup_mock(5, results[3]);
+  auto mock5 = setup_mock(5, results[4]);
+
+  // map on the same flowable twice
+  auto stream1 = f->map([](auto i) { return i + 1; });
+  auto stream2 = f->map([](auto i) { return i * 2; });
+  auto stream3 = stream2->skip(2); // skip operator tests (chained on a map)
+  auto stream4 = stream1->take(3); // skip operator tests (chained on a map)
```
alexmalyshev: You mean `take operator tests` here, right?

yarpl/test/yarpl-tests.cpp
```diff
@@ -1,8 +1,12 @@
 // Copyright 2004-present Facebook. All Rights Reserved.
 
+#include <folly/init/Init.h>
+#include <glog/logging.h>
 #include <gtest/gtest.h>
 
 int main(int argc, char** argv) {
+  FLAGS_logtostderr = true;
```
alexmalyshev: 👍

yarpl/test/FlowableTest.cpp
dymk: good call; prior to this resps was on the stack

PR #707, opened 2017-08-31
Add more things to tune in the streaming tcp benchmark

alexmalyshev:
Number of server threads, number of clients, and the number of client threads.
Assign client threads round-robin to clients.

### DIFF
--- a/benchmarks/StreamThroughputTcp.cpp
+++ b/benchmarks/StreamThroughputTcp.cpp
@@ -2,6 +2,8 @@
 
 #include "benchmarks/Throughput.h"
 
+#include <deque>
+
 #include <folly/Baton.h>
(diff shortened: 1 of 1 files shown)


dymk:
Why are we using folly::Benchmark if we're ignoring the tuning `n` parameter, which can automatically find a good `FLAGS_ items` parameter?

alexmalyshev:
I don't want it to automatically find a good `n`, I want to control it explicitly.

alexmalyshev:
Also when I use `n`, follybenchmark will always set `n=1` (which is almost useless) until I manually do `--bm_min_iters=1000000`, which ends up being a more obtuse way of doing `--items=1000000`.

PR #708, opened 2017-08-31
IteratorObservable to encapsulate common 'make from an iterator' pattern

dymk:
Opens up implementing 'range', 'just', 'justn', 'justOnce' without nested lambdas and Observable<T>::create (which does confusing things wrt lifecycle of Subscriptions). Also enables creating an Observable from an arbitrary range (by putting it in a subclass of IteratorObservable and supplying a copy constructor). 

Idea for this came from talking from @benjchristensen and @lehecka about a `::fromIterator` static method on Flowable and Observable. We wanted to avoid separate "generator" lambdas and holding the initial state of a Flowable in whatever superclass implements `fromIterator`. This puts the responsibility of maintaining that initial state in the subclass, as we as being a generator for when `subscribe` is called multiple times. 

Subclasses encapsulate state for an iterator, and implement two methods: `haveNext` and `getNext` to signal to the superclass if the iterator is empty and to get the next element. Superclass handles cancellation. 

Not merge ready, looking for feedback on how the API looks.

### DIFF
--- a/yarpl/include/yarpl/observable/ObservableOperator.h
+++ b/yarpl/include/yarpl/observable/ObservableOperator.h
@@ -481,6 +481,73 @@ class SubscribeOnOperator
   std::unique_ptr<Worker> worker_;
 };
 
+template <typename T, typename GeneratorClass>
+class IteratorObservable : public Observable<T> {
+public:
+  Reference<Subscription> subscribe(Reference<Observer<T>> observer) override {
+    static_assert(
+      std::is_base_of<
+        IteratorObservable,
+        GeneratorClass
+      >::value,
+      "Inherit from IteratorObservable like `MyClass : public IteratorObservable<T, MyClass>`"
+    );
+
+    // can be stack allocated; we know nothing escapes
+    GeneratorClass instantiated(*static_cast<GeneratorClass const*>(this));
+    // so we can call protected methods on IteratorObservable even if they were
+    // made private in the subclass
+    IteratorObservable* withvtable = &instantiated;
+
+    auto subscription = make_ref<IteratorObservableSubscription>();
+
+    withvtable->beforeOnSubscribe();
+    observer->onSubscribe(subscription);
+
+    while(!subscription->isCancelled() && withvtable->haveNext()) {
+      observer->onNext(withvtable->getNext());
+    }
+    observer->onComplete();
+    withvtable->afterOnComplete();
+
+    return subscription;
+  }
+
+private:
+  struct IteratorObservableSubscription : public Subscription {
+    void cancel() override {
+      cancelled_ = true;
+    }
+  };
+
+protected:
+  virtual bool haveNext() = 0;
+  virtual T getNext() = 0;
+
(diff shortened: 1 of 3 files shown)


yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -481,6 +481,73 @@ class SubscribeOnOperator
...
+protected:
+  virtual bool haveNext() = 0;
+  virtual T getNext() = 0;
+
+  virtual void beforeOnSubscribe() {};
+  virtual void afterOnComplete() {};
+};
+
+template <typename T>
+class RangeObservable : public IteratorObservable<T, RangeObservable<T>> {
```
dymk: example usage here

yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -481,6 +481,73 @@ class SubscribeOnOperator
...
+        GeneratorClass
+      >::value,
+      "Inherit from IteratorObservable like `MyClass : public IteratorObservable<T, MyClass>`"
+    );
+
+    // can be stack allocated; we know nothing escapes
+    GeneratorClass instantiated(*static_cast<GeneratorClass const*>(this));
+    // so we can call protected methods on IteratorObservable even if they were
+    // made private in the subclass
+    IteratorObservable* withvtable = &instantiated;
```
dymk: I wonder if the compiler is smart enough to devirtualize the method calls to `withvtable` - it should have enough information to know it always points to a concrete `GeneratorClass`. Will investigate with godbolt.

yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -481,6 +481,73 @@ class SubscribeOnOperator
...
+      observer->onNext(withvtable->getNext());
+    }
+    observer->onComplete();
+    withvtable->afterOnComplete();
+
+    return subscription;
+  }
+
+private:
+  struct IteratorObservableSubscription : public Subscription {
```
lehecka: what is this subclass for? Subscription::cancel() does exactly what your implementation of cancel does...

yarpl/test/Observable_test.cpp
```diff
@@ -638,3 +638,15 @@ TEST(Observable, CancelSubscriptionChain) {
 
   LOG(INFO) << "cancelled after " << emitted << " items";
 }
+
+TEST(Observable, RangeIteratorObservable) {
+  auto range_observable = make_ref<RangeObservable<int>>(10, 15);
```
lehecka: what is the difference between this and 
Observables::range(...)?

dymk: This is a replacement for the implementation of Range which relies on the currenet `Flowable::create`, which is the weird one that requires the generator lambda return a tuple of (are we done yet, how many elements were sent).

lehecka:
this feels like a stale PR. Do we still want to land this?

PR #709, opened 2017-09-01
Test for racing case of server sending to client on a canceled Subscription 

dymk:
On the client side, rsocket incorrectly would not call `onComplete` or `onError` after the `Subscription` was canceled (in the synchronous case this was deterministic; it races in a multithreading environment however, we should make that clear. This fix does not address multithreading concerns).

I put this test in a separate file because it takes 10+ seconds to build the current RequestStreamTest.cpp file, and linking time isn't slow for us yet.

clang-format results are in a separate commit because it's impossible to filter out the noise otherwise.

### DIFF
--- a/rsocket/statemachine/ConsumerBase.cpp
+++ b/rsocket/statemachine/ConsumerBase.cpp
@@ -32,6 +32,7 @@ void ConsumerBase::checkConsumerRequest() {
   CHECK((state_ == State::RESPONDING) == !!consumingSubscriber_);
 }
 
+// TODO: this is probably buggy and misused and not needed (when completeConsumer exists)
 void ConsumerBase::cancelConsumer() {
   state_ = State::CLOSED;
   consumingSubscriber_ = nullptr;
@@ -45,11 +46,14 @@ void ConsumerBase::generateRequest(size_t n) {
 }
 
 void ConsumerBase::endStream(StreamCompletionSignal signal) {
+  VLOG(5) << "ConsumerBase::endStream(" << signal << ")";
   if (auto subscriber = std::move(consumingSubscriber_)) {
     if (signal == StreamCompletionSignal::COMPLETE ||
         signal == StreamCompletionSignal::CANCEL) { // TODO: remove CANCEL
+      VLOG(5) << "Closing ConsumerBase subscriber with calling onComplete";
       subscriber->onComplete();
     } else {
+      VLOG(5) << "Closing ConsumerBase subscriber with calling onError";
       subscriber->onError(StreamInterruptedException(static_cast<int>(signal)));
     }
   }
--- a/rsocket/statemachine/ConsumerBase.cpp
+++ b/rsocket/statemachine/ConsumerBase.cpp
@@ -32,7 +32,8 @@ void ConsumerBase::checkConsumerRequest() {
   CHECK((state_ == State::RESPONDING) == !!consumingSubscriber_);
 }
 
-// TODO: this is probably buggy and misused and not needed (when completeConsumer exists)
+// TODO: this is probably buggy and misused and not needed (when
+// completeConsumer exists)
 void ConsumerBase::cancelConsumer() {
   state_ = State::CLOSED;
   consumingSubscriber_ = nullptr;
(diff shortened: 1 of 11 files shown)


test/RequestResponseTest.cpp
```diff
@@ -157,6 +157,23 @@ TEST(RequestResponseTest, RequestOnDisconnectedClient) {
   ASSERT(did_call_on_error);
 }
 
+// TODO: test that multiple requests on a requestResponse
+// fail in a well-defined way (right now it'd nullptr deref)
+TEST(RequestResponseTest, MultipleRequestsError) {
```
dymk: oops, this should go in a separate commit; ignore this

test/RequestStreamTest_concurrency.cpp
```diff
@@ -0,0 +1,144 @@
...
+using namespace yarpl;
+using namespace yarpl::flowable;
+using namespace rsocket;
+using namespace rsocket::tests;
+using namespace rsocket::tests::client_server;
+
+struct LockstepBatons {
+  folly::Baton<> on_second_payload_sent;
+  folly::Baton<> on_cancel_sent;
+  folly::Baton<> on_cancel_recieved_toserver;
```
alexmalyshev: Nit: `s|recieved|received|g`.  Also the usual nag about camelCase.

test/RequestStreamTest_concurrency.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+#define LOCKSTEP_DEBUG(expr) VLOG(3) << expr
+
+class LockstepAsyncHandler : public rsocket::RSocketResponder {
+  LockstepBatons& batons_;
+  Sequence& subscription_seq_;
+
+ public:
+  LockstepAsyncHandler(LockstepBatons& batons, Sequence& subscription_seq)
+      : batons_(batons), subscription_seq_(subscription_seq){};
```
alexmalyshev: Nit: Rogue semicolon.

test/RSocketTests.h
```diff
@@ -8,6 +8,11 @@
 
 #include "rsocket/transports/tcp/TcpConnectionFactory.h"
 
+auto const default_baton_timeout = std::chrono::milliseconds(100);
+#define CHECK_WAIT(baton) CHECK(baton.timed_wait(default_baton_timeout))
+// #define CHECK_WAIT(baton) baton.timed_wait(default_baton_timeout)
```
alexmalyshev: I get that they're commented out so they can be easily swapped in, but I don't think it's necessary and clutters things up.

test/RequestStreamTest_concurrency.cpp
```diff
@@ -0,0 +1,144 @@
...
+  Sequence server_seq;
+  Sequence client_seq;
+
+  auto server =
+      makeServer(std::make_shared<LockstepAsyncHandler>(batons, server_seq));
+  folly::ScopedEventBaseThread worker;
+  auto client = makeClient(worker.getEventBase(), *server->listeningPort());
+  auto requester = client->getRequester();
+
+  auto scbr_mock =
```
alexmalyshev: Nit: Just name it `subscriber`, it's clearer and we can afford the vowels :)

rsocket/statemachine/ConsumerBase.cpp
```diff
@@ -45,11 +47,14 @@ void ConsumerBase::generateRequest(size_t n) {
 }
 
 void ConsumerBase::endStream(StreamCompletionSignal signal) {
+  VLOG(5) << "ConsumerBase::endStream(" << signal << ")";
```
alexmalyshev: :+1:

test/RSocketTests.h
```diff
@@ -8,6 +8,9 @@
 
 #include "rsocket/transports/tcp/TcpConnectionFactory.h"
 
+auto const default_baton_timeout = std::chrono::milliseconds(100);
```
alexmalyshev: Nit: `constexpr`

test/RequestResponseTest.cpp
alexmalyshev: So is this being removed?

PR #710, opened 2017-09-01
Delete StreamStateMachineBase::Parameters

alexmalyshev:
It's just two parameters, pass them manually without all the hassle.

Gets rid of some unneeded "Base" typedefs along the way, kills commented out
code, removes some unused headers, and of course runs clang-format.


alexmalyshev:
Following up on this, I'm gonna try to remove the uses of `std::shared_ptr<StreamsWriter>`, and see if it makes sense to replace them with `StreamsWriter&`, as the individual state machine objects should not persist past `RSocketStateMachine` (at least I hope they don't have to).

PR #711, opened 2017-09-01
add test/fix for racing to call 'cancel' on client subscription

dymk:
Also fixes a bug in FlowableOperator, which wasn't calling onComplete or onError of its Subscription if cancel() was called on it (another bug related to inheriting, and sharing state between, a Subscriber and Subscription in one object. Multiple inheritance is the devil).

### DIFF
--- a/rsocket/statemachine/ConsumerBase.cpp
+++ b/rsocket/statemachine/ConsumerBase.cpp
@@ -22,7 +22,9 @@ void ConsumerBase::subscribe(
   }
 
   DCHECK(!consumingSubscriber_);
-  consumingSubscriber_ = std::move(subscriber);
+  // keep a refernce to consumingSubscriber_ on the stack in case
+  // its call to onSubscribe causes us to drop our instance reference to it
+  consumingSubscriber_ = subscriber;
   consumingSubscriber_->onSubscribe(get_ref(this));
 }
 
@@ -48,6 +50,7 @@ void ConsumerBase::generateRequest(size_t n) {
 
 void ConsumerBase::endStream(StreamCompletionSignal signal) {
   VLOG(5) << "ConsumerBase::endStream(" << signal << ")";
+  auto this_ = get_ref(this);
   if (auto subscriber = std::move(consumingSubscriber_)) {
     if (signal == StreamCompletionSignal::COMPLETE ||
         signal == StreamCompletionSignal::CANCEL) { // TODO: remove CANCEL
@@ -71,7 +74,10 @@ void ConsumerBase::processPayload(Payload&& payload, bool onNext) {
     // figuring out flow control allowance.
     if (allowance_.tryAcquire()) {
       sendRequests();
-      consumingSubscriber_->onNext(std::move(payload));
+      // keep a reference to the subscriber on the stack in case it causes us to drop
+      // our instance's reference to it
+      auto ref = consumingSubscriber_;
+      ref->onNext(std::move(payload));
     } else {
       handleFlowControlError();
       return;
--- a/test/RequestStreamTest_CancelAfterRequest.cpp
+++ b/test/RequestStreamTest_CancelAfterRequest.cpp
@@ -1,0 +1,157 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#include <folly/io/async/ScopedEventBaseThread.h>
+#include <gtest/gtest.h>
+#include <random>
+#include <thread>
+
+#include "RSocketTests.h"
+#include "yarpl/Flowable.h"
+#include "yarpl/flowable/TestSubscriber.h"
+
+#include "yarpl/test_utils/Mocks.h"
+
+using namespace yarpl;
+using namespace yarpl::flowable;
+using namespace rsocket;
+using namespace rsocket::tests;
+using namespace rsocket::tests::client_server;
+
+using namespace yarpl::mocks;
+using namespace ::testing;
+
+struct Batons {
+  folly::Baton<> serverFinished;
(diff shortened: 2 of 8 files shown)


test/RequestStreamTest_CancelAfterRequest.cpp
```diff
@@ -0,0 +1,157 @@
...
+    auto at_response = 0;
+    Reference<Subscription> subscription;
+
+    auto subscriber = Subscribers::create<std::string>(
+        // onSubscribe
+        [&](Reference<Subscription> subscription_) {
+          subscription = subscription_;
+          subscription->request(will_request);
+
+          auto chance = dis_0_100(rand);
```
alexmalyshev: Nit: Just use `folly::Random::rand32(100)`, much shorter, doesn't need the distribution and generator objects.

rsocket/statemachine/ConsumerBase.cpp
```diff
@@ -22,7 +22,9 @@ void ConsumerBase::subscribe(
   }
 
   DCHECK(!consumingSubscriber_);
-  consumingSubscriber_ = std::move(subscriber);
+  // keep a refernce to consumingSubscriber_ on the stack in case
+  // its call to onSubscribe causes us to drop our instance reference to it
```
alexmalyshev: How does this, and all follow-up cases help? I'm confused how having the extra reference on the stack is useful, given that this is the last operation we're doing in the function anyway?

test/RequestStreamTest_CancelAfterRequest.cpp
dymk: We want something pseudorandom that we control the seed to, though. I couldn't find docs on folly::Random for how to do that :(

yarpl/include/yarpl/flowable/Subscribers.h
```diff
@@ -62,6 +62,23 @@ class Subscribers {
...
   }
 
+  template <
+      typename T,
+      typename Sub,
+      typename Next,
+      typename Error,
+      typename Complete,
+      typename = typename std::enable_if<
+          std::is_callable<Sub(Reference<Subscription>), void>::value &&
```
alexmalyshev: I really gotta get on that `s|enable_if|static_assert|g` killing spree I've been planning on doing...

rsocket/statemachine/ConsumerBase.cpp
dymk: because the object that we're calling methods on might cause itself to be free'd, and then access uninitialized memory when it references `this`. We can't rely on users adding `auto this_ = get_ref(this)` as a prologue to methods in all of their classes that inherit from Subscription/Subscriber.

alexmalyshev: Yeah you're right... ugh this sucks though.  I guess it's more impetus for `onSubscribe` versus `onSubscribeImpl`.

yarpl/include/yarpl/flowable/Subscribers.h
```diff
@@ -128,6 +145,48 @@ class Subscribers {
...
+      Subscriber<T>::onSubscribe(s);
+      sub_(s);
+    }
+
+    void onNext(T elem) override {
+      next_(std::move(elem));
+    }
+
+    void onComplete() override {
+      complete_();
```
lehecka: should you call Subscriber<T>::onComplete() to release the subscription?

yarpl/include/yarpl/flowable/Subscribers.h
```diff
@@ -128,6 +145,48 @@ class Subscribers {
...
+    }
+
+    void onError(folly::exception_wrapper ex) override {
+      error_(std::move(ex));
```
lehecka: same here

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -51,7 +51,8 @@ class FlowableOperator : public Flowable<D> {
 
     void subscriberOnNext(D value) {
       if (subscriber_) {
-        subscriber_->onNext(std::move(value));
+        auto ref = subscriber_;
```
lehecka: curious whether we have a test case covering this case. Do we need this here?
I just feel like we are sprinkling stack ref copies to a bunch of places and I don't know if we we are safe now or still not yet. I also don't know any better solution.

dymk: You're right, it might not be sufficient to sprinkle these just in the locations where we call methods on the `subscriber_`; we need a better solution.

test/RequestStreamTest_CancelAfterRequest.cpp
alexmalyshev: Oh you're right, I didn't notice you were controlling the seed.  I don't see a way of doing that with `folly::Random` either :(

yarpl/include/yarpl/flowable/FlowableOperator.h
alexmalyshev: Can't believe I'm saying this but I'm wondering if the best path forward for now is to disable the failing tests (there should only be two in trunk, `RequestStreamTest.OperationsAfterCancel` and `Observable.CancelSubscriptionChain`).

If the issue is more systemic than a few lines of code can fix, maybe we should table any "better" solutions for later.

alexmalyshev:
@dymk - Is this still relevant?  Should we close it out?

dymk:
It's still relevant in that the tests are useful to have, but the implementation is likely pretty out of date, and needs a rebase. Will reopen in the future once it's rebased and brought up to date.

PR #714, opened 2017-09-02
Test setting credits to canceled again in EmitterSubscription::on{Com…

alexmalyshev:
…plete,Error}

Curious what Travis will think about it.


alexmalyshev:
Okay, didn't help at all.

PR #715, opened 2017-09-02
Disable the two tests that are repeatedly failing in trunk

alexmalyshev:
We're not able to commit any type of fixes while these are here.  Let's stop the
bleeding for the immediate short term.


alexmalyshev:
Travis, get a hold of yourself.  All test runs are green but you're saying they're not completed yet.

PR #716, opened 2017-09-04
Sharing ConnectionFactory instance for RSocketClient instances

lehecka:
The core of this PR is change of unique_ptr<ConnectionFactory> to shared_ptr<ConnectionFactory> in the constructor methods of RSocketClient. The Reason for this is that in the case of HTTPConnectionFactory if we reuse the same instance, we can reuse the same HTTP session object and within the session object we can just create HTTP streams for each duplex connection.

Along the way I noticed some unused parameters in the ctor of RSocketClient so I removed them.

I also removed the awkward call sequence for createConnectedClient. Now it is straightforward that after we get DuplexConnection instance we create RSocketClient instance. It allowed cleaning up of setupParams_ member variable which was used only from the connect method.

### DIFF
--- a/rsocket/RSocket.cpp
+++ b/rsocket/RSocket.cpp
@@ -2,77 +2,106 @@
 
 #include "rsocket/RSocket.h"
 
-#include <folly/io/async/EventBaseManager.h>
-
 namespace rsocket {
 
-folly::Future<std::shared_ptr<RSocketClient>> RSocket::createConnectedClient(
-    std::unique_ptr<ConnectionFactory> connectionFactory,
+folly::Future<std::unique_ptr<RSocketClient>> RSocket::createConnectedClient(
+    std::shared_ptr<ConnectionFactory> connectionFactory,
     SetupParameters setupParameters,
     std::shared_ptr<RSocketResponder> responder,
     std::unique_ptr<KeepaliveTimer> keepaliveTimer,
     std::shared_ptr<RSocketStats> stats,
     std::shared_ptr<RSocketConnectionEvents> connectionEvents,
     std::shared_ptr<ResumeManager> resumeManager,
-    std::shared_ptr<ColdResumeHandler> coldResumeHandler,
-    OnRSocketResume) {
-  auto c = std::shared_ptr<RSocketClient>(new RSocketClient(
-      std::move(connectionFactory),
-      std::move(setupParameters),
-      std::move(responder),
-      std::move(keepaliveTimer),
-      std::move(stats),
-      std::move(connectionEvents),
-      std::move(resumeManager),
-      std::move(coldResumeHandler)));
+    std::shared_ptr<ColdResumeHandler> coldResumeHandler) {
+  auto createRSC = [
+      connectionFactory,
+      setupParameters = std::move(setupParameters),
+      responder = std::move(responder),
+      keepaliveTimer = std::move(keepaliveTimer),
+      stats = std::move(stats),
+      connectionEvents = std::move(connectionEvents),
+      resumeManager = std::move(resumeManager),
+      coldResumeHandler = std::move(coldResumeHandler)](
+      ConnectionFactory::ConnectedDuplexConnection connection) mutable {
+    VLOG(3) << "createConnectedClient received DuplexConnection";
+    return RSocket::createClientFromConnection(
+        std::move(connection.connection),
+        connection.eventBase,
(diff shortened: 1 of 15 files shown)


rsocket/RSocket.cpp
```diff
@@ -2,77 +2,106 @@
...
-  return c->connect().then([c]() mutable { return c; });
+  return connectionFactory->connect().then(
+      [createRSC = std::move(createRSC)](
+          ConnectionFactory::ConnectedDuplexConnection connection) mutable {
+        // fromConnection method must be called from the eventBase and since
+        // there is no guarantee that the Future returned from the
+        // connectionFactory::connect method is executed on the event base,
+        // we have to ensure it by using folly::via
+        auto* eventBase = &connection.eventBase;
+        return via(eventBase).then(
```
alexmalyshev: Nit: You can do `return folly::via(eventBase, [...] { ... });` and drop the `.then()`.

rsocket/RSocket.cpp
```diff
@@ -2,77 +2,106 @@
...
     std::shared_ptr<RSocketStats> stats,
     std::shared_ptr<RSocketConnectionEvents> connectionEvents,
     std::shared_ptr<ResumeManager> resumeManager,
-    std::shared_ptr<ColdResumeHandler> coldResumeHandler,
-    OnRSocketResume) {
-  auto c = std::shared_ptr<RSocketClient>(new RSocketClient(
-      nullptr,
-      std::move(setupParameters),
+    std::shared_ptr<ColdResumeHandler> coldResumeHandler) {
+  auto c = std::unique_ptr<RSocketClient>(new RSocketClient(
```
alexmalyshev: Nit: `std::make_unique<RSocketClient>(...)`

rsocket/RSocketClient.cpp
```diff
@@ -14,52 +14,33 @@ using namespace folly;
...
       responder_(std::move(responder)),
       keepaliveTimer_(std::move(keepaliveTimer)),
       stats_(stats),
       connectionEvents_(connectionEvents),
       resumeManager_(resumeManager),
       coldResumeHandler_(coldResumeHandler),
-      protocolVersion_(setupParameters_.protocolVersion),
-      token_(setupParameters_.token) {}
+      protocolVersion_(protocolVersion),
+      token_(token) {}
```
alexmalyshev: `token_(std::move(token))` since `token` is a `std::vector`.

rsocket/RSocketClient.h
```diff
@@ -51,34 +51,28 @@ class RSocketClient {
...
+      std::shared_ptr<ResumeManager> resumeManager,
+      std::shared_ptr<ColdResumeHandler> coldResumeHandler);
 
   // Create stateMachine with the given DuplexConnection
   void fromConnection(
       std::unique_ptr<DuplexConnection> connection,
-      folly::EventBase& eventBase);
+      folly::EventBase& eventBase,
+      SetupParameters setupParameters
+  );
```
alexmalyshev: Nit: I have a feeling `clang-format` is going to change this, but whatever, it won't trigger lint errors.

rsocket/framing/FrameSerializer.cpp
```diff
@@ -20,7 +20,7 @@ namespace rsocket {
 constexpr const ProtocolVersion ProtocolVersion::Latest =
     FrameSerializerV1_0::Version;
 
-ProtocolVersion FrameSerializer::getCurrentProtocolVersion() {
+const ProtocolVersion ProtocolVersion::Current() {
```
alexmalyshev: There's no point putting `const` on a value return, it does nothing.

rsocket/internal/Common.h
```diff
@@ -124,6 +124,7 @@ struct ProtocolVersion {
 
   static const ProtocolVersion Unknown;
   static const ProtocolVersion Latest;
+  static const ProtocolVersion Current();
```
alexmalyshev: Ditto about the `const` return.

rsocket/RSocket.cpp
lehecka: ctor of RSocketClient is private..

rsocket/internal/Common.h
somasun: Like where this function resides now.  It did not make sense to have it in FrameSerializer.

PR #717, opened 2017-09-04
Drop clang-4.0 package from Travis

alexmalyshev:
`apt-get` is complaining that it doesn't exist.  Seems like all we need is the
`llvm-toolchain-trusty-4.0` package.


alexmalyshev:
Hmm no.  Despite the description on llvm.org, `llvm-toolchain-trusty-4.0` doesn't seem to provide a `clang-4.0` binary, at least not under that name...

PR #718, opened 2017-09-05
Fix use-after-free in Mocks

dymk:
should fix the flakeyness in `RequestStreamTest.OperationsAfterCancel`

### DIFF
--- a/yarpl/test/yarpl/test_utils/Mocks.h
+++ b/yarpl/test/yarpl/test_utils/Mocks.h
@@ -50,6 +50,7 @@ class MockSubscriber : public flowable::Subscriber<T> {
   void onSubscribe(
       yarpl::Reference<flowable::Subscription> subscription) override {
     subscription_ = subscription;
+    auto this_ = get_ref(this);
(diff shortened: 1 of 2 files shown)


yarpl/test/yarpl/test_utils/Mocks.h
```diff
@@ -50,6 +50,7 @@ class MockSubscriber : public flowable::Subscriber<T> {
   void onSubscribe(
       yarpl::Reference<flowable::Subscription> subscription) override {
     subscription_ = subscription;
+    auto this_ = get_ref(this);
```
alexmalyshev: Nit: Suffixing the identifier with a `_` collides with the convention for private field naming.  Maybe `self`, `this2`, or `thiz`?  Don't want to bikeshed too hard...

dymk: prefix with _ maybe? `_this`. It's at least pretty obvious that this isn't a private member, and it'll always be declared in function rather than class scope.

PR #719, opened 2017-09-05
Ensure cold-resume always happens in the right evb

somasun:
https://github.com/rsocket/rsocket-cpp/pull/629/files changed the connectionFactory.connect() to return a Future.  As a consequence, the evb in which the stateMachine inside resume() gets created depends on the ordering of when the connect() Future is fulfilled and when the .then() call is registered.  This PR is to ensure the stateMachine always gets created in the right evb.


dymk:
LGTM

dymk:
... oops

PR #720, opened 2017-09-05
remove racey checks on what should be idempotent calls

dymk:
We're accessing shared state between a Subscriber and a Subscription which indicates if the Subscriber has completed yet; another thread calling cancel() on this Subscription will crash if the StreamRequester thinks it's already in a closed state. 

We shouldn't be making this check at all, as we can't assume any ordering guarantees on calls made across Subscriber/Subscription.


alexmalyshev:
> ... another thread calling cancel() on this Subscription will crash ...

The stream state machines should all be accessed from the same thread.  But I agree that we shouldn't do this check in `request()`, as for all we know, the EventBase just processed an `onComplete()` right before scheduling a `request()`.

PR #723, opened 2017-09-06
make get_ref a 'covariant' method on objects rather than a free function

dymk:
because let's not keep around a confusing API which may, hypothetically, lead to a memory leak

### DIFF
--- a/rsocket/framing/FrameTransport.cpp
+++ b/rsocket/framing/FrameTransport.cpp
@@ -31,7 +31,7 @@ void FrameTransport::connect() {
   }
 
   connectionOutput_ = connection_->getOutput();
-  connectionOutput_->onSubscribe(yarpl::get_ref(this));
+  connectionOutput_->onSubscribe(this->ref_from_this(this));
 
   // The onSubscribe call on the previous line may have called the terminating
   // signal which would call disconnect/close.
@@ -43,7 +43,7 @@ void FrameTransport::connect() {
     // will create a hard reference for that case and keep the object alive
     // until setInput method returns
     auto connectionCopy = connection_;
-    connectionCopy->setInput(yarpl::get_ref(this));
+    connectionCopy->setInput(this->ref_from_this(this));
   }
 }
 
--- a/yarpl/include/yarpl/Refcounted.h
+++ b/yarpl/include/yarpl/Refcounted.h
@@ -40,11 +40,11 @@ class Refcounted {
   template <typename U>
   friend class Reference;
 
-  void incRef() {
+  void incRef() const {
     refcount_.fetch_add(1, std::memory_order_relaxed);
   }
 
-  void decRef() {
+  void decRef() const {
     auto previous = refcount_.fetch_sub(1, std::memory_order_relaxed);
     assert(previous >= 1 && "decRef on a destroyed object!");
     if (previous == 1) {
@@ -54,7 +54,7 @@ class Refcounted {
   }
 
   // refcount starts at 1 always, so we don't destroy ourselves in
-  // the constructor if we call `get_ref` in it
+  // the constructor if we call `ref_from_this` in it
   mutable std::atomic_size_t refcount_{1};
 };
 
@@ -85,7 +85,7 @@ class Reference {
      class MyClass : Refcounted {
        MyClass() {
         // count() == 0
-        auto r = get_ref<MyClass>(this)
+        auto r = ref_from_this<MyClass>(this)
         // count() == 1
         do_something_with(r);
         // if do_something_with(r) doens't keep a reference to r somewhere, then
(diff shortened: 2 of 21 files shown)


dymk:
I'd prefer making `get_ref` a method on Refcounted rather than this new class, but there's issues casting `this` in a virtually inherited class apparently.

rsocket/framing/FrameTransport.cpp
```diff
@@ -31,7 +31,7 @@ void FrameTransport::connect() {
   }
 
   connectionOutput_ = connection_->getOutput();
-  connectionOutput_->onSubscribe(yarpl::get_ref(this));
+  connectionOutput_->onSubscribe(this->get_ref(this));
```
phoad: `get_ref_from_this()`

why do you need to give `this` to `get_ref` function if you already call it on `this`

dymk: C++ doesn't have covariant return types; eg the return type of `get_ref` can't be deduced to anything but `enable_get_ref`, so we have to provide the subclass type by way of passing in `this`.

We _could_ have `enable_get_ref` be parameterized over the class inheriting from it, but this doesn't work well when a further derived class wants to call `get_this`; you end up with calls like:

```cpp
class MyFooBase : public enable_get_ref<MyFooBase> {};
class MyBarBase : public enable_get_ref<MyBarBase> {};

class Derived : MyFooBase, MyBarBase {
  void doSomething() {
    auto self = this->template yarpl::enable_this<MyFooBase>get_ref(); // :(
  }
};
```

which just isn't going to fly. The same problem pops up when a parent class `Bar` inherits from `enable_get_ref`, and a child class `Foo` wants to get a `Reference<Foo>`. 

So we do something which is at least consistent, which is (at runtime, on dev builds) verify that in `enable_get_ref::get_ref`, `this` can be downcasted to the concrete type of `ptr`, and check that the caller of `get_ref` is indeed getting a caller of this instance. We basically enforce the semantics of `shared_from_this` this way. 

I didn't rename this to `get_ref_from_this` because it's pretty explicit what you're getting a reference to, and that'd be three instance of the word `this` per line.

yarpl/include/yarpl/flowable/Flowable.h
```diff
@@ -120,41 +120,41 @@ template <typename T>
...
-      get_ref(this), std::move(function));
+      this->ref_from_this(this), std::move(function));
 }
 
 template <typename T>
 template <typename Function>
 Reference<Flowable<T>> Flowable<T>::filter(Function function) {
   return make_ref<FilterOperator<T, Function>>(
-      get_ref(this), std::move(function));
+      this->ref_from_this(this), std::move(function));
```
lehecka: just curious, is the "this->" necessary?

yarpl/include/yarpl/observable/Observer.h
```diff
@@ -13,7 +13,7 @@ namespace yarpl {
 namespace observable {
 
 template <typename T>
-class Observer : public virtual Refcounted {
+class Observer : public virtual Refcounted, public yarpl::enable_get_ref {
```
lehecka: curious.. inheriting from yarpl::enable_get_ref doesn't have to be public, right? its just for the class internal stuff..

yarpl/include/yarpl/Refcounted.h
```diff
@@ -318,14 +318,32 @@ Reference<CastTo> make_ref(Args&&... args) {
   );
 }
 
-template <typename T>
-Reference<T> get_ref(T* object) {
-  static_assert(
-      std::is_base_of<Refcounted, std::decay_t<T>>::value,
-      "Reference can only be constructed with a Refcounted object");
+class enable_get_ref {
```
lehecka: I believe linter will complain that you should add a virtual dtor for this class... can you disable the warning?

yarpl/include/yarpl/Refcounted.h
```diff
@@ -318,14 +318,32 @@ Reference<CastTo> make_ref(Args&&... args) {
...
+ private:
+  // force the class to be polymorphic so we can dynamic_cast<T>(this)
+  virtual void dummy_internal_get_ref() {}
```
lehecka: is it true that you need this virtual method only in DEBUG build?
do you trust compiler to optimize out the vtable if the method is not used in RELEASE? or should we add #ifdef?

yarpl/include/yarpl/Refcounted.h
```diff
@@ -318,14 +318,32 @@ Reference<CastTo> make_ref(Args&&... args) {
...
+
+ protected:
+  // materialize a reference to 'this', but a type even further derived from
+  // Derived, because C++ doesn't have covariant return types on methods
+  template <typename As>
+  Reference<As> ref_from_this(As* ptr) {
```
lehecka: do we need overload with "const As*" ? so we can call ref_from_this in a const method?

lehecka:
generally I like this approach and that enable_get_ref is not templated. Nice work!

yarpl/include/yarpl/observable/Observer.h
dymk: It unfortunately does, or the compiler complains about illegally casting to the base class.

yarpl/include/yarpl/Refcounted.h
dymk: This is a good point, I'll play around with making this conditional on a DEBUG build.

yarpl/include/yarpl/Refcounted.h
dymk: Also a good point; we don't really have many uses of 'const' in the codebase, but that's not exactly something to strive towards maintaining. I'll add a 'const' version.

yarpl/include/yarpl/flowable/Flowable.h
dymk: It's not; it'd only be required if `enable_get_ref` was templated. All instances being prefixed with `this->` are a holdover from a version which did have a templated `enable_get_ref`.

dymk:
@alexmalyshev agreed, the one benefit about Reference that I can think off of the top of my head is ability to get a reference from `this` in the constructor (which isn't legal with `shared_from_this`). Instances of this are easily fixable though with moving ctor logic `init()` method.

alexmalyshev:
> the one benefit about Reference that I can think off of the top of my head is ability to get a reference from this in the constructor (which isn't legal with shared_from_this). 

Is that really true?  I've never seen that limitation.  The one I know of is not being able to use it with virtual inheritance (we used to have `EnableSharedFromThisBase` to finagle around that).

More conversation about this in https://github.com/rsocket/rsocket-cpp/issues/551

dymk:
Indeed, the hidden shared_pointer field added by `enable_shared_from_this` hasn't been initialized by the time the constructor has been called, so it's not valid to call `shared_from_this` in the constructor. 

https://stackoverflow.com/questions/31924396/why-shared-from-this-cant-be-used-in-constructor-from-technical-standpoint

`shared_ptr` also runs into the covariance issue; I'm not sure what the standard way is to get a shared pointer to an `A` in this case without inheriting multiple times from `enable_shared_from_this`: 

```cpp
class B : public enable_shared_from_this<B> {};
class A : public B {
  auto get_a() { return shared_from_this(); } // returns a shared_ptr<B>, we want shared_ptr<A>
}
```

dymk:
Also enables `Reference<T const>` objects, which previously would not have been allowed (even though the atomic counter was `mutable`)

PR #725, opened 2017-09-06
Fix the crash when an unknown client connects to RSocketServer

phoad:
When FramedReader does not accept a client, it causes onError call, which calls Subscriber::onError. In that function, it checks if there is any subscription or not.

In the unrecognized client case, there is no subscription available, so this call causes crash. So we need to check if the `subscription` is available before calling onError on the Subscriber class.

### DIFF
--- a/rsocket/framing/FramedReader.cpp
+++ b/rsocket/framing/FramedReader.cpp
@@ -152,7 +152,9 @@ void FramedReader::onComplete() noexcept {
 void FramedReader::onError(folly::exception_wrapper ex) noexcept {
   completed_ = true;
   payloadQueue_.move(); // equivalent to clear(), releases the buffers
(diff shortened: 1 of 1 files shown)


rsocket/framing/FramedReader.cpp
```diff
@@ -152,7 +152,9 @@ void FramedReader::onComplete() noexcept {
 void FramedReader::onError(folly::exception_wrapper ex) noexcept {
   completed_ = true;
   payloadQueue_.move(); // equivalent to clear(), releases the buffers
-  DuplexConnection::Subscriber::onError(folly::exception_wrapper());
+  if (DuplexConnection::Subscriber::subscription()) {
```
somasun: Do you want to add some comments on when this can happen ?

PR #726, opened 2017-09-06
RSocketClient::disconnect() should return a future

alexmalyshev:
Right now it does the disconnect asynchronously, with no good way to force it to
be synchronous.  Similarly, ~RSocketClient didn't call disconnect(), so it seems
like it would destroy the state machine on a different thread than the
ConnectionFactory worker thread.

### DIFF
--- a/rsocket/RSocketClient.cpp
+++ b/rsocket/RSocketClient.cpp
@@ -37,6 +37,7 @@ RSocketClient::RSocketClient(
 
 RSocketClient::~RSocketClient() {
   VLOG(4) << "RSocketClient destroyed ..";
+  disconnect().get();
 }
 
 const std::shared_ptr<RSocketRequester>& RSocketClient::getRequester() const {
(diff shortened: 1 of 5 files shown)


rsocket/RSocketClient.cpp
```diff
@@ -116,13 +117,24 @@ folly::Future<folly::Unit> RSocketClient::resume() {
...
-void RSocketClient::disconnect(folly::exception_wrapper ex) {
-  CHECK(stateMachine_);
-  evb_->runInEventBaseThread(
-      [ stateMachine = stateMachine_, ex = std::move(ex) ]() mutable {
-        VLOG(2) << "Disconnecting RSocketStateMachine on EventBase";
-        stateMachine->disconnect(std::move(ex));
-      });
+folly::Future<folly::Unit> RSocketClient::disconnect(
+    folly::exception_wrapper ew) {
+  if (!stateMachine_) {
```
somasun: This should probably be a CHECK().  If I remember correctly, you cannot create a RSocketClient instance without creating a stateMachine_

dymk: What about returning an Error future in this case instead?

alexmalyshev: Oh whoops, this is from an older part of my diff when I was (mistakenly) moving the state machine into the disconnect lambda.  Let me fix that.

alexmalyshev: @dymk I made it a `DCHECK`, as this is kind of an invariant violation on our part.  I guess I could make it return an exception just in case, think it's worth it?

dymk: Well, it's not an invariant that we can enforce internally, as the application is the one with a direct handle on the client. So it's probably worth it to throw something catchable rather than something that aborts the program.

alexmalyshev: Works for me.

PR #729, opened 2017-09-07
Make RSocketConnectionManager unit testable and add unit tests for it

phoad:
I have removed the hack of returning ConnectionEvents with ConnectionEventsWrapper in RSocketConnectionManager.

I have made the RSocketConnectionManager be easy to unit test.

I have added some tests to verify the functionality.

### DIFF
--- a/rsocket/internal/ManageableConnection.h
+++ b/rsocket/internal/ManageableConnection.h
@@ -1,0 +1,53 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+#pragma once
+
+#include <folly/ExceptionWrapper.h>
+#include <folly/futures/SharedPromise.h>
+#include <folly/io/async/EventBase.h>
+
+namespace rsocket {
+
+class MockManageableConnection;
+enum class StreamCompletionSignal;
+
+/// Utility class that enables RSocketConnectionManager to handle
+/// lifetime of RSocketStateMachine objects.
+class ManageableConnection {
+public:
+  virtual ~ManageableConnection() {
+    if (!closePromise_.isFulfilled()) {
+      // don't let any object indefinitely wait for the close event.
+      closePromise_.setValue();
+    }
+  }
+
+  /// Listen for the close event.
+  ///
+  /// The proposed usage is utilizing `via` function and appending
+  /// more functionality with `then` function.
+  ///
+  /// \return a Future to bind to.
+  folly::Future<folly::Unit> listenCloseEvent() {
+    return closePromise_.getFuture();
+  }
+
+  /// Terminates underlying connection.
+  ///
+  /// This may synchronously deliver terminal signals to all
+  /// StreamAutomatonBase attached to this ConnectionAutomaton.
+  virtual void close(folly::exception_wrapper,
+                     StreamCompletionSignal) = 0;
+
+protected:
+  void onClose(folly::exception_wrapper) {
+    // ignore the exception
+    closePromise_.setValue();
+  }
+
+private:
+  folly::SharedPromise<folly::Unit> closePromise_;
+
+  friend MockManageableConnection;
+};
+
+} // namespace rsocket
--- a/rsocket/internal/RSocketConnectionManager.cpp
+++ b/rsocket/internal/RSocketConnectionManager.cpp
@@ -7,15 +7,14 @@
 #include <folly/io/async/EventBase.h>
 
 #include "rsocket/RSocketConnectionEvents.h"
-#include "rsocket/statemachine/RSocketStateMachine.h"
+#include "rsocket/internal/Common.h"
+#include "rsocket/internal/ManageableConnection.h"
 
 namespace rsocket {
 
 RSocketConnectionManager::~RSocketConnectionManager() {
   VLOG(1) << "Started ~RSocketConnectionManager";
-  SCOPE_EXIT {
-    VLOG(1) << "Finished ~RSocketConnectionManager";
-  };
+  SCOPE_EXIT { VLOG(1) << "Finished ~RSocketConnectionManager"; };
 
   StateMachineMap map;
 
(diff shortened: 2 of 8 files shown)


rsocket/statemachine/RSocketStateMachine.h
```diff
@@ -58,6 +59,7 @@ class FrameSink {
 class RSocketStateMachine final
     : public FrameSink,
       public FrameProcessor,
+      public ManageableConnection,
```
dymk: Can we have RSocketStateMachine compose, rather than inherit, from another class? It's already difficult to tell what `override` methods correspond to what base class.

rsocket/internal/ManageableConnection.h
```diff
@@ -0,0 +1,53 @@
...
+#include <folly/io/async/EventBase.h>
+
+namespace rsocket {
+
+class MockManageableConnection;
+enum class StreamCompletionSignal;
+
+/// Utility class that enables RSocketConnectionManager to handle
+/// lifetime of RSocketStateMachine objects.
+class ManageableConnection {
```
dymk: Does this need to be its own separate class? It's just a wrapper around a promise at this point.

phoad: Better to be simple :)

rsocket/statemachine/RSocketStateMachine.h
phoad: I did not add any functions to RSocketStateMachine and actually removed a hack-y function. I will take a look but it might be another improvement on top of this. I want to keep my changes limited.

dymk: (y) fair enough

rsocket/internal/ManageableConnection.h
dymk: Why is introducing a new class and inheriting from it simpler than just putting the promise to indicate "is the connection closed yet" in RSocketStateMachine directly?

rsocket/internal/ManageableConnection.h
```diff
@@ -0,0 +1,53 @@
...
+  folly::Future<folly::Unit> listenCloseEvent() noexcept {
+    return closePromise_.getFuture();
+  }
+
+  /// Terminates underlying connection.
+  ///
+  /// This may synchronously deliver terminal signals to all
+  /// StreamAutomatonBase attached to this ConnectionAutomaton.
+  virtual void close(folly::exception_wrapper,
+                     StreamCompletionSignal) noexcept = 0;
```
alexmalyshev: Regardless of whether we use `ManageableConnection` or a `folly::Promise` directly, we should not be using `noexcept`.  It has no effect on typechecking, and makes it a lot harder to debug things if an exception is thrown.

rsocket/internal/RSocketConnectionManager.cpp
```diff
@@ -37,9 +36,9 @@ RSocketConnectionManager::~RSocketConnectionManager() {
     map.swap(*locked);
   }
 
-  for (auto& kv : map) {
+  for (auto &kv : map) {
```
alexmalyshev: Why `auto &kv` (and all following uses of `&` later in this file)?  The rest of the codebase does `auto& kv`.

rsocket/internal/ManageableConnection.h
```diff
@@ -0,0 +1,53 @@
...
+
+protected:
+  void onClose(folly::exception_wrapper) noexcept {
+    // ignore the exception
+    closePromise_.setValue();
+  }
+
+private:
+  folly::SharedPromise<folly::Unit> closePromise_;
```
alexmalyshev: Why a `SharedPromise`?  Don't we expect there to be a single call of `RSocketConnectionManager::manageConnection()` per connection lifetime?

Also doesn't this promise need to be reset to a new one after closing?  What happens if we do SETUP -> CLOSE -> RESUME -> CLOSE?  We'll set the same promise twice, right?

dymk:
@alexmalyshev in-band discussion with Fuat has convinced me that it makes sense to inherit from another interface for now, so the ConnectionManager can have a shared pointer to the RSocketStateMachine (but as a shared_pointer<ManagedConnection>). 

Design nit, but it also could make sense to have ManagedConnection be a just an interface, and put the Promise in RSocketStateMachine, if that's the only bit of state that ManagedConnection cares about, and forego inheritance-as-composition until we can actually compose objects properly.

rsocket/internal/ManageableConnection.h
phoad: I thought CLOSE is terminal, I saw that comment in the RSocketConnectionEvents interface. Is this comment wrong?

  // This gets called when the RSocketStateMachine is closed.  You cant use this
  // RSocketStateMachine anymore.

rsocket/internal/RSocketConnectionManager.cpp
phoad: I am using the clang+format, auto formatter. I have noticed this too but I wonder if we should use clang+format or I should just revert this line?

rsocket/internal/ManageableConnection.h
phoad: I thought we might have multiple listeners for the close event. Namely, now the connection manager listens for the close event, but it does not mean that we might also have one more listener for that event. So all of them should be able to listen that event, that's the reason of having it a SharedPromise.

Check the unit tests too, you will notice that in there we listen if the connection is closed, while the ConnectionManager is also listening for that.

I will push the Promise to the RSocketStateMachine and will make the statemachine to reset it if RESUME or SETUP is called.

alexmalyshev: Wait, you're right, `SETUP -> CLOSE -> RESUME -> CLOSE` can't happen.  I keep confusing `CLOSE` with `DISCONNECT`...

rsocket/internal/RSocketConnectionManager.cpp
alexmalyshev: We do use `clang-format` and it doesn't format things like this for any of us...  Is your IDE respecting the `.clang-format` file we have in our top-level directly?

alexmalyshev:
> in-band discussion with Fuat has convinced me that it makes sense to inherit from another interface for now

Sure.

> Design nit, but it also could make sense to have ManagedConnection be a just an interface...

Interface?  What are those?  Classes that are prefixed with `I`? 😛

rsocket/internal/RSocketConnectionManager.cpp
phoad: Ops, you are right. I was unaware of this file 💯

PR #731, opened 2017-09-07
fix undefined use-after-move ordering bug

dymk:
`connection.eventBase` might be evaluated _after_ the move of `connection` into the lambda.


dymk:
let's see if this manually triggers a travis build...

PR #733, opened 2017-09-07
RSocketConnectionManager: If client is already disconnected, no need to use evb_ to close the connection

phoad:
Users of our RSocketClient might only call `disconnect` method and they assume that its now fine to get rid of the corresponding eventBase that we used to add the StateMachine to the ConnectionManager.

Disconnect method already closes the connection. So we don't need to call the `close` method of the RSocketStateMachine within the `evb` if the connection is already disconnected.
This eliminates the possibility of trying to use the deleted `evb`.

So, either close() or disconnect() is called, it should be fine to have `evb`, out of the scope/deleted when the ConnectionManager is being destroyed.

### DIFF
--- a/rsocket/internal/ManageableConnection.h
+++ b/rsocket/internal/ManageableConnection.h
@@ -27,7 +27,7 @@ public:
   /// more functionality with `then` function.
   ///
   /// \return a Future to bind to.
-  folly::Future<folly::Unit> listenCloseEvent() {
+  virtual folly::Future<folly::Unit> listenCloseEvent() {
     return closePromise_.getFuture();
   }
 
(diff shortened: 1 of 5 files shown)


test/test_utils/MockManageableConnection.h
```diff
@@ -25,10 +25,20 @@ class MockManageableConnection : public rsocket::ManageableConnection {
   /// StreamAutomatonBase attached to this ConnectionAutomaton.
   void close(folly::exception_wrapper ex,
              StreamCompletionSignal scs) noexcept override {
+    closed_ = true;
     close_(ex, scs);
     onClose(ex);
   }
 
+  bool isDisconnectedOrClosed() const {
```
alexmalyshev: You want `override` here right?

rsocket/internal/ManageableConnection.h
```diff
@@ -38,6 +38,13 @@ class ManageableConnection {
   virtual void close(folly::exception_wrapper,
                      StreamCompletionSignal) = 0;
 
+  /// If the connection is already disconnected, then there is
+  /// no need to close it in the given eventbase, we can call
+  /// close() function inline.
```
alexmalyshev: Is that truly safe?  We don't depend on being on the right EventBase thread in the stream state machine methods that get called from `RSocketStateMachine::closeStreams()`?

somasun: This can be too risky to do.  We have a lot of callbacks in close() and the application might have some expectation on it.  I think a hard failure here is better (fail fast).  It is better to set the expectation that the evb which drives the RSocket should outlive the instances.

phoad:
Abandon

PR #734, opened 2017-09-08
Refactor and test FramedReader

alexmalyshev:
This is usually the first point of failure in crashes, and it's a nightmare to
understand.  Move out the functions that do things with the frame size and the
version (those can be shared with FramedWriter later too), add tests for error
cases.

### DIFF
--- a/rsocket/framing/FramedReader.cpp
+++ b/rsocket/framing/FramedReader.cpp
@@ -12,78 +12,70 @@ namespace rsocket {
 using namespace yarpl::flowable;
 
 namespace {
-constexpr auto kFrameLengthFieldLengthV0_1 = sizeof(int32_t);
-constexpr auto kFrameLengthFieldLengthV1_0 = 3; // bytes
-} // namespace
 
-size_t FramedReader::getFrameSizeFieldLength() const {
-  DCHECK(*protocolVersion_ != ProtocolVersion::Unknown);
-  if (*protocolVersion_ < FrameSerializerV1_0::Version) {
-    return kFrameLengthFieldLengthV0_1;
-  } else {
-    return kFrameLengthFieldLengthV1_0; // bytes
-  }
+constexpr size_t kFrameLengthFieldLengthV0_1 = sizeof(int32_t);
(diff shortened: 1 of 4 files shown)


rsocket/framing/FramedReader.cpp
```diff
@@ -12,214 +12,206 @@ namespace rsocket {
...
+    auto payloadSize = frameSizeWithoutLengthField(*version_, nextFrameSize);
+
+    DCHECK_GT(payloadSize, 0)
+        << "folly::IOBufQueue::split(0) returns a nullptr, can't have that";
+    auto nextFrame = payloadQueue_.split(payloadSize);
 
     CHECK(allowance_.tryAcquire(1));
 
-    VLOG(4) << "parsed frame length=" << nextFrame->length() << std::endl
+    VLOG(4) << "parsed frame length=" << nextFrame->length() << '\n'
```
lehecka: '\n' is not going to work on all platforms. On windows new line is encoded with 2 characters so the std:endl is internally inserting "\n\r"..

alexmalyshev: Are we even compiling on Windows?  Also `std::endl` is the worst because it's defined to flush the stream.  This isn't the end of the world when it's being used in a `VLOG`, but still...

rsocket/framing/FramedReader.cpp
```diff
@@ -12,214 +12,206 @@ namespace rsocket {
...
-  error("could not detect protocol version from framing");
+  error("Could not detect protocol version from framing");
   return false;
 }
 
 void FramedReader::error(std::string errorMsg) {
   VLOG(1) << "error: " << errorMsg;
-  onError(std::runtime_error(std::move(errorMsg)));
+
+  if (DuplexConnection::Subscriber::subscription()) {
```
lehecka: so this is really the only change in this PR, is that correct? The rest is only renaming and reformatting? And unit tests.

alexmalyshev: Should be yeah.  I'm looking to clean up some of the error/termination logic in `FramedReader`, and this was just the start of it.

PR #735, opened 2017-09-08
Resumption supported on multi-threaded server

somasun:
Major changes in this PR are 
- `FrameTransport` is an interface which `FrameTransportImpl` and `ScheduledFrameTransport` implement.  `FrameTransportImpl` is the old `FrameTransport` (just renaming here).
- `RSocketStateMachine` uses the `FrameTransport` interface to interact with the transport
- `ScheduledFrameProcessor` has been introduced.
- If the resumed connection is on a different EventBase, we use `ScheduledFrameTransport` and `ScheduledFrameProcessor` to ensure the `RSocketStateMachine` continues to live on the old EventBase, but communicate with the transport on the new EventBase.
- Else, there is no thread hopping and regular FrameTransport and FrameProcessor (RSocketStateMachine) is used.

In the future, if/when we get rid of FrameTransport, we have to just give a different implementation of the FrameTransport interface.  The current PR is not bound to FrameTransport in anyway. 

I tested by increasing the number of threads in the resumption-server (both unittest and examples) and verified that the resumption succeeds when resumed on different EventBase.  AsyncServerSocket round-robins among the available worker threads, so this was pretty deterministic.   I will follow up with stress testing resumption

### DIFF
--- a/rsocket/RSocketServer.cpp
+++ b/rsocket/RSocketServer.cpp
@@ -2,11 +2,12 @@
 
 #include "rsocket/RSocketServer.h"
 #include <folly/io/async/EventBaseManager.h>
+
 #include <rsocket/internal/ScheduledRSocketResponder.h>
 #include "rsocket/RSocketErrors.h"
 #include "rsocket/RSocketStats.h"
-#include "rsocket/framing/FrameTransport.h"
 #include "rsocket/framing/FramedDuplexConnection.h"
+#include "rsocket/framing/ScheduledFrameTransport.h"
 #include "rsocket/internal/RSocketConnectionManager.h"
 
 namespace rsocket {
@@ -119,8 +120,8 @@ void RSocketServer::onRSocketSetup(
     std::shared_ptr<RSocketServiceHandler> serviceHandler,
     yarpl::Reference<FrameTransport> frameTransport,
     SetupParameters setupParams) {
-  VLOG(1) << "Received new setup payload";
   auto* eventBase = folly::EventBaseManager::get()->getExistingEventBase();
+  VLOG(2) << "Received new setup payload on " << eventBase->getName();
   CHECK(eventBase);
   auto result = serviceHandler->onNewSetup(setupParams);
   if (result.hasError()) {
@@ -161,11 +162,32 @@ void RSocketServer::onRSocketResume(
   }
   auto serverState = std::move(result.value());
   CHECK(serverState);
+  auto* eventBase = folly::EventBaseManager::get()->getExistingEventBase();
+  VLOG(2) << "Resuming client on " << eventBase->getName();
   if (!serverState->eventBase_.isInEventBaseThread()) {
-    throw RSocketException("Trying to RESUME in a different worker thread");
+    // If the resumed connection is on a different EventBase, then use
+    // ScheduledFrameTransport and ScheduledFrameProcessor to ensure the
+    // RSocketStateMachine continues to live on the same EventBase and the
+    // IO happens in the new EventBase
+    auto scheduledFT = yarpl::make_ref<ScheduledFrameTransport>(
+        std::move(frameTransport),
+        eventBase, /* Transport EventBase */
+        &serverState->eventBase_); /* StateMachine EventBase */
+    serverState->eventBase_.runInEventBaseThread([
+      serverState,
+      scheduledFT = std::move(scheduledFT),
+      resumeParams = std::move(resumeParams)
+    ]() {
+      serverState->rSocketStateMachine_->resumeServer(
+          std::move(scheduledFT), resumeParams);
+    });
+  } else {
+    // If the resumed connection is on the same EventBase, then the
+    // RSocketStateMachine and Transport can continue living in the same
+    // EventBase without any thread hopping between them.
+    serverState->rSocketStateMachine_->resumeServer(
+        std::move(frameTransport), resumeParams);
   }
-  serverState->rSocketStateMachine_->resumeServer(
-      std::move(frameTransport), resumeParams);
 }
 
 void RSocketServer::startAndPark(
--- a/rsocket/framing/FrameTransport.h
+++ b/rsocket/framing/FrameTransport.h
@@ -1,68 +1,17 @@
-// Copyright 2004-present Facebook. All Rights Reserved.
-
 #pragma once
 
-#include <folly/ExceptionWrapper.h>
-#include "rsocket/DuplexConnection.h"
-#include "rsocket/internal/Common.h"
-#include "yarpl/flowable/Subscription.h"
+#include "yarpl/Refcounted.h"
+
+#include "rsocket/framing/FrameProcessor.h"
 
 namespace rsocket {
 
-class FrameProcessor;
-
-class FrameTransport final :
-    /// Registered as an input in the DuplexConnection.
-    public DuplexConnection::Subscriber,
-    /// Receives signals about connection writability.
-    public yarpl::flowable::Subscription {
+// Refer to FrameTransportImpl for documentation on the implementation
+class FrameTransport : public virtual yarpl::Refcounted {
  public:
-  explicit FrameTransport(std::unique_ptr<DuplexConnection> connection);
-  ~FrameTransport();
-
-  void setFrameProcessor(std::shared_ptr<FrameProcessor>);
-
-  /// Writes the frame directly to output. If the connection was closed it will
-  /// drop the frame.
-  void outputFrameOrDrop(std::unique_ptr<folly::IOBuf>);
-
-  /// Cancel the input, complete the output, and close the underlying
-  /// connection.
-  void close();
-
-  /// Cancel the input, error the output, and close the underlying connection.
-  /// This must be closed with a non-empty exception_wrapper.
-  void closeWithError(folly::exception_wrapper);
-
-  bool isClosed() const {
-    return !connection_;
-  }
-
- private:
-  void connect();
-
-  // Subscriber.
-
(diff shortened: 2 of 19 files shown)


rsocket/framing/ScheduledFrameProcessor.cpp
```diff
@@ -0,0 +1,26 @@
+#include "rsocket/framing/ScheduledFrameProcessor.h"
+
+#include <folly/ExceptionWrapper.h>
+#include <folly/io/IOBuf.h>
+
+namespace rsocket {
+
+void ScheduledFrameProcessor::processFrame(
+    std::unique_ptr<folly::IOBuf> ioBuf) {
+  // ScheduledFrameProcessor could get destroyed before the scheduled
```
lexs: You can get around this by submitting a copy of it only `~ScheduledFrameProcessor` gets called. Due to EventBases strict ordering you'll guarantee destruction doesn't happen until all callbacks have been processed.

This way we can avoid incrementing/decrementing the ref-count for every frame.

rsocket/framing/ScheduledFrameProcessor.cpp
```diff
@@ -0,0 +1,26 @@
...
+  // event gets executed.  Preserve a copy of inner frameProcessor_
+  evb_->runInEventBaseThread(
+      [ frameProcessor = frameProcessor_, ioBuf = std::move(ioBuf) ]() mutable {
+        frameProcessor->processFrame(std::move(ioBuf));
+      });
+}
+
+void ScheduledFrameProcessor::onTerminal(folly::exception_wrapper ex) {
+  // ScheduledFrameProcessor could get destroyed before the scheduled
+  // event gets executed.  Preserve a copy of inner frameProcessor_
```
lexs: ^ same

rsocket/framing/ScheduledFrameTransport.cpp
```diff
@@ -0,0 +1,43 @@
...
+    auto scheduledFP = std::make_shared<ScheduledFrameProcessor>(
+        std::move(fp), stateMachineEvb_);
+    frameTransport_->setFrameProcessor(std::move(scheduledFP));
+  });
+}
+
+void ScheduledFrameTransport::outputFrameOrDrop(
+    std::unique_ptr<folly::IOBuf> ioBuf) {
+  // ScheduledFrameTransport could get destroyed before the scheduled
+  // event gets executed.  Preserve a copy of inner frameTransport_
```
lexs: Same reasoning as other class.

rsocket/framing/ScheduledFrameProcessor.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include <folly/io/async/EventBase.h>
+
+#include "rsocket/framing/FrameProcessor.h"
+
+namespace rsocket {
+
+class ScheduledFrameProcessor : public FrameProcessor {
```
lexs: Can you add some doc what this solves? Other class too.

rsocket/RSocketServer.cpp
```diff
@@ -161,11 +162,32 @@ void RSocketServer::onRSocketResume(
...
   CHECK(serverState);
+  auto* eventBase = folly::EventBaseManager::get()->getExistingEventBase();
+  VLOG(2) << "Resuming client on " << eventBase->getName();
   if (!serverState->eventBase_.isInEventBaseThread()) {
-    throw RSocketException("Trying to RESUME in a different worker thread");
+    // If the resumed connection is on a different EventBase, then use
+    // ScheduledFrameTransport and ScheduledFrameProcessor to ensure the
+    // RSocketStateMachine continues to live on the same EventBase and the
+    // IO happens in the new EventBase
+    auto scheduledFT = yarpl::make_ref<ScheduledFrameTransport>(
```
phoad: Can we move the creation of scheduledFT into the lambda?

rsocket/framing/FrameTransport.h
```diff
@@ -1,68 +1,20 @@
...
-
-  void closeImpl(folly::exception_wrapper);
-
-  std::shared_ptr<FrameProcessor> frameProcessor_;
-  std::shared_ptr<DuplexConnection> connection_;
+namespace rsocket {
 
-  yarpl::Reference<DuplexConnection::Subscriber> connectionOutput_;
-  yarpl::Reference<yarpl::flowable::Subscription> connectionInputSub_;
+class FrameTransport : public virtual yarpl::Refcounted {
```
phoad: nit: do we need an interface to be ref counted?

rsocket/framing/ScheduledFrameProcessor.cpp
phoad: 💯

rsocket/framing/ScheduledFrameProcessor.h
```diff
@@ -0,0 +1,24 @@
...
+ public:
+  ScheduledFrameProcessor(
+      std::shared_ptr<FrameProcessor> fp,
```
phoad: Why not a unique_ptr? I would prefer the object that a class decorates to take the ownership of it.

rsocket/framing/FrameTransport.h
```diff
@@ -1,68 +1,20 @@
...
+#include <folly/io/async/EventBase.h>
 
-  bool isClosed() const {
-    return !connection_;
-  }
+#include "rsocket/framing/FrameProcessor.h"
 
- private:
-  void connect();
+using folly::EventBase;
```
alexmalyshev: Drive-by nit: This shouldn't be in the header, nor is it used.

rsocket/RSocketServer.cpp
somasun: No strong preference either way.  I found this easier to read.

rsocket/framing/FrameTransport.h
somasun: FrameTransport is passed around in rsocket lib as yarpl::ref

rsocket/framing/ScheduledFrameProcessor.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <folly/io/IOBuf.h>
+
+namespace rsocket {
+
+ScheduledFrameProcessor::~ScheduledFrameProcessor() {
+  // ScheduledFrameProcessor could get destroyed before the scheduled events of
+  // the class get executed.  So preserve a copy of inner frameProcessor_ in
+  // the event queue.  Since the events in the queue are executed in order we
+  // are guaranteed to have the frameProcessor_ during all the other events
+  // which were added before the destruction of this class.
```
alexmalyshev: :+1: for comment

rsocket/transports/tcp/TcpConnectionAcceptor.cpp
```diff
@@ -72,8 +73,10 @@ void TcpConnectionAcceptor::start(OnDuplexConnectionAccept onAccept) {
   callbacks_.reserve(options_.threads);
   for (size_t i = 0; i < options_.threads; ++i) {
     callbacks_.push_back(std::make_unique<SocketCallback>(onAccept_));
-    callbacks_[i]->eventBase()->runInEventBaseThread(
-        [] { folly::setThreadName("TcpConnectionAcceptor.Worker"); });
+    callbacks_[i]->eventBase()->runInEventBaseThread([i] {
+      folly::EventBaseManager::get()->getEventBase()->setName(
+          folly::sformat("TCPAcceptor.Worker.{}", i));
```
alexmalyshev: :+1:

rsocket/framing/ScheduledFrameTransport.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include <folly/io/async/EventBase.h>
+
+#include "rsocket/framing/FrameTransportImpl.h"
+#include "rsocket/framing/ScheduledFrameProcessor.h"
+
+using folly::EventBase;
```
alexmalyshev: Nit: We shouldn't use `using' in headers.

rsocket/framing/FrameTransport.h
```diff
@@ -1,68 +1,18 @@
...
-
-  void closeImpl(folly::exception_wrapper);
-
-  std::shared_ptr<FrameProcessor> frameProcessor_;
-  std::shared_ptr<DuplexConnection> connection_;
+namespace rsocket {
 
-  yarpl::Reference<DuplexConnection::Subscriber> connectionOutput_;
-  yarpl::Reference<yarpl::flowable::Subscription> connectionInputSub_;
+class FrameTransport : public virtual yarpl::Refcounted {
```
alexmalyshev: Nit: Drop a comment here saying to read `FrameTransportImpl`'s comments 😛 .

rsocket/framing/FrameTransport.h
```diff
@@ -1,68 +1,18 @@
...
-
-  /// Cancel the input, complete the output, and close the underlying
-  /// connection.
-  void close();
+#include "yarpl/Refcounted.h"
 
-  /// Cancel the input, error the output, and close the underlying connection.
-  /// This must be closed with a non-empty exception_wrapper.
-  void closeWithError(folly::exception_wrapper);
+#include <folly/io/async/EventBase.h>
```
alexmalyshev: We're not using `EventBase` in this header, can drop this.

alexmalyshev:
> Why aren't there any changes to WarmResumptionTest? IIRC WarmResumptionTest.FailedResumption1 runs a server with 2 threads, and assumes the first resumption attempt will fail. That should no longer be the case right?

Talked IRL, but it's cause the server in those tests isn't resumable.  Ignore me.

somasun:
We will use SwappableEventBase or better approaches for the longer term.  This is to get things running in the short term.

PR #736, opened 2017-09-08
Cleaner termination semantics for FramedReader

alexmalyshev:
FramedReader is a Subscriber, so we should expect
FramedReader::on{Complete,Error}() to be called at most once, and after a call
to FramedReader::onSubscribe().

As such, we shouldn't call them from FramedReader::error(), that should have its
own logic to cancel the subscription safely and to error the inner subscriber
safely, while still being resilient to running
FramedReader::on{Complete,Error}() afterwards.  Added tests for this.

Connecting to RSocket server with `nc` and typing garbage still works properly.

### DIFF
--- a/rsocket/framing/FramedReader.cpp
+++ b/rsocket/framing/FramedReader.cpp
@@ -131,9 +131,7 @@ void FramedReader::parseFrames() {
 
 void FramedReader::onComplete() {
   payloadQueue_.move();
-  if (DuplexConnection::Subscriber::subscription()) {
-    DuplexConnection::Subscriber::onComplete();
-  }
+  DuplexConnection::Subscriber::onComplete();
   if (auto subscriber = std::move(inner_)) {
     // After this call the instance can be destroyed!
     subscriber->onComplete();
@@ -142,9 +140,7 @@ void FramedReader::onComplete() {
 
 void FramedReader::onError(folly::exception_wrapper ex) {
   payloadQueue_.move();
-  if (DuplexConnection::Subscriber::subscription()) {
-    DuplexConnection::Subscriber::onError({});
-  }
+  DuplexConnection::Subscriber::onError({});
   if (auto subscriber = std::move(inner_)) {
     // After this call the instance can be destroyed!
     subscriber->onError(std::move(ex));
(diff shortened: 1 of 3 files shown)


rsocket/framing/FramedReader.cpp
```diff
@@ -131,9 +131,7 @@ void FramedReader::parseFrames() {
 
 void FramedReader::onComplete() {
   payloadQueue_.move();
-  if (DuplexConnection::Subscriber::subscription()) {
```
phoad: I believe we fixed a bug with this check.
Subscriber depends on having a Subscription otherwise it crashes.

Would you mind running a server and trying to connect this server with a non-rsocket client and observe if the server is crashing?

test/framing/FramedReaderTest.cpp
```diff
@@ -53,3 +53,49 @@ TEST(FramedReader, CantDetectVersion) {
...
+
+TEST(FramedReader, SubscriberErrorAfterError) {
+  auto version = std::make_shared<ProtocolVersion>(ProtocolVersion::Latest);
+  auto reader = yarpl::make_ref<FramedReader>(version);
+
+  auto subscription = yarpl::make_ref<StrictMock<MockSubscription>>();
+  EXPECT_CALL(*subscription, request_(_));
+  EXPECT_CALL(*subscription, cancel_());
+
+  reader->onSubscribe(subscription);
```
phoad: Make a test and assume that Reader doesn't have any subscription.

rsocket/framing/FramedReader.cpp
lehecka: I agree with Alex, that if somebody is calling onComplete, they have called onSubscribe and we should have subscription. Maybe we should turn it into DCHECK(DuplexConnection::Subscriber::subscription()) ?
Let's find the problem Fuat is describing, maybe the fix he is talking about should have been in another place, not here.

alexmalyshev: @phoad It's in the summary

> Connecting to RSocket server with nc and typing garbage still works properly.

@lehecka That exact DCHECK is already done by `DuplexConnection::Subscriber::onComplete()`

test/framing/FramedReaderTest.cpp
alexmalyshev: That's not valid then, as we'd call onError() without calling onSubscribe().

I understand that's what the bug you fixed was, but I'm pretty sure we were hitting it only because we had called `FramedReader::error()` earlier.

PR #737, opened 2017-09-08
onComplete() should not be called after sync call to cancel()

dymk:
and fixes a use after free when FlowableOperator is terminated

### DIFF
--- a/yarpl/include/yarpl/flowable/FlowableOperator.h
+++ b/yarpl/include/yarpl/flowable/FlowableOperator.h
@@ -127,18 +127,17 @@ class FlowableOperator : public Flowable<D> {
     void terminateImpl(
         TerminateState state,
         folly::exception_wrapper ex = folly::exception_wrapper{nullptr}) {
-      if (isTerminated()) {
-        return;
-      }
+      auto self = this->ref_from_this(this);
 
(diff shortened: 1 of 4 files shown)


yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -127,18 +127,17 @@ class FlowableOperator : public Flowable<D> {
...
         TerminateState state,
         folly::exception_wrapper ex = folly::exception_wrapper{nullptr}) {
-      if (isTerminated()) {
-        return;
-      }
+      auto self = this->ref_from_this(this);
 
-      if (auto upstream = std::move(upstream_)) {
-        if (state.up) {
+      if (state.up) {
```
alexmalyshev: Nit: It might be worth it to drop a comment as to why these ifs are ordered the way they are.  It took me a minute to understand it.

dymk: Sure.

In the 'terminate up' case, we drop references to both upstream and downstream.
In the 'terminate down' case, we only drop references to the downstream (upstream can still have cancel called on it). 

It might make sense to still just drop both? At the moment this works, though.

PR #742, opened 2017-09-12
Add fire-and-forget benchmark

alexmalyshev:
Pretty similar to the request-response one, except this one puts the latch on
the server's responder.

Adds `(void)n;` statements to benchmarks so the compiler shuts up about unused
variables.  Note: We can remove the `n` argument entirely from the macro, but
that'll just make the compiler cry about macros being called with the wrong
number of arguments.

### DIFF
--- a/benchmarks/StreamThroughputMemory.cpp
+++ b/benchmarks/StreamThroughputMemory.cpp
@@ -150,6 +150,8 @@ std::shared_ptr<RSocketClient> makeClient() {
 }
 
 BENCHMARK(StreamThroughput, n) {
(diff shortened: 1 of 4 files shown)


benchmarks/StreamThroughputMemory.cpp
```diff
@@ -150,6 +150,8 @@ std::shared_ptr<RSocketClient> makeClient() {
 }
 
 BENCHMARK(StreamThroughput, n) {
+  (void)n;
```
dymk: <3

PR #743, opened 2017-09-12
Request stream tests

ragansa:
Adds some tests to requestStream

### DIFF
--- a/test/RSocketClientServerTest.cpp
+++ b/test/RSocketClientServerTest.cpp
@@ -87,7 +87,7 @@ TEST(RSocketClientServer, ServerGetsGarbage) {
   auto server = makeServer(std::make_shared<HelloStreamRequestHandler>());
   folly::SocketAddress address{"::1", *server->listeningPort()};
 
-  folly::ScopedEventBaseThread worker{"rsocket-client"};
+  folly::ScopedEventBaseThread worker;
   auto factory =
       std::make_shared<TcpConnectionFactory>(*worker.getEventBase(), address);
 
--- a/test/RequestStreamTest.cpp
+++ b/test/RequestStreamTest.cpp
@@ -152,7 +152,7 @@ class TestHandlerError : public rsocket::RSocketResponder {
  public:
   Reference<Flowable<Payload>> handleRequestStream(Payload, StreamId) override {
     return Flowables::error<Payload>(
-        std::runtime_error("A wild error appeared"));
+        std::runtime_error("A wild Error appeared!"));
   }
 };
 
@@ -166,7 +166,7 @@ TEST(RequestStreamTest, HandleError) {
       ->map([](auto p) { return p.moveDataToString(); })
       ->subscribe(ts);
   ts->awaitTerminalEvent();
-  ts->assertOnErrorMessage("A wild error appeared");
+  ts->assertOnErrorMessage("A wild Error appeared!");
 }
 
 class TestErrorThrown : public rsocket::RSocketResponder {
@@ -175,20 +175,22 @@ class TestErrorThrown : public rsocket::RSocketResponder {
       override {
     // string from payload data
     auto requestString = request.moveDataToString();
-      return Flowables::range(1, 10)->map([name = std::move(requestString)](
-          int64_t v) {
-        if (v == 5) {
-          throw std::runtime_error{"What an unexpected turn of events!"};
-        }
-        std::stringstream ss;
-        ss << "Hello " << name << " " << v << "!";
-        std::string s = ss.str();
-        return Payload(s, "metadata");
-      });
+
+    return Flowable<Payload>::create([name = std::move(requestString)](Reference<Subscriber<Payload>> subscriber,
+                                 int64_t requested) {
+          EXPECT_GT(requested, 1);
+          subscriber->onNext(Payload(name, "meta"));
+      subscriber->onNext(Payload(name, "meta"));
+      subscriber->onNext(Payload(name, "meta"));
+      subscriber->onNext(Payload(name, "meta"));
+      subscriber->onError(std::runtime_error("A wild Error appeared!"));
+          return std::make_tuple(int64_t(1), true);
+        });
+
   }
 };
 
-TEST(RequestStreamTest, DISABLED_HandleErrorMidStream) {
+TEST(RequestStreamTest, HandleErrorMidStream) {
   folly::ScopedEventBaseThread worker;
   auto server = makeServer(std::make_shared<TestErrorThrown>());
   auto client = makeClient(worker.getEventBase(), *server->listeningPort());
@@ -197,5 +199,8 @@ TEST(RequestStreamTest, DISABLED_HandleErrorMidStream) {
     requester->requestStream(Payload("Bob"))
              ->map([](auto p) { return p.moveDataToString(); })
              ->subscribe(ts);
-    ts->awaitTerminalEvent(); // crashes, do we want to test that something else occurs?
+    ts->awaitTerminalEvent();
+  ts->assertValueCount(4);
+  ts->assertOnErrorMessage("A wild Error appeared!");
+
 }
--- a/test/RequestStreamTest.cpp
+++ b/test/RequestStreamTest.cpp
@@ -176,17 +176,16 @@ class TestErrorThrown : public rsocket::RSocketResponder {
     // string from payload data
     auto requestString = request.moveDataToString();
 
-    return Flowable<Payload>::create([name = std::move(requestString)](Reference<Subscriber<Payload>> subscriber,
-                                 int64_t requested) {
-          EXPECT_GT(requested, 1);
-          subscriber->onNext(Payload(name, "meta"));
+    return Flowable<Payload>::create([name = std::move(requestString)](
+        Reference<Subscriber<Payload>> subscriber, int64_t requested) {
+      EXPECT_GT(requested, 1);
+      subscriber->onNext(Payload(name, "meta"));
       subscriber->onNext(Payload(name, "meta"));
       subscriber->onNext(Payload(name, "meta"));
       subscriber->onNext(Payload(name, "meta"));
       subscriber->onError(std::runtime_error("A wild Error appeared!"));
-          return std::make_tuple(int64_t(1), true);
-        });
-
+      return std::make_tuple(int64_t(1), true);
+    });
   }
 };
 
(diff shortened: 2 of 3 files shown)


test/RSocketClientServerTest.cpp
```diff
@@ -87,7 +87,7 @@ TEST(RSocketClientServer, ServerGetsGarbage) {
   auto server = makeServer(std::make_shared<HelloStreamRequestHandler>());
   folly::SocketAddress address{"::1", *server->listeningPort()};
 
-  folly::ScopedEventBaseThread worker{"rsocket-client"};
+  folly::ScopedEventBaseThread worker;
```
ragansa: This was crashing the build until I removed it?  Was this causing anyone else issues?

test/RequestStreamTest.cpp
```diff
@@ -120,3 +145,29 @@ TEST(RequestStreamTest, RequestOnDisconnectedClient) {
   CHECK_WAIT(wait_for_on_error);
   ASSERT(did_call_on_error);
 }
+
+class TestHandlerError : public rsocket::RSocketResponder {
+ public:
+  Reference<Flowable<Payload>> handleRequestStream(
```
dymk:
```
handleRequestStream(Payload, StreamId) {
```

test/RequestStreamTest.cpp
```diff
@@ -120,3 +145,29 @@ TEST(RequestStreamTest, RequestOnDisconnectedClient) {
...
+      Payload request,
+      StreamId) override {
+    // string from payload data
+    auto requestString = request.moveDataToString();
```
dymk: can remove this

test/RequestStreamTest.cpp
```diff
@@ -120,3 +145,29 @@ TEST(RequestStreamTest, RequestOnDisconnectedClient) {
...
+  folly::ScopedEventBaseThread worker;
+  auto server = makeServer(std::make_shared<TestHandlerError>());
+  auto client = makeClient(worker.getEventBase(), *server->listeningPort());
+  auto requester = client->getRequester();
+  auto ts = TestSubscriber<std::string>::create();
+  requester->requestStream(Payload("Bob"))
+           ->map([](auto p) { return p.moveDataToString(); })
+           ->subscribe(ts);
+  ts->awaitTerminalEvent();
+  ts->assertOnErrorMessage("A wild error appeared");
```
dymk: Add a test for what happens when we receive {0, 1, many} elements and then an error
Also test what happens when an exception is thrown in:
 - thrown in handleRequestStream
 - return a Flowable::error
 -  return a Flowable::create which emits elements, and within the Flowable throws 

and assertValueCount for those cases?

yarpl/include/yarpl/flowable/TestSubscriber.h
```diff
@@ -178,7 +178,7 @@ class TestSubscriber : public Subscriber<T> {
   void assertOnErrorMessage(std::string msg) {
     if (!e_ || e_.get_exception()->what() != msg) {
       std::stringstream ss;
-      ss << "Error is: " << e_ << " but expected: " << msg;
+      ss << "Error is: '" << e_ << "' but expected: '" << msg << "'";
```
dymk: (y)

test/RequestStreamTest.cpp
```diff
@@ -49,6 +49,31 @@ TEST(RequestStreamTest, HelloSync) {
...
+  folly::ScopedEventBaseThread worker;
+  auto server = makeServer(std::make_shared<TestHandlerSync>());
+  auto client = makeClient(worker.getEventBase(), *server->listeningPort());
+  auto requester = client->getRequester();
+  auto ts = TestSubscriber<std::string>::create(5);
+  requester->requestStream(Payload("Bob"))
+      ->map([](auto p) { return p.moveDataToString(); })
+      ->subscribe(ts);
+
+  // cause data to flow?  I can't use awaitTerminalEvent, it sits forever
```
dymk: what's 'cause data to flow' refer to?

test/RequestStreamTest.cpp
```diff
@@ -120,3 +145,29 @@ TEST(RequestStreamTest, RequestOnDisconnectedClient) {
...
+class TestHandlerError : public rsocket::RSocketResponder {
+ public:
+  Reference<Flowable<Payload>> handleRequestStream(
+      Payload request,
+      StreamId) override {
+    // string from payload data
+    auto requestString = request.moveDataToString();
+
+    return Flowables::error<Payload>(folly::exception_wrapper(
+        std::runtime_error("A wild error appeared")));
```
dymk: doesn't need to be wrapped explicitly in a folly::exception_wrapper

test/RequestStreamTest.cpp
```diff
@@ -120,3 +145,29 @@ TEST(RequestStreamTest, RequestOnDisconnectedClient) {
...
+};
+
+TEST(RequestStreamTest, HandleError) {
+  folly::ScopedEventBaseThread worker;
+  auto server = makeServer(std::make_shared<TestHandlerError>());
+  auto client = makeClient(worker.getEventBase(), *server->listeningPort());
+  auto requester = client->getRequester();
+  auto ts = TestSubscriber<std::string>::create();
+  requester->requestStream(Payload("Bob"))
+           ->map([](auto p) { return p.moveDataToString(); })
```
dymk: I wish our TestSubscriber consumed Payloads rather than strings :/ we end up testing MapOperator all the time in addition to the request/response path. Could be worthwhile to do in a later change.

test/RequestStreamTest.cpp
```diff
@@ -120,3 +145,29 @@ TEST(RequestStreamTest, RequestOnDisconnectedClient) {
...
+           ->subscribe(ts);
+  ts->awaitTerminalEvent();
+  ts->assertOnErrorMessage("A wild error appeared");
+}
```
dymk: Looks like clang-format didn't add a newline at end of file?

test/RequestStreamTest.cpp
ragansa: When line 65 executes, nothing has returned yet.  But I don't want to use the testSubscriber's awaitTerminalEvent(), since it will block forever (since I'm not requesting the full 10 all at once and no onComplete will be called)

dymk: Makes sense. We probably want a method on TestSubscriber that says "awaitUntilNumElements(int n)" and have that block until 1) a timeout of some hundreds of ms expires (and throw), 2) onComplete/onError (and throw), 3) onNext is called 'n' times (and return normally).

test/RequestStreamTest.cpp
ragansa: * thrown in handleRequestStream
_Unless caught, and something (Flowables::error probably) is sent, what would we want to happen?  An uncaught throw should crash no?_
* return a Flowable::error
_Isn't this what testHandlerError is doing?_
* return a Flowable::create which emits elements, and within the Flowable throws
_Same story as above, if we're catching it, we're just sending a Flowables::error, and if not, it should crash no?_

I guess I could kind of see doing this and confirming that the correct number of elements made it through before the crash?  That's not really testing any path that we're not seeing in the normal HelloSync test.  I'm not seeing how this adds to the test cases.  Am I misunderstanding what you're suggesting?

ragansa: Threw together an example that throws after a few go out, and it just crashes the whole thing.  

```
class TestErrorThrown : public rsocket::RSocketResponder {
 public:
  Reference<Flowable<Payload>> handleRequestStream(Payload request, StreamId)
      override {
    // string from payload data
    auto requestString = request.moveDataToString();
      return Flowables::range(1, 10)->map([name = std::move(requestString)](
          int64_t v) {
        if (v == 5) {
          throw std::runtime_error{"What an unexpected turn of events!"};
        }
        std::stringstream ss;
        ss << "Hello " << name << " " << v << "!";
        std::string s = ss.str();
        return Payload(s, "metadata");
      });
  }
};
```

Do we expect better behavior?

ragansa:
Ack, I got another commit into my PR somehow, doesn't look like it'd be a problem?

test/RequestStreamTest.cpp
```diff
@@ -120,3 +147,58 @@ TEST(RequestStreamTest, RequestOnDisconnectedClient) {
...
+  auto requester = client->getRequester();
+  auto ts = TestSubscriber<std::string>::create();
+  requester->requestStream(Payload("Bob"))
+      ->map([](auto p) { return p.moveDataToString(); })
+      ->subscribe(ts);
+  ts->awaitTerminalEvent();
+  ts->assertOnErrorMessage("A wild Error appeared!");
+}
+
+class TestErrorThrown : public rsocket::RSocketResponder {
```
dymk: (don't worry about changing) naming nit: TestErrorAfterOnNextResponder

yarpl/include/yarpl/flowable/TestSubscriber.h
```diff
@@ -111,6 +112,24 @@ class TestSubscriber : public Subscriber<T> {
     terminalEventCV_.wait(lk, [this] { return terminated_; });
   }
 
+  void awaitValueCount(
+      int64_t n,
+      std::chrono::milliseconds ms = std::chrono::seconds{5}) {
+    // now block this thread
+    std::unique_lock<std::mutex> lk(m_);
+    if (!terminalEventCV_.wait_for(lk, ms, [this, n] {
```
dymk: lambda here is big enough to warrant: 
```cpp
auto didTimeOut = terminalEventCV_.wait_for(...);
if (!didTimeOut) { ...
```

test/RequestStreamTest.cpp
```diff
@@ -72,11 +98,12 @@ class TestHandlerAsync : public rsocket::RSocketResponder {
               return Payload(s, "metadata");
             })
             ->subscribe(subscriber);
-      }).detach();
+      })
+          .detach();
```
dymk: is this clang-format doing something weird?

test/RequestStreamTest.cpp
```diff
@@ -120,3 +147,58 @@ TEST(RequestStreamTest, RequestOnDisconnectedClient) {
   CHECK_WAIT(wait_for_on_error);
   ASSERT(did_call_on_error);
 }
+
+class TestHandlerError : public rsocket::RSocketResponder {
```
dymk: (don't worry about changing) naming nit: TestErrorResponder

test/RequestStreamTest.cpp
```diff
@@ -120,3 +147,58 @@ TEST(RequestStreamTest, RequestOnDisconnectedClient) {
...
+
+    return Flowable<Payload>::create([name = std::move(requestString)](
+        Reference<Subscriber<Payload>> subscriber, int64_t requested) {
+      EXPECT_GT(requested, 1);
+      subscriber->onNext(Payload(name, "meta"));
+      subscriber->onNext(Payload(name, "meta"));
+      subscriber->onNext(Payload(name, "meta"));
+      subscriber->onNext(Payload(name, "meta"));
+      subscriber->onError(std::runtime_error("A wild Error appeared!"));
+      return std::make_tuple(int64_t(1), true);
```
dymk: std::make_tuple(int64_t(4), true);

test/RequestStreamTest.cpp
ragansa: must've been, looks weird to me, I'll try to put it back

PR #744, opened 2017-09-12
TCPConnectionAcceptor should not destroy its thread on the `stop` function

phoad:
TcpConnectionAcceptor was destroying the corresponding thread when it is stopped. 
But, when I took a look at the RSocketServer::shutdownAndWait() function, I see that `connectionManager` is reset later then stopping the acceptors. 
ConnectionManager uses the event bases of the threads in the Acceptors. 
So if we get rid of these threads, then the corresponding event bases are deleted before it can be used by the ConnectionManager.

So we should not destroy the thread at the TCPConnectionAcceptor::stop() function but at the destructor of the TCPConnectionAcceptor.

I have verified that the HttpConnectionAcceptor's implementation is already correct.

+ Unit tests for creating connections with existing sockets.

### DIFF
--- a/rsocket/transports/tcp/TcpConnectionAcceptor.cpp
+++ b/rsocket/transports/tcp/TcpConnectionAcceptor.cpp
@@ -26,8 +26,7 @@ class TcpConnectionAcceptor::SocketCallback
     folly::AsyncSocket::UniquePtr socket(
         new folly::AsyncSocket(eventBase(), fd));
 
-    auto connection = std::make_unique<TcpDuplexConnection>(
-        std::move(socket));
+    auto connection = std::make_unique<TcpDuplexConnection>(std::move(socket));
     onAccept_(std::move(connection), *eventBase());
   }
 
@@ -55,6 +54,7 @@ TcpConnectionAcceptor::TcpConnectionAcceptor(Options options)
 TcpConnectionAcceptor::~TcpConnectionAcceptor() {
   if (serverThread_) {
     stop();
+    serverThread_.reset();
   }
 }
 
@@ -111,8 +111,7 @@ void TcpConnectionAcceptor::stop() {
   VLOG(1) << "Shutting down TCP listener";
 
   serverThread_->getEventBase()->runInEventBaseThread(
-      [this] { serverSocket_.reset(); });
-  serverThread_.reset();
+      [serverSocket = std::move(serverSocket_)]() {});
 }
 
 folly::Optional<uint16_t> TcpConnectionAcceptor::listeningPort() const {
@@ -122,4 +121,4 @@ folly::Optional<uint16_t> TcpConnectionAcceptor::listeningPort() const {
   return serverSocket_->getAddress().getPort();
 }
 
-}
+} // namespace rsocket


test/RSocketClientServerTest.cpp
```diff
@@ -103,3 +108,139 @@ TEST(RSocketClientServer, ServerGetsGarbage) {
...
+      int fd = callback->getConnectedSocket().get();
+      LOG(INFO) << "fd: " << fd;
+
+      connectedClientEvb->runInEventBaseThreadAndWait([&server,
+                                                          &connectedClientEvb,
+                                                          fd]() {
+        folly::AsyncSocket::UniquePtr socket(
+            new folly::AsyncSocket(connectedClientEvb, fd));
+
+        auto connection = std::make_unique<TcpDuplexConnection>(
```
phoad: I need to do some cleanup in these lines.

rsocket/RSocketClient.cpp
```diff
@@ -47,6 +47,12 @@ const std::shared_ptr<RSocketRequester>& RSocketClient::getRequester() const {
 folly::Future<folly::Unit> RSocketClient::resume() {
   VLOG(2) << "Resuming connection";
 
+  if (disconnectBaton_) {
```
lehecka: this is not thread safe (and all the other places where we use the Baton). Why do we need it anyways?

rsocket/RSocketClient.cpp
```diff
@@ -118,15 +124,61 @@ folly::Future<folly::Unit> RSocketClient::resume() {
...
+  }
+
+  if (disconnectBaton_) {
+    disconnectBaton_->wait();
+    disconnectBaton_.clear();
+  }
+
+  disconnectBaton_.emplace();
+
+  auto work = [ sm = stateMachine_, e = std::move(ew), this ]() mutable {
```
lehecka: you can't pass this inside of the lambda, Dylan just fixed an issue where we were accessing "this" which was already gone.

rsocket/RSocketClient.cpp
phoad: I was just about to remove the baton's. It was for temporarily testing.

phoad:
Yes, I have eliminated the Baton.

rsocket/transports/tcp/TcpConnectionAcceptor.cpp
```diff
@@ -111,8 +111,7 @@ void TcpConnectionAcceptor::stop() {
   VLOG(1) << "Shutting down TCP listener";
 
   serverThread_->getEventBase()->runInEventBaseThread(
-      [this] { serverSocket_.reset(); });
-  serverThread_.reset();
+      [serverSocket = std::move(serverSocket_)]() mutable {});
```
alexmalyshev: Nit: This doesn't need `mutable`

test/RSocketClientServerTest.cpp
```diff
@@ -103,3 +108,132 @@ TEST(RSocketClientServer, ServerGetsGarbage) {
...
+TEST_P(RSocketClientServerRaw, ServerWithNoAcceptor) {
+  // ConnectionAcceptor's should destroy the threads the last!
+  auto connectedClientThread = std::make_unique<folly::ScopedEventBaseThread>();
+  {
+    // Make a server with no Acceptor as we will already provide the connection!
+    auto server = makeServerWithNoAcceptor();
+
+    folly::ScopedEventBaseThread serverThread;
+    auto serverSocket =
+        new folly::AsyncServerSocket(serverThread.getEventBase());
```
alexmalyshev: Where does this get destroyed?  It's a raw pointer.

test/RSocketClientServerTest.cpp
```diff
@@ -103,3 +108,132 @@ TEST(RSocketClientServer, ServerGetsGarbage) {
...
+    auto callback = std::make_unique<SocketCallback>();
+    folly::EventBase* connectedClientEvb =
```
alexmalyshev: Nit: `auto`

test/RSocketClientServerTest.cpp
```diff
@@ -103,3 +108,132 @@ TEST(RSocketClientServer, ServerGetsGarbage) {
...
+         &serverPortPromise,
+         connectedClientEvb] {
+          serverSocket->bind(folly::SocketAddress("::", 0));
+          serverSocket->addAcceptCallback(callback.get(), connectedClientEvb);
+          serverSocket->listen(/*backlog=*/1);
+          serverSocket->startAccepting();
+          for (auto& i : serverSocket->getAddresses()) {
+            VLOG(1) << "Listening on " << i.describe();
+            serverPortPromise.setValue(i.getPort());
+          }
```
alexmalyshev: How about you return `i.getPort()` here, instead of using `serverPortPromise`?

test/RSocketClientServerTest.cpp
```diff
@@ -103,3 +108,132 @@ TEST(RSocketClientServer, ServerGetsGarbage) {
...
+        })
+        .get();
+
+    int serverPort = serverPortPromise.getFuture().get();
+
+    // Thread should outlive the client.
+    auto clientThread = std::make_unique<folly::ScopedEventBaseThread>();
+    {
+      // Client connects to the server.
+      std::unique_ptr<RSocketClient> client =
```
alexmalyshev: Nit: `auto`

test/RSocketClientServerTest.cpp
```diff
@@ -103,3 +108,132 @@ TEST(RSocketClientServer, ServerGetsGarbage) {
...
+          serverSocket->listen(/*backlog=*/1);
+          serverSocket->startAccepting();
+          for (auto& i : serverSocket->getAddresses()) {
+            VLOG(1) << "Listening on " << i.describe();
+            serverPortPromise.setValue(i.getPort());
+          }
+        })
+        .get();
+
+    int serverPort = serverPortPromise.getFuture().get();
```
alexmalyshev: Nit: `auto`.

phoad:
@alexmalyshev - I got your message of only keeping the update at the TcpConnectionAcceptor and moving the rest to another PR. So here it is ;)

alexmalyshev:
@phoad Hah, that didn't mean that I *didn't* like the test changes.

PR #745, opened 2017-09-13
`observeOn`, `subscribeOn` which take a folly::Executor&

dymk:
And associated tests to verify the subscriber/producer is executing on the correct thread.


yarpl/include/yarpl/flowable/FlowableObserveOnOperator.h
```diff
@@ -0,0 +1,65 @@
...
+ public:
+  ObserveOnOperatorSubscriber(
+      Reference<Subscriber<T>> inner,
+      folly::Executor& executor)
+      : inner_(std::move(inner)), executor_(executor) {}
+
+  // called from upstream EB
+  void onSubscribe(Reference<Subscription> s) override {
+    executor_.add(
+        [ this, self = this->ref_from_this(this), s = std::move(s) ]{
```
alexmalyshev: Why not:

`[inner = inner_, s = std::move(s)]() mutable { inner->onSubscribe(std::move(s)); }`

And for all followup ones too.

yarpl/include/yarpl/flowable/FlowableObserveOnOperator.h
```diff
@@ -0,0 +1,65 @@
...
+  }
+  void onComplete() override {
+    executor_.add([this, self = this->ref_from_this(this)] {
+        inner_->onComplete();
+    });
+  }
+  void onError(folly::exception_wrapper err) override {
+    executor_.add(
+        [ this, self = this->ref_from_this(this), err = std::move(err) ] {
+            inner_->onError(err);
```
alexmalyshev: Should do `std::move(err)`, and make the lambda `mutable`.  I like naming the two variables differently, so I know if I'm accidentally making a copy where I meant to make a move (i.e. `[e = std::move(err)]() mutable { /* use e */ }`).

yarpl/include/yarpl/flowable/TestSubscriber.h
```diff
@@ -104,11 +104,11 @@ class TestSubscriber : public Subscriber<T> {
   /**
    * Block the current thread until either onSuccess or onError is called.
    */
-  void awaitTerminalEvent() {
+  void awaitTerminalEvent(std::chrono::milliseconds ms = std::chrono::milliseconds(5000)) {
```
alexmalyshev: `ms = std::chrono::seconds{5}`

yarpl/test/yarpl/test_utils/utils.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
```
alexmalyshev: `yarpl/test/yarpl/test_utils/utils.h`

Really?  We repeat EVERY word twice?

yarpl/include/yarpl/flowable/FlowableObserveOnOperator.h
dymk: inner_ goes away if all reference to `this` are lost

yarpl/test/yarpl/test_utils/utils.h
dymk: that's a suggestion to rename to `yarpl_test_utils.h`, yes??

alexmalyshev: It was more just me griping about the directory structure in general.  It's fine to leave this file as is, I don't expect `yarpl/test/yarpl` to be restructured in this diff.

yarpl/include/yarpl/flowable/FlowableObserveOnOperator.h
```diff
@@ -0,0 +1,105 @@
...
+      Reference<ObserveOnOperatorSubscriber<T>> subscriber,
+      Reference<Subscription> subscription)
+      : subscriber_(std::move(subscriber)),
+        subscription_(std::move(subscription)) {}
+
+  // all requesting methods are called from 'executor_' in the
+  // associated subscriber
+  void cancel() override {
+    if (auto subscriber = std::move(subscriber_)) {
+      subscriber_->isCanceled_ = true;
```
alexmalyshev: You mean `subscriber`?  `subscriber_` has been moved away.

Should we keep the `subscriber` reference till the end of the function, just in case `subscription_->cancel()` blows away the last reference to `this`?

dymk: added a test to verify that this case is covered, good catch!
I'll keep a 'self' around, in case that happens. The call to subscription_->cancel() might also blow our last reference away, in a way that makes accessing subscriber_ unsafe as well.

PR #749, opened 2017-09-15
Make ~RSocketConnectionManager asynchronous, rename to ConnectionSet

alexmalyshev:
RSocketStateMachine now has a weak pointer to the ConnectionSet that holds it.
When the state machine is closed, it'll attempt to remove itself from the set.
When the set is about to be destroyed, it schedules RSocketStateMachine::close()
on all of its state machines, and exits immediately.


rsocket/RSocketServer.cpp
```diff
@@ -143,7 +142,10 @@ void RSocketServer::onRSocketSetup(
       std::move(connectionParams.connectionEvents),
       nullptr, /* resumeManager */
       nullptr /* coldResumeHandler */);
-  connectionManager_->manageConnection(rs, *eventBase);
+
+  connectionSet_->insert(rs, eventBase);
+  rs->registerSet(connectionSet_);
```
alexmalyshev: I could fold all these calls to `registerSet()` into `ConnectionSet::insert()`, but that'd mean `ConnectionSet` would have to inherit from `enable_shared_from_this`, something which I don't really feel like doing....

rsocket/RSocketClient.h
```diff
@@ -72,7 +72,7 @@ class RSocketClient {
   void createState(folly::EventBase& eventBase);
 
   std::shared_ptr<ConnectionFactory> connectionFactory_;
-  std::unique_ptr<RSocketConnectionManager> connectionManager_;
+  std::shared_ptr<ConnectionSet> connectionSet_;
```
phoad: The forward declaration of RSocketConnectionManager is still there.

phoad:
This has resolved a bug in our codebase already: T21893729
💯 👍

rsocket/internal/ConnectionSet.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+ConnectionSet::ConnectionSet() {}
+
+ConnectionSet::~ConnectionSet() {
+  VLOG(1) << "Started ~ConnectionSet";
+  SCOPE_EXIT { VLOG(1) << "Finished ~ConnectionSet"; };
+
+  StateMachineMap map;
+
+  // Set up the baton and counter of connections yet to be closed.  Move all the
```
somasun: Stale comments ?

rsocket/internal/ConnectionSet.cpp
```diff
@@ -0,0 +1,66 @@
...
+  {
+    auto locked = machines_.lock();
+    if (locked->empty()) {
+      VLOG(2) << "No connections to close, early exit";
+      return;
+    }
+
+    VLOG(2) << "Need to close " << locked->size() << " connections";
+
+    map.swap(*locked);
```
somasun: Now that we are not waiting for all the stateMachines to close, can we get rid of the copy.  The code can become cleaner (and the destruction will also be faster)

alexmalyshev: Technically this is a move and not a copy.  This way should be faster, as other threads running `RSocketStateMachine::close()` can interleave with `~ConnectionSet`, as the `~ConnectionSet` thread isn't holding the lock.

PR #750, opened 2017-09-15
Replace EmitterSubscription's mutex with an atomic bool

alexmalyshev:
It's there to prevent re-entrancy into process().  We can do that with an atomic
boolean, given that we know we won't be calling process() from multiple threads
at any given time.

### DIFF
--- a/yarpl/include/yarpl/flowable/EmitterFlowable.h
+++ b/yarpl/include/yarpl/flowable/EmitterFlowable.h
@@ -2,8 +2,12 @@
 
 #pragma once
 
+#include "yarpl/utils/credits.h"
+
+#include <folly/ScopeGuard.h>
+
+#include <atomic>
 #include <memory>
-#include <mutex>
 #include <utility>
 
 namespace yarpl {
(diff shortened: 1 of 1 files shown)


lexs:
How do we know that we won't have concurrent request() calls? folly::MicroLock might be a good choice here as it'll have the same size as a bool.

dymk:
My understanding of it is: It's out of the spec to call request() and cancel() without external synchronization, so those methods shouldn't race. The only state that needs to be protected against races are between the {on*()} signaling methods, and {request(n),cancel()} methods.

dymk:
Is this mutex in particular shown to be expensive when we profile rsocket?

vjn:
@alexmalyshev's (and @dymk's) comment above explains the constraints here. The change is fine.

lehecka:
I believe the mutex had 2 reasons which we can't get both by the atomic:
* reentrancy (breaking recursion, which the atomic can handle)
* casting memory barrier so that the Subscription methods can be called from any thread safely.

@vjn can you please add your comments here since you wrote the first version of this?

alexmalyshev:
> Is this mutex in particular shown to be expensive when we profile rsocket?

Nope, just looking to make `EmitterSubscription` smaller and simpler.

lexs:
It looks like the original code couldn't have worked, it used a `std::mutex` but expected `try_lock()` to noop if it was an re-entrant call? In fact this is UB:
> If try_lock is called by a thread that already owns the mutex, the behavior is undefined.
http://en.cppreference.com/w/cpp/thread/mutex/try_lock

Do we not have any tests for this behavior?

vjn:
Thanks @lexs.  This should have been an std::recursive_mutex. I believe we had tests depending upon, but not explicitly testing for this behavior. (Presumably we got lucky.)

PR #752, opened 2017-09-15
Multiple bug fixes and associated stress test

somasun:
We can now run resumption test with multiple clients against a single server.  It defaults to 5 clients running against a multi-threaded (10 threads) server.  Each client creates 3 streams at various points of time and cold-resumes twice during its lifetime.

I stress tested on my laptop by running 125 clients simultaneously (after that it hits system limits).  I ran 1000 iterations of the setup.

I kept repeating the experiments till there were no issues.

The bug fix descriptions are inline

### DIFF
--- a/examples/resumption/Resumption_Server.cpp
+++ b/examples/resumption/Resumption_Server.cpp
@@ -23,10 +23,7 @@ class HelloStreamRequestResponder : public RSocketResponder {
     auto requestString = request.moveDataToString();
     return Flowables::range(1, 1000)->map([name = std::move(requestString)](
         int64_t v) {
-      std::stringstream ss;
-      ss << "Hello " << name << " " << v << "!";
-      std::string s = ss.str();
-      return Payload(s, "metadata");
+      return Payload(folly::to<std::string>(v), "metadata");
     });
   }
 };
--- a/rsocket/framing/FrameTransportImpl.cpp
+++ b/rsocket/framing/FrameTransportImpl.cpp
@@ -20,7 +20,7 @@ FrameTransportImpl::FrameTransportImpl(
 }
 
 FrameTransportImpl::~FrameTransportImpl() {
-  VLOG(1) << "~FrameTransport";
+  VLOG(1) << "~FrameTransport (" << this << ")";
 }
 
 void FrameTransportImpl::connect() {
@@ -130,12 +130,12 @@ void FrameTransportImpl::terminateProcessor(folly::exception_wrapper ex) {
 }
 
 void FrameTransportImpl::onComplete() {
-  VLOG(6) << "onComplete";
+  VLOG(3) << "FrameTransport received onComplete";
   terminateProcessor(folly::exception_wrapper());
 }
 
 void FrameTransportImpl::onError(folly::exception_wrapper ex) {
-  VLOG(6) << "onError" << ex;
+  VLOG(3) << "FrameTransport received onError: " << ex.what();
   terminateProcessor(std::move(ex));
 }
 
@@ -145,7 +145,7 @@ void FrameTransportImpl::request(int64_t n) {
 }
 
 void FrameTransportImpl::cancel() {
-  VLOG(6) << "cancel";
+  VLOG(3) << "FrameTransport received cancel";
   terminateProcessor(folly::exception_wrapper());
 }
 
--- a/rsocket/statemachine/RSocketStateMachine.cpp
+++ b/rsocket/statemachine/RSocketStateMachine.cpp
@@ -87,8 +87,11 @@ bool RSocketStateMachine::resumeServer(
     yarpl::Reference<FrameTransport> frameTransport,
     const ResumeParameters& resumeParams) {
   if (!isDisconnectedOrClosed()) {
-    disconnect(folly::exception_wrapper(
-        std::runtime_error("A new transport is resuming the session")));
+    VLOG(1) << "Old connection (" << frameTransport_.get() << ") exists. "
+            << "Terminating new connection (" << frameTransport.get() << ")";
+    frameTransport->closeWithError(folly::exception_wrapper(
+        std::runtime_error("Old connection still exists")));
+    return false;
   }
   CHECK(connect(std::move(frameTransport), resumeParams.protocolVersion));
   return resumeFromPositionOrClose(
@@ -99,6 +102,7 @@ bool RSocketStateMachine::resumeServer(
 bool RSocketStateMachine::connect(
     yarpl::Reference<FrameTransport> frameTransport,
     ProtocolVersion protocolVersion) {
+  VLOG(2) << "Connecting to transport " << frameTransport.get();
   CHECK(isDisconnectedOrClosed());
   CHECK(frameTransport);
   if (protocolVersion != ProtocolVersion::Unknown) {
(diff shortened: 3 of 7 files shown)


rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -87,8 +87,11 @@ bool RSocketStateMachine::resumeServer(
     yarpl::Reference<FrameTransport> frameTransport,
     const ResumeParameters& resumeParams) {
   if (!isDisconnectedOrClosed()) {
-    disconnect(folly::exception_wrapper(
-        std::runtime_error("A new transport is resuming the session")));
+    VLOG(1) << "Old connection (" << frameTransport_.get() << ") exists. "
```
somasun: The bug fix here: 

Say server runs two threads T1 and T2.
client's statemachine runs on T1.
client has a resumed transport on T2.
client resumes very quickly again and connects on T1.  (the old connection on T2 hasn't been teared off yet).

A: In resumeServer() we make a check whether the old connection exists, if so we disconnect() that and connect to the new transport.   Since the old transport is running on T2, the FrameTransport::close() gets scheduled on T2 and the stateMachine is connected to the new transport.

B: Before FrameTransport::close() gets executed, the old connection gets actually teared from network and FrameTransport::cancel() gets called which in turn invokes FrameProcessor::onTerminal.  The onTerminal() gets scheduled on T1 and it ends up closing the new connection.

This happens very rarely and I was able to reproduce this only with stress testing (also the reconnections have to be really quick).

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -219,21 +223,23 @@ void RSocketStateMachine::closeFrameTransport(
   }
 
   if (resumeCallback_) {
-    resumeCallback_->onResumeError(ConnectionException(
+    auto resumeCallback = std::move(resumeCallback_);
+    resumeCallback->onResumeError(ConnectionException(
```
somasun: The bug fix here:

Say resumption fails for some reason and we call closeFrameTransport().  onResumeError() will then fulfill a promise.  Depending on scheduling, immediately after fulfilling the promise the thread will destroy RSocketClient (before resumeCallback_.reset() gets executed).  The destruction of RSocketClient will lead to invoking resumeCallback_->onResumeError() if resumeCallback_ is not null (which will be the case since resumeCallback_.reset() never got to run).

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -187,9 +191,9 @@ void RSocketStateMachine::close(
   VLOG(6) << "close";
 
   if (resumeCallback_) {
-    resumeCallback_->onResumeError(
+    auto resumeCallback = std::move(resumeCallback_);
```
somasun: ditto

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -219,21 +223,23 @@ void RSocketStateMachine::closeFrameTransport(
...
   }
 
   // Echo the exception to the frameTransport only if the frameTransport started
   // closing with error.  Otherwise we sent some error frame over the wire and
   // we are closing the transport cleanly.
-  if (signal == StreamCompletionSignal::CONNECTION_ERROR) {
-    frameTransport_->closeWithError(std::move(ex));
-  } else {
-    frameTransport_->close();
+  if (frameTransport_) {
```
somasun: The bug fix here:

In the same situation as above, after onResumeError() gets invoked, RSocketClient can be destroyed which can lead to frameTransport_ being reset.  And we will be accessing a nullptr here.

examples/resumption/Resumption_Server.cpp
```diff
@@ -24,7 +24,7 @@ class HelloStreamRequestResponder : public RSocketResponder {
     return Flowables::range(1, 1000)->map([name = std::move(requestString)](
         int64_t v) {
       std::stringstream ss;
-      ss << "Hello " << name << " " << v << "!";
+      ss << v;
```
phoad: Do we still need the `stringstream`?

rsocket/framing/FrameTransportImpl.cpp
```diff
@@ -20,7 +20,7 @@ FrameTransportImpl::FrameTransportImpl(
 }
 
 FrameTransportImpl::~FrameTransportImpl() {
-  VLOG(1) << "~FrameTransport";
+  VLOG(1) << "~FrameTransport (" << this << ")";
```
phoad: nit: Is it safe to call `operator<<` of `this` at the destructor? general question.

rsocket/statemachine/RSocketStateMachine.cpp
phoad: There is a TODO comment at the end of the function, stating that the return value should not be ignored. Can we do anything for that? Should we just accept a lambda and execute it at the inside of this function for the cases that we are willing to return false?

rsocket/statemachine/RSocketStateMachine.cpp
phoad: Can we do this move at the if check, so if the value is already moved then we will not need to do any check?
```
if (auto resumeCallback = std::move(resumeCallback_) {
   resumeCallback->onResumeError(...);
}
```

rsocket/statemachine/RSocketStateMachine.cpp
phoad: nit: move the move to the if check :)

test/ColdResumptionTest.cpp
```diff
@@ -93,6 +95,33 @@ class HelloResumeHandler : public ColdResumeHandler {
...
 }
 
+std::unique_ptr<rsocket::RSocketClient> createResumedClient(
+    folly::EventBase* evb,
+    uint32_t port,
+    ResumeIdentificationToken token,
+    std::shared_ptr<ResumeManager> resumeManager,
+    std::shared_ptr<ColdResumeHandler> coldResumeHandler) {
+  auto retries = 10;
+  while (true) {
```
phoad: What's the reason of retry? Is it because of having multiple threads?
Can we utilize the eventbases and manually loop them to have the treads be in sync? 
Afaik there is a class,  `folly::DeterministicSchedule`, that we can utilize.

rsocket/framing/FrameTransportImpl.cpp
somasun: the usage here should be fine.

rsocket/statemachine/RSocketStateMachine.cpp
somasun: The return value is not used anywhere.  If resumeServer() fails it will inform the client by sending a ERROR frame or by closing the transport, else it will send a RESUME_OK.  We should change the return value (probably in another PR)

test/ColdResumptionTest.cpp
somasun: createResumedClient() could fail if the old transport hasn't disconnected on the server side.  It is intentional to have some failure cases so that we can test the error paths.

rsocket/framing/FrameTransportImpl.cpp
lehecka: it prints out only address so its always safe.

PR #753, opened 2017-09-15
BaseSubscriber

dymk:
codemods to migrate most name changes:

```
codemod --accept-all -d . --extensions h,cpp 'public (yarpl::flowable::?)Subscriber' 'public \1InternalSubscriber'
codemod --accept-all -d . --extensions h,cpp 'public (flowable::?)Subscriber'        'public \1InternalSubscriber'
codemod --accept-all -d . --extensions h,cpp 'public Subscriber'                     'public InternalSubscriber'
codemod --accept-all -d . --extensions h,cpp '\bSubscriber<(.+)>::subscription'      'InternalSubscriber<\1>::subscription'
codemod --accept-all -d . --extensions h,cpp 'using (.+?)\bSubscriber;'              'using \1InternalSubscriber;'
codemod --accept-all -d . --extensions h,cpp '\bSubscriber<(.+)>::on(\w+)\('         'InternalSubscriber<\1>::on\2('
codemod --accept-all -d . --extensions h,cpp '\bSubscriber::(\w+)\('                 'InternalSubscriber::\1('
```

Changes to Subscriber
======

Makes Subscriber a pure interface to implement; renames `Subscriber -> InternalSubscriber`, adds `BaseSubscriber` which manages its `Subscription` on behalf of the subclass, and provides guarentees about signaling methods not being called after a synchronized call to cancel(). Also guarentees that cancel() and request() are nilpotent after a call to cancel(). 

`InternalSubscriber` exists as a drop-in for what used to be `Subscriber` to allow code to be incrementally changed to use `BaseSubscriber`. Above codemods will handle 98% of cases where `Subscriber` is currently used. Eventually `InternalSubscriber` can be deprecated and removed (or usage can be reduced and moved over to BaseSubscriber).

`BaseSubscriber` checks that a `Subscriber` hasn't been subscribed to multiple times (the spec is inconsistent about if this should be allowed; for now this is conservative, but the constraint can be relaxed later if need be).

New Subscriber implementations should inherit from `BaseSubscriber`, and implement: 
 * `onSubscribeImpl()` (note that it does not take a `Reference<Subscription>`)
 * `onNextImpl(T)`
 * `onCompleteImpl()`
 * `onErrorImpl(folly::exception_wrapper)`
* (optional) `onTerminateImpl()`

Subclass accesses the associated subscription via methods `request(n)` and `cancel()` on the subscriber itself. 

Why make `Subscriber` a pure interface and remove `Subscription` state?
 - Most subclasses of Subscriber add their own `Subscription` member and manage it rather than calling `Subscriber::subscription()` to get a reference to it
 - Most subclasses of Subscriber don't properly call their superclass methods, and miss out on `D?CHECK`s and reference clearing in the parent (which may lead to missed bugs and leaks)
 - The ideal would be all implementations of Subscriber can inherit a set of checks guaranteed to run
 - `BaseSubscriber` can be safely (and in a non-breaking way) extended with additional callbacks such as `onTerminate()` 
 - `BaseSubscriber` can provide a wrapping `Subscription` which provides stronger, weaker, more flexible threading guarantees based on what the application requires 
 - Forcing implementation of `onComplete`/`onError` makes a missed callback or missed `override` obvious; you're not going to be able to instantiate a subclass unless you know you've implemented all required callbacks.
 - `onTerminateImpl()` adds a singular location to put teardown code, and when cancel() and request() are called from the same thread as the one calling signaling methods, is guarenteed to be the last signaling method to run.

Changes to Reference/Refcounted: #767

### DIFF
--- a/benchmarks/Throughput.h
+++ b/benchmarks/Throughput.h
@@ -41,8 +41,10 @@ class BoundedSubscriber : public yarpl::flowable::LegacySubscriber<Payload> {
 
   void onSubscribe(
       yarpl::Reference<yarpl::flowable::Subscription> subscription) override {
-    yarpl::flowable::LegacySubscriber<Payload>::onSubscribe(std::move(subscription));
-    yarpl::flowable::LegacySubscriber<Payload>::subscription()->request(requested_);
+    yarpl::flowable::LegacySubscriber<Payload>::onSubscribe(
+        std::move(subscription));
+    yarpl::flowable::LegacySubscriber<Payload>::subscription()->request(
+        requested_);
   }
 
   void onNext(Payload) override {
--- a/benchmarks/Throughput.h
+++ b/benchmarks/Throughput.h
@@ -34,15 +34,15 @@ class FixedResponder : public RSocketResponder {
 
 /// Subscriber that requests N items and cancels the subscription once all of
 /// them arrive.  Signals a latch when it terminates.
-class BoundedSubscriber : public yarpl::flowable::Subscriber<Payload> {
+class BoundedSubscriber : public yarpl::flowable::LegacySubscriber<Payload> {
  public:
   BoundedSubscriber(Latch& latch, size_t requested)
       : latch_{latch}, requested_{requested} {}
 
   void onSubscribe(
       yarpl::Reference<yarpl::flowable::Subscription> subscription) override {
-    yarpl::flowable::Subscriber<Payload>::onSubscribe(std::move(subscription));
-    yarpl::flowable::Subscriber<Payload>::subscription()->request(requested_);
+    yarpl::flowable::LegacySubscriber<Payload>::onSubscribe(std::move(subscription));
+    yarpl::flowable::LegacySubscriber<Payload>::subscription()->request(requested_);
   }
 
   void onNext(Payload) override {
@@ -51,19 +51,19 @@ class BoundedSubscriber : public yarpl::flowable::Subscriber<Payload> {
       latch_.post();
 
       // After this cancel we could be destroyed.
-      yarpl::flowable::Subscriber<Payload>::subscription()->cancel();
+      yarpl::flowable::LegacySubscriber<Payload>::subscription()->cancel();
     }
   }
 
   void onComplete() override {
-    yarpl::flowable::Subscriber<Payload>::onComplete();
+    yarpl::flowable::LegacySubscriber<Payload>::onComplete();
     if (!terminated_.exchange(true)) {
       latch_.post();
     }
   }
 
   void onError(folly::exception_wrapper) override {
-    yarpl::flowable::Subscriber<Payload>::onError({});
+    yarpl::flowable::LegacySubscriber<Payload>::onError({});
     if (!terminated_.exchange(true)) {
       latch_.post();
     }
--- a/benchmarks/Throughput.h
+++ b/benchmarks/Throughput.h
@@ -34,16 +34,16 @@ class FixedResponder : public RSocketResponder {
 
 /// Subscriber that requests N items and cancels the subscription once all of
 /// them arrive.  Signals a latch when it terminates.
-class BoundedSubscriber : public yarpl::flowable::LegacySubscriber<Payload> {
+class BoundedSubscriber : public yarpl::flowable::InternalSubscriber<Payload> {
  public:
   BoundedSubscriber(Latch& latch, size_t requested)
       : latch_{latch}, requested_{requested} {}
 
   void onSubscribe(
       yarpl::Reference<yarpl::flowable::Subscription> subscription) override {
-    yarpl::flowable::LegacySubscriber<Payload>::onSubscribe(
+    yarpl::flowable::InternalSubscriber<Payload>::onSubscribe(
         std::move(subscription));
-    yarpl::flowable::LegacySubscriber<Payload>::subscription()->request(
+    yarpl::flowable::InternalSubscriber<Payload>::subscription()->request(
         requested_);
   }
 
@@ -53,19 +53,19 @@ class BoundedSubscriber : public yarpl::flowable::LegacySubscriber<Payload> {
       latch_.post();
 
       // After this cancel we could be destroyed.
-      yarpl::flowable::LegacySubscriber<Payload>::subscription()->cancel();
+      yarpl::flowable::InternalSubscriber<Payload>::subscription()->cancel();
     }
   }
 
   void onComplete() override {
-    yarpl::flowable::LegacySubscriber<Payload>::onComplete();
+    yarpl::flowable::InternalSubscriber<Payload>::onComplete();
     if (!terminated_.exchange(true)) {
       latch_.post();
     }
   }
 
   void onError(folly::exception_wrapper) override {
-    yarpl::flowable::LegacySubscriber<Payload>::onError({});
+    yarpl::flowable::InternalSubscriber<Payload>::onError({});
     if (!terminated_.exchange(true)) {
       latch_.post();
     }
--- a/yarpl/include/yarpl/flowable/Subscriber.h
+++ b/yarpl/include/yarpl/flowable/Subscriber.h
@@ -22,6 +22,7 @@ class Subscriber : public virtual Refcounted, public yarpl::enable_get_ref {
 };
 
 // codemod all things that inherit from Subscriber to inherit from
+// InternalSubscriber
 template <typename T>
 class InternalSubscriber : public Subscriber<T> {
  public:
--- a/yarpl/include/yarpl/flowable/Subscriber.h
+++ b/yarpl/include/yarpl/flowable/Subscriber.h
@@ -21,9 +21,10 @@ class Subscriber : public virtual Refcounted, public yarpl::enable_get_ref {
   virtual void onNext(T) = 0;
 };
 
-// codemod all things that inherit from Subscriber to inherit from LegacySubscriber
+// codemod all things that inherit from Subscriber to inherit from
+// InternalSubscriber
 template <typename T>
-class LegacySubscriber : public Subscriber<T> {
+class InternalSubscriber : public Subscriber<T> {
  public:
   // Note: If any of the following methods is overridden in a subclass, the new
   // methods SHOULD ensure that these are invoked as well.
@@ -54,6 +55,12 @@ class LegacySubscriber : public Subscriber<T> {
   Reference<Subscription> subscription_;
 };
 
+#define SUBSCRIBER_KEEP_SELF() \
+  Reference<SafeSubscriber> self; \
+  if (!keeps_reference_to_this) { \
+    self = this->ref_from_this(this); \
+  }
+
 template <typename T>
 class SafeSubscriber : public Subscriber<T> {
  public:
@@ -62,6 +69,8 @@ class SafeSubscriber : public Subscriber<T> {
   void onSubscribe(Reference<Subscription> subscription) final override {
     DCHECK(subscription);
     CHECK(!subscription_);
+
+    SUBSCRIBER_KEEP_SELF()
     subscription_ = std::move(subscription);
     onSubscribeImpl();
   }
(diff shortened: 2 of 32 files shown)


phoad:
Why do we need to have two versions of Subscriber, let's have one safe version alone? 
Might we do it once in everywhere fix it once and get rid of all faulty usages? Do we have lot's of users that use the Subscriber wrong way, and will it take too much effort to fix them?

benchmarks/Throughput.h
```diff
@@ -34,15 +34,17 @@ class FixedResponder : public RSocketResponder {
 
 /// Subscriber that requests N items and cancels the subscription once all of
 /// them arrive.  Signals a latch when it terminates.
-class BoundedSubscriber : public yarpl::flowable::Subscriber<Payload> {
+class BoundedSubscriber : public yarpl::flowable::LegacySubscriber<Payload> {
  public:
   BoundedSubscriber(Latch& latch, size_t requested)
       : latch_{latch}, requested_{requested} {}
 
   void onSubscribe(
```
phoad: Would it be enough to 
 - remove the `virtual` keyword from this function only 
 - and introduce another virtual function - which is `onSubscribe()`, which doesn't have any parameters
So it will force the child classes to call `subscription()` function to get the subscription. 
That way, we can be sure that this parent class has full autonomy over the Subscription instance and the inherited classes are forced to obey it.

dymk:
@phoad 
There is only one version of `Subscriber`. `Subscriber` becomes a proper interface, instead of having implementation tied to it. `{Legacy,Safe,SafeDefault}Subscriber` are partial implementations which simplify writing correct and non-repetative Subscribers in the application (for the Safe and SafeDefault case). 

Eventually `LegacySubscriber` goes away. I didn't want to start migrating anything over (except in one or two places to exercise it) to `SafeSubscriber` yet because I'm sure we're going to iterate a bit on the design.

phoad:
Let's create an `issue` for this. We can discuss this as an issue or in the discussion group (rsocket engineering). We can ask people to join and comment if they would like to by pointing that issue link.

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -15,40 +15,96 @@ namespace flowable {
...
+    CHECK(!subscription_);
     subscription_ = std::move(subscription);
+    onSubscribeImpl();
   }
 
   // No further calls to the subscription after this method is invoked.
-  virtual void onComplete() {
+  void onComplete() final override {
     DCHECK(subscription_) << "Calling onComplete() without a subscription";
     subscription_.reset();
```
alexmalyshev: I think this needs to be `auto sub = std::move(subscription_);` or `auto sub = std::atomic_exchange(&subscription_, nullptr);` in a thread-safe world.

The last reference to the `Subscriber` could be the `Subscription`, so we don't want to kill it just yet.  The easy test case is to make a cycle between a `Subscriber` and a `Subscription`, grab a raw pointer to the `Subscriber`, leak it, and then call `rawSubscriber->onComplete();`.

dymk:
@lehecka 
Agreed regarding SafeDefaultSubscription; I thought it would be more useful, but it only saved implementing signal methods in one or two places in the codebase. It's otherwise a net increase in LOC. 

Regarding keeping `Subscriber` as it is for internal use; that's basically what's already happening in this PR, it's just been renamed to LegacySubscriber. It can be renamed to InternalSubscriber or something if that's better. 

If `Subscriber` is kept unmodified and as an abstract class (rather than a pure interface), `SafeSubscriber` will need to awkwardly inherit from it. If we're going for a pure reactive streams implementation, we want to get all of the implementation out of Subscription anyways; it's an interface in Reactive Streams anyways.

yarpl/include/yarpl/flowable/Subscriber.h
dymk: Spec states that signaling methods won't be called without external synchronization, so we shouldn't have any races here. 

However you're right, we might kill ourselves when we reset the subscription object, so let's hold it on the stack until we're done.

dymk: Actually it's slightly worse than that; a terminating signal method might do something that destroys `this` (a bug we've seen often); if this is SafeSubscriber we should keep a reference to ourselves on the stack.

phoad:
@dymk please don't check this in up till we have time to do a nice review of ideas around. You might start a discussion and we can focus on it in the following week.

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -15,40 +15,119 @@ namespace flowable {
...
-  virtual void onComplete() {
+  void onComplete() final override {
     DCHECK(subscription_) << "Calling onComplete() without a subscription";
-    subscription_.reset();
+
+    Reference<SafeSubscriber> self;
+    if(!keeps_reference_to_this) {
+      self = ref_from_this(this);
+    }
+    auto sub = std::move(subscription_);
```
phoad: Destroy the subscription after the `onCompleteImpl` is called.

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -15,40 +15,119 @@ namespace flowable {
...
+
+  virtual void onSubscribeImpl() = 0;
+  virtual void onCompleteImpl() = 0;
+  virtual void onNextImpl(T) = 0;
+  virtual void onErrorImpl(folly::exception_wrapper) = 0;
 
  protected:
-  Subscription* subscription() {
-    return subscription_.operator->();
+  Reference<Subscription> subscription() {
```
phoad: Ownership of the `subscription_` seems like an issue that makes the job of classes that extends Subscriber harder. Can we have a function to detach the `subscription_` from the parent class if child class wants to own the object.

Like, we call `onSubscribe(void)`, but the child class want to force to have the ownership of the `subscription_`, so instead of calling `subscription()` function, it wants `detachSubscription() : Reference<Subscription>`, which gets rid of any reference in the parent class.

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -15,40 +15,119 @@ namespace flowable {
...
+  virtual void onSubscribe(Reference<Subscription>) = 0;
+  virtual void onComplete() = 0;
+  virtual void onError(folly::exception_wrapper) = 0;
+  virtual void onNext(T) = 0;
+};
+
+// codemod all things that inherit from Subscriber to inherit from
+// InternalSubscriber
+template <typename T>
+class InternalSubscriber : public Subscriber<T> {
```
phoad: Do we ever need this class?
What it does it only keeping a `subscription_` and just calls `reset` on this object. It is one line of code that the class which extends this one can do easily. 
Do we see that we are going to have any instance of this class at all?
So just comments in the interface should suffice, so we would not force the implementers of the interface to have `subscription_` object be in the parent class.

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -15,40 +15,119 @@ namespace flowable {
...
-  virtual void onSubscribe(Reference<Subscription> subscription) {
+  void onSubscribe(Reference<Subscription> subscription) final override {
     DCHECK(subscription);
+    CHECK(!subscription_);
 
-    if (subscription_) {
-      subscription->cancel();
-      return;
+    Reference<SafeSubscriber> self;
+    if(!keeps_reference_to_this) {
```
phoad: I am not sure if we need this, is it a `premature` optimization that might lead issues? Will it affect the performance noticeably?

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -15,40 +15,119 @@ namespace flowable {
...
+  void onSubscribe(Reference<Subscription> subscription) override {
+    DCHECK(subscription);
+    CHECK(!subscription_);
+    subscription_ = std::move(subscription);
+  }
+
+  // No further calls to the subscription after this method is invoked.
+  void onComplete() override {
+    DCHECK(subscription_) << "Calling onComplete() without a subscription";
+    subscription_.reset();
```
phoad: I would ask one question about this. What if we think in reverse, how? Instead of keeping `subscription_` object in the parent class and providing a function to the child class to access this object, can we have a function, which is virtual that the parent class can call and get access to the `subscription_` data member in the child class?

So parent will not keep and reference to the `subscription_` object, but still can access it through that virtual function call. 

In that case, the parent will never need to call `subscription_.reset()` at all, as it already doesn't hold any reference to the `subscription_` permanently.

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -15,40 +15,119 @@ namespace flowable {
...
+  }
+
+  // No further calls to the subscription after this method is invoked.
+  void onError(folly::exception_wrapper) override {
+    DCHECK(subscription_) << "Calling onError() without a subscription";
+    subscription_.reset();
+  }
+
+ protected:
+  Reference<Subscription> subscription() {
```
phoad: What if we make this function `pure virtual`? Can it help getting rid of the `subscription_` data member in the parent class?

yarpl/include/yarpl/flowable/Subscriber.h
dymk: `InternalSubscriber` is verbatim what `Subscriber` used to be

lehecka: Can we get rid of the InternalSubscriber and use the old Subscriber instead? In the next diff we can remove the subscription pointer from the Subscriber and fix the places which were using it.

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -15,40 +15,129 @@ namespace flowable {
...
+// Classes that ensure this can set keeps_reference_to_this to `true`
+// as an optimization (to prevent an atomic inc/dec pair)
+#define SUBSCRIBER_KEEP_SELF() \
+  Reference<SafeSubscriber> self; \
+  if (!keeps_reference_to_this) { \
+    self = this->ref_from_this(this); \
+  }
+
+template <typename T, bool keeps_reference_to_this = false>
+class SafeSubscriber : public Subscriber<T> {
```
lehecka: Lets rename this to SubscriberBase.

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -15,40 +15,129 @@ namespace flowable {
...
+// to the object within a method of that object)
+// Classes that ensure this can set keeps_reference_to_this to `true`
+// as an optimization (to prevent an atomic inc/dec pair)
+#define SUBSCRIBER_KEEP_SELF() \
+  Reference<SafeSubscriber> self; \
+  if (!keeps_reference_to_this) { \
+    self = this->ref_from_this(this); \
+  }
+
+template <typename T, bool keeps_reference_to_this = false>
```
lehecka: I don't see any uses of keeps_reference_to_this being any different than default false.
Can we discuss it in a separate PR? It doesn't seem to be essential to the SubscriberBase.

dymk:
AtomicReference and its helpers will be best broken out into a separate PR, yeah. I'll do that now that it looks like we know what we want the SafeSubscriber API to look like

lehecka:
needs rebasing and answers to the rest of the comments..

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -15,40 +15,118 @@ namespace flowable {
...
     DCHECK(subscription_) << "Calling onError() without a subscription";
     subscription_.reset();
   }
 
-  virtual void onNext(T) = 0;
-
  protected:
-  Subscription* subscription() {
-    return subscription_.operator->();
+  Reference<Subscription> subscription() {
```
phoad: I thought we would not have any reference to `subscription_`, but we would just call the cancel & request using a pure virtual method. @alexmalyshev, @lehecka

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -15,40 +15,118 @@ namespace flowable {
...
+  }
+
+  void onNext(T t) final override {
+    if(auto sub = subscription_.load()) {
+      onNextImpl(std::move(t));
+    }
+  }
+
+  void cancel() {
+    if(auto sub = subscription_.exchange(nullptr)) {
```
phoad: I thought we would not have any reference to subscription_, but we would just call the cancel & request using a pure virtual method. @alexmalyshev, @lehecka

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -15,40 +15,118 @@ namespace flowable {
...
+
+    if(auto sub = subscription_.exchange(nullptr)) {
+      onErrorImpl(std::move(e));
+      onTerminateImpl();
+    }
+  }
+
+  void onNext(T t) final override {
+    if(auto sub = subscription_.load()) {
+      onNextImpl(std::move(t));
```
phoad: try/catch this and close the subscriber with onError if this call throws.
```
try{ onNextImpl(..) } catch( exception ) {.. onError(..) }
```
If onNextImpl throws, discard rest of the onNext calls.

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -15,40 +15,118 @@ namespace flowable {
...
+    }
+  }
+
+  void request(int64_t n) {
+    if(auto sub = subscription_.load()) {
+      sub->request(n);
+    }
+  }
+
+  virtual void onSubscribeImpl() = 0;
```
phoad: Why don't we provide a default/empty implementation for these functions?
Do we have to have implementation for all of them in the child classes?

PR #754, opened 2017-09-15
Channel tests

ragansa:
The last 2 tests:
- failure on request, requester sees
- failure from requester ... what happens?

I'm not clear on what this exactly means.  I can have the requester stream send an onError, is that failure on request?  How does the requester "see" this?

### DIFF
--- a/rsocket/statemachine/ConsumerBase.cpp
+++ b/rsocket/statemachine/ConsumerBase.cpp
@@ -31,6 +31,7 @@ void ConsumerBase::subscribe(
 // completeConsumer exists)
 void ConsumerBase::cancelConsumer() {
   state_ = State::CLOSED;
+  VLOG(5) << "ConsumerBase::cancelConsumer()";
   consumingSubscriber_ = nullptr;
 }
 
@@ -75,6 +76,7 @@ void ConsumerBase::processPayload(Payload&& payload, bool onNext) {
 
 void ConsumerBase::completeConsumer() {
   state_ = State::CLOSED;
+  VLOG(5) << "ConsumerBase::completeConsumer()";
   if (auto subscriber = std::move(consumingSubscriber_)) {
     subscriber->onComplete();
   }
@@ -82,6 +84,7 @@ void ConsumerBase::completeConsumer() {
 
 void ConsumerBase::errorConsumer(folly::exception_wrapper ex) {
   state_ = State::CLOSED;
+  VLOG(5) << "ConsumerBase::errorConsumer()";
   if (auto subscriber = std::move(consumingSubscriber_)) {
     subscriber->onError(std::move(ex));
   }
@@ -103,4 +106,4 @@ void ConsumerBase::handleFlowControlError() {
   }
   errorStream("flow control error");
 }
-}
+} // namespace rsocket
--- a/test/RequestChannelTest.cpp
+++ b/test/RequestChannelTest.cpp
@@ -88,22 +88,18 @@ TEST(RequestChannelTest, RequestOnDisconnectedClient) {
   ASSERT(did_call_on_error);
 }
 
-class ResponderLongOutPut : public rsocket::RSocketResponder {
+class ResponderLongStream : public rsocket::RSocketResponder {
  public:
   yarpl::Reference<Flowable<rsocket::Payload>> handleRequestChannel(
       rsocket::Payload,
       yarpl::Reference<Flowable<rsocket::Payload>> requestStream,
       rsocket::StreamId) override {
     auto ts = TestSubscriber<std::string>::create();
-    requestStream
-        ->map([](auto p) { LOG(INFO) << "received: " << p;
-          return p.moveDataToString(); })
+    requestStream->map([](auto p) { return p.moveDataToString(); })
         ->subscribe(ts);
 
     // output 1 - 100
-    return Flowables::range(1, 1000)->map([](
-        int64_t v) {
-      wait();
+    return Flowables::range(1, 100)->map([](int64_t v) {
       std::stringstream ss;
       ss << "Server stream: " << v;
       std::string s = ss.str();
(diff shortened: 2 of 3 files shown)


test/RequestChannelTest.cpp
```diff
@@ -88,11 +88,295 @@ TEST(RequestChannelTest, RequestOnDisconnectedClient) {
...
+  auto requester = client->getRequester();
+
+  auto requestSubscriber = TestSubscriber<std::string>::create(100);
+  auto responderSubscriber = responder->getChannelSubscriber();
+
+  int64_t requesterRangeEnd = 100;
+
+  auto requesterFlowable =
+      Flowables::range(1, requesterRangeEnd)->map([&](int64_t v) {
+        std::stringstream ss;
```
phoad: Yield might help for this test to pass.
```
    std::this_thread::yield();
```

ragansa:
That should solve the cancel error, still looking for some signal regarding what the final two test cases mean:

    failure on request, requester sees
    failure from requester ... what happens?

What kind of failure "on request"?  And what is meant by "requester sees"?
Then, what is different between failure on request, and failure from requester?  (and what does happen? hehe 😃 )

test/RequestChannelTest.cpp
```diff
@@ -88,11 +88,299 @@ TEST(RequestChannelTest, RequestOnDisconnectedClient) {
...
+  responderSubscriber->assertSuccess();
+
+  requestSubscriber->assertValueAt(0, "Responder stream: 1 of 10");
+  requestSubscriber->assertValueAt(9, "Responder stream: 10 of 10");
+
+  responderSubscriber->assertValueAt(0, "Requester stream: 1 of 10");
+  responderSubscriber->assertValueAt(9, "Requester stream: 10 of 10");
+}
+
+TEST(RequestChannelTest, DISABLED_CancelFromRequester) {
```
ragansa: Yup, the leak is definitely a result of running the cancel test.

ragansa: Leak when CancelFromRequester runs:
https://travis-ci.org/rsocket/rsocket-cpp/jobs/277537750#L1984

I can't reproduce this locally, does anyone have any ideas?

test/RequestChannelTest.cpp
```diff
@@ -88,11 +88,299 @@ TEST(RequestChannelTest, RequestOnDisconnectedClient) {
...
+        requestSubscriber->cancel();
+        requestSubscriber->request(50);
+      });
+
+  responderSubscriber->awaitTerminalEvent();
+  EXPECT_LT(requestSubscriber->getValueCount(), 100);
+
+  // Responder Subscriber should be at 100
+  // Requester Subscriber should be < 100
+  LOG(INFO) << "Responder Subscriber: " << responderSubscriber->getValueCount();
```
phoad: VLOG?

test/RequestChannelTest.cpp
```diff
@@ -88,11 +88,299 @@ TEST(RequestChannelTest, RequestOnDisconnectedClient) {
...
+  // failure streaming from Responder
+  requestSubscriber->awaitTerminalEvent();
+  requestSubscriber->assertOnErrorMessage("A wild Error appeared!");
+
+  responderSubscriber->awaitTerminalEvent();
+  responderSubscriber->assertValueAt(0, "Requester stream: 1 of 10");
+  responderSubscriber->assertValueAt(9, "Requester stream: 10 of 10");
+}
+
+TEST(RequestChannelTest, DISABLED_FailureOnRequestRequesterSees) {
```
phoad: Let's have a chat session with Alex and learn the details of this.

test/RequestChannelTest.cpp
```diff
@@ -88,11 +88,299 @@ TEST(RequestChannelTest, RequestOnDisconnectedClient) {
...
+        testSubscriber_{TestSubscriber<std::string>::create(initialSubReq)} {}
+
+  yarpl::Reference<Flowable<rsocket::Payload>> handleRequestChannel(
+      rsocket::Payload initialPayload,
+      yarpl::Reference<Flowable<rsocket::Payload>> requestStream,
+      rsocket::StreamId) override {
+    // add initial payload to testSubscriber values list
+    testSubscriber_->onNext(initialPayload.moveDataToString());
+
+    requestStream->map([](auto p) { return p.moveDataToString(); })
```
phoad: Looks like travis is pointing at this line as where the leak happens for the `cancel` test. I would test copying the payload instead of moving it and see if it fails still.

rsocket/statemachine/ConsumerBase.cpp
```diff
@@ -64,8 +65,12 @@ void ConsumerBase::processPayload(Payload&& payload, bool onNext) {
     // Frames carry application-level payloads are taken into account when
     // figuring out flow control allowance.
     if (allowance_.tryAcquire()) {
-      sendRequests();
-      consumingSubscriber_->onNext(std::move(payload));
+      VLOG(5) << "ConsumerBase::processPayload(" << onNext << ")";
+      if (consumingSubscriber_) {
```
lehecka: curious. under which circumstances is the consumingSubscriber_ null?

test/RequestChannelTest.cpp
```diff
@@ -88,11 +88,299 @@ TEST(RequestChannelTest, RequestOnDisconnectedClient) {
...
+      int64_t initialSubReq = credits::kNoFlowControl)
+      : rangeEnd_{rangeEnd},
+        testSubscriber_{TestSubscriber<std::string>::create(initialSubReq)} {}
+
+  yarpl::Reference<Flowable<rsocket::Payload>> handleRequestChannel(
+      rsocket::Payload initialPayload,
+      yarpl::Reference<Flowable<rsocket::Payload>> requestStream,
+      rsocket::StreamId) override {
+    // add initial payload to testSubscriber values list
+    testSubscriber_->onNext(initialPayload.moveDataToString());
```
lehecka: I guess the testSubscriber doesn't mind you don't call onSubscribe before onNext?

test/RequestChannelTest.cpp
```diff
@@ -88,11 +88,299 @@ TEST(RequestChannelTest, RequestOnDisconnectedClient) {
...
+  responderSubscriber->awaitTerminalEvent();
+  responderSubscriber->assertValueAt(0, "Requester stream: 1 of 10");
+  responderSubscriber->assertValueAt(9, "Requester stream: 10 of 10");
+}
+
+TEST(RequestChannelTest, DISABLED_FailureOnRequestRequesterSees) {
+  // ???
+}
+
+TEST(RequestChannelTest, DISABLED_FailureFromRequester) {
```
lehecka: ??

test/RequestChannelTest.cpp
lehecka: I would guess the leak is that the allocated string is put in the TestSubscriber which is held by the TestChannelResponder which is held by the rsocket server? Can you verify that the responder is released?
Also you can locally compile the yarpl library with Dylan's new switch which will allow you to call "print_ref_counted_objects" at the end of the test so you can see what refcounted objects are still alive.

test/RequestChannelTest.cpp
ragansa: The description for these test cases was:
- failure on request, requester sees
- failure from requester ... what happens?
I'm not clear on what this means.

test/RequestChannelTest.cpp
ragansa: All it does is add it to it's internal vector of values.  Adding the initial payload let me stream the same things both directions and then make sure they were the same.  It's not truly necessary.

rsocket/statemachine/ConsumerBase.cpp
ragansa: @dymk had initially added this to fix a null ptr error.  This PR has gone through a lot of iterations.

test/RequestChannelTest.cpp
lehecka: I don't know what it means either. Let's remove this then.

rsocket/statemachine/ConsumerBase.cpp
lehecka: under which circumstances we get into the nullptr error? We either need to understand this or not add it in the PR. If this is a fix of something else, it should be in a different PR.

ragansa: Acknowledged

ragansa:
Removed the error test, and the consumerBase change that was trying to fix for the error test.
Let's land these, and I'll address the error test in another PR

PR #757, opened 2017-09-18
Fix include; cosmetic changes

somasun:


test/test_utils/ColdResumeManager.cpp
```diff
@@ -1,6 +1,6 @@
 // Copyright 2004-present Facebook. All Rights Reserved.
 
-#include "test/test_utils/ColdResumeManager.h"
```
phoad: We had this issue for one more header in fbcode. I wonder what's the real cause of requiring this update?

PR #764, opened 2017-09-22
Add separate method to persist state

somasun:


test/test_utils/ColdResumeManager.cpp
```diff
@@ -104,7 +104,9 @@ ColdResumeManager::ColdResumeManager(
   }
 }
 
-ColdResumeManager::~ColdResumeManager() {
+ColdResumeManager::~ColdResumeManager() {}
```
phoad: nit: remove?

test/test_utils/ColdResumeManager.cpp
```diff
@@ -138,9 +140,8 @@ ColdResumeManager::~ColdResumeManager() {
     f << jsonState;
     f.close();
   } catch (const std::exception& ex) {
-    LOG(ERROR) << "Persisting state to " << outputFile_ << " failed. "
-               << ex.what();
-    return;
+    throw std::runtime_error(folly::sformat(
```
phoad: Should we give error, or LOG anything if the `output_file.empty()` branch is taken?

phoad: Why not throw an error, saying that the filename is empty, as we already throw.

examples/resumption/ColdResumption_Client.cpp
```diff
@@ -136,19 +136,14 @@ int main(int argc, char* argv[]) {
...
-            // This scheduled event ensures that ResumeManager gets destroyed
-            // after RSocketClient has destroyed.  And then we exit the scope.
+    worker.getEventBase()->runInEventBaseThreadAndWait(
+        [resumeManager = std::move(resumeManager)]() {
+          // We want to persist state after RSocketStateMachine of the client
+          // has been completely destroyed and before we start the next scope.
+          // Since the RSocketStateMachine's destruction proceeds
+          // asynchronously in worker thread, we have to schedule the
+          // persistence in the worker thread.
+          resumeManager->persistState();
```
phoad: Should this be called at the destructor of the ColdResumeManager?

PR #765, opened 2017-09-22
Client pushes data immediately without any request, which causes server to crash

phoad:
EagerSubscriptionBridge classes onError method gets called even prior to onSubscribe method of it is being called. So it causes the server to crash.

The failure is caused by a thrown exception: 
```
V0922 14:19:00.530442 1238641 RSocketStateMachine.cpp:575] SERVER In: PAYLOAD[NEXT, 1], Metadata(0): <null>, Data(5): '\x04Fuat'
V0922 14:19:00.530485 1238641 RSocketResponder.cpp:73] handleRequestChannelCore::onError: std::runtime_error: surplus response
F0922 14:19:00.530529 1238641 RSocketResponder.cpp:74] Check failed: inner_
```
When client tries to push data immediately without waiting a request to be performed from server to client.

This update eliminates the crash and successfully sends the error through the wire to the client.

### DIFF
--- a/rsocket/RSocketResponder.cpp
+++ b/rsocket/RSocketResponder.cpp
@@ -63,21 +63,19 @@ RSocketResponder::handleRequestChannelCore(
 
     void onComplete() noexcept override {
       DCHECK(inner_);
-      inner_->onComplete();
-
-      inner_.reset();
-      subscription_.reset();
+      if (auto inner = std::move(inner_)) {
+        inner->onComplete();
+        subscription_.reset();
+      } else {
+        completed_ = true;
+      }
     }
 
     void onError(folly::exception_wrapper ex) noexcept override {
       VLOG(3) << "handleRequestChannelCore::onError: " << ex.what();
-      std::lock_guard<std::mutex> guard(errorMutex_);
-      if (inner_) {
-        inner_->onError(std::move(ex));
-
-        inner_.reset();
+      if (auto inner = std::move(inner_)) {
+        inner->onError(std::move(ex));
         subscription_.reset();
-        error_.reset();
       } else {
         error_ = std::move(ex);
       }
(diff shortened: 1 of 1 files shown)


rsocket/RSocketResponder.cpp
```diff
@@ -70,35 +70,51 @@ RSocketResponder::handleRequestChannelCore(
...
+      std::lock_guard<std::mutex> guard(errorMutex_);
+
       inner_ = std::move(inner);
       if (subscription_) {
         inner_->onSubscribe(subscription_);
       }
+
+      // it's possible to get an error before subscribe happens, delay sending
+      // it but send it when this class gets subscribed
+      if (error_) {
```
alexmalyshev: You can move this inside of the `if (subscription_)` check, and also do `onError(std::move(error_))`.

rsocket/RSocketResponder.cpp
```diff
@@ -70,35 +70,51 @@ RSocketResponder::handleRequestChannelCore(
...
+      VLOG(3) << "handleRequestChannelCore::onError: " << ex.what();
+      std::lock_guard<std::mutex> guard(errorMutex_);
+      if (inner_) {
+        inner_->onError(std::move(ex));
 
-      inner_.reset();
-      subscription_.reset();
+        inner_.reset();
+        subscription_.reset();
+        error_.reset();
```
alexmalyshev: This `error_.reset()` shouldn't be necessary if we do `onError(std::move(error_))` in `subscribe()`.

rsocket/RSocketResponder.cpp
```diff
@@ -70,35 +70,51 @@ RSocketResponder::handleRequestChannelCore(
     }
 
     void onError(folly::exception_wrapper ex) noexcept override {
-      DCHECK(inner_);
-      inner_->onError(std::move(ex));
+      VLOG(3) << "handleRequestChannelCore::onError: " << ex.what();
+      std::lock_guard<std::mutex> guard(errorMutex_);
+      if (inner_) {
```
alexmalyshev: Shorter to do

```
  if (auto inner = std::move(inner_)) {
    inner->onError(std::move(ex));
    subscription_.reset();
  }
```

rsocket/RSocketResponder.cpp
```diff
@@ -70,35 +70,51 @@ RSocketResponder::handleRequestChannelCore(
...
+      // it but send it when this class gets subscribed
+      if (error_) {
+        onError(error_);
+      }
     }
 
    private:
     yarpl::Reference<yarpl::flowable::Subscriber<rsocket::Payload>> inner_;
     yarpl::Reference<yarpl::flowable::Subscription> subscription_;
+    std::mutex errorMutex_;
```
alexmalyshev: Do we actually need the mutex?  Are we calling `onError` and `subscribe` from different threads?

rsocket/RSocketResponder.cpp
```diff
@@ -63,42 +63,56 @@ RSocketResponder::handleRequestChannelCore(
...
 
     void subscribe(
         yarpl::Reference<yarpl::flowable::Subscriber<rsocket::Payload>> inner) {
       CHECK(!inner_); // only one call to subscribe is supported
       CHECK(inner);
+
       inner_ = std::move(inner);
       if (subscription_) {
         inner_->onSubscribe(subscription_);
+        // it's possible to get an error before subscribe happens, delay sending
```
alexmalyshev: No longer just an error, but any of the two possible termination signals.

PR #768, opened 2017-09-26
Handle exceptions thrown from map operators

alexmalyshev:
If the function passed to a map operator throws, then we should cancel the
subscription, and pass the exception as the onError to the downstream (unless
this is a Single, where we shouldn't cancel the subscription).

Improves the SingleOperator termination semantics, which haven't kept up with
ObservableOperator and FlowableOperator.

### DIFF
--- a/yarpl/include/yarpl/flowable/FlowableOperator.h
+++ b/yarpl/include/yarpl/flowable/FlowableOperator.h
@@ -132,12 +132,12 @@ class FlowableOperator : public Flowable<D> {
       auto self = this->ref_from_this(this);
 
       auto upstream = std::move(upstream_);
-      if(state.up && upstream) {
+      if (state.up && upstream) {
         upstream->cancel();
       }
 
       auto subscriber = std::move(subscriber_);
-      if(state.down && subscriber) {
+      if (state.down && subscriber) {
         if (ex) {
           subscriber->onError(std::move(ex));
         } else {
@@ -179,8 +179,8 @@ class MapOperator : public FlowableOperator<U, D, MapOperator<U, D, F>> {
       : Super(std::move(upstream)), function_(std::move(function)) {}
 
   void subscribe(Reference<Subscriber<D>> subscriber) override {
-    Super::upstream_->subscribe(
-        make_ref<Subscription>(this->ref_from_this(this), std::move(subscriber)));
+    Super::upstream_->subscribe(make_ref<Subscription>(
+        this->ref_from_this(this), std::move(subscriber)));
   }
 
  private:
@@ -193,8 +193,13 @@ class MapOperator : public FlowableOperator<U, D, MapOperator<U, D, F>> {
         : SuperSubscription(std::move(flowable), std::move(subscriber)) {}
 
     void onNext(U value) override {
-      auto&& map = SuperSubscription::getFlowableOperator();
-      SuperSubscription::subscriberOnNext(map->function_(std::move(value)));
+      try {
+        auto&& map = this->getFlowableOperator();
+        this->subscriberOnNext(map->function_(std::move(value)));
+      } catch (const std::exception& exn) {
+        folly::exception_wrapper ew{std::current_exception(), exn};
+        this->terminateErr(std::move(ew));
+      }
     }
   };
 
@@ -216,8 +221,8 @@ class FilterOperator : public FlowableOperator<U, U, FilterOperator<U, F>> {
       : Super(std::move(upstream)), function_(std::move(function)) {}
 
   void subscribe(Reference<Subscriber<U>> subscriber) override {
-    Super::upstream_->subscribe(
-        make_ref<Subscription>(this->ref_from_this(this), std::move(subscriber)));
+    Super::upstream_->subscribe(make_ref<Subscription>(
+        this->ref_from_this(this), std::move(subscriber)));
   }
 
  private:
@@ -258,8 +263,8 @@ class ReduceOperator : public FlowableOperator<U, D, ReduceOperator<U, D, F>> {
       : Super(std::move(upstream)), function_(std::move(function)) {}
 
   void subscribe(Reference<Subscriber<D>> subscriber) override {
-    Super::upstream_->subscribe(
-        make_ref<Subscription>(this->ref_from_this(this), std::move(subscriber)));
+    Super::upstream_->subscribe(make_ref<Subscription>(
+        this->ref_from_this(this), std::move(subscriber)));
   }
 
  private:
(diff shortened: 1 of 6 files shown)


yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -193,8 +193,13 @@ class MapOperator : public FlowableOperator<U, D, MapOperator<U, D, F>> {
         : SuperSubscription(std::move(flowable), std::move(subscriber)) {}
 
     void onNext(U value) override {
-      auto&& map = SuperSubscription::getFlowableOperator();
-      SuperSubscription::subscriberOnNext(map->function_(std::move(value)));
+      try {
+        auto&& map = this->getFlowableOperator();
+        this->subscriberOnNext(map->function_(std::move(value)));
+      } catch (const std::exception& exn) {
+        folly::exception_wrapper ew{std::current_exception(), exn};
```
phoad: I believe we need to have this try/catch at the base Subscriber.
If we receive an exception when onNext is called, it should be caught and onError should be called.

alexmalyshev:
> I believe we should move the try/catch to the yarpl::flowable::Subscriber.

> If we receive any exception from the user code, it should be caught and onError should be called, instead of propagating the error up.

On top of calling `onError` on the downstream, we also want to `cancel` the upstream when this happens because we're an operator.  This can't really be done from the `flowable::Subscriber` interface.

alexmalyshev:
I think we should base the general throwing from `Subscriber::onNext()` case in the `SafeSubscriber/SubscriberBase` implementation that @dymk is working on.  I think it'd look something like:

```
void SafeSubscriber::onNext(T item) {
  if (auto const sub = subscription_.load()) {
    SUBSCRIBER_KEEP_SELF();
    try {
      onNextImpl(std::move(item));
    } catch (const std::exception& exn) {
      folly::exception_wrapper ew{std::current_exception(), exn};
      sub->cancel(); // Assuming onNextImpl() doesn't call cancel(), or cancel() is idempotent.
      onErrorImpl(std::move(ew));
    }
  }
}
```

We can add this case for `map()` in the meantime, or wait for `SafeSubscriber`.

yarpl/include/yarpl/single/SingleOperator.h
```diff
@@ -121,9 +177,13 @@ class MapOperator : public SingleOperator<U, D> {
         : OperatorSubscription(std::move(single), std::move(observer)) {}
 
     void onSuccess(U value) override {
-      auto map_operator = OperatorSubscription::getOperator();
-      OperatorSubscription::observerOnSuccess(
-          map_operator->function_(std::move(value)));
+      try {
```
phoad: Should we check "isTerminated()" in here?

yarpl/include/yarpl/observable/ObservableOperator.h
```diff
@@ -199,8 +199,13 @@ class MapOperator : public ObservableOperator<U, D, MapOperator<U, D, F>> {
         : SuperSub(std::move(observable), std::move(observer)) {}
 
     void onNext(U value) override {
-      auto& map = SuperSub::getObservableOperator();
-      SuperSub::observerOnNext(map->function_(std::move(value)));
+      try {
```
phoad: Should we check `isTerminated` in here? As, even though previous onNext might have thrown an exception, it doesn't force onNext to be not called again, isn't it?

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -179,8 +179,8 @@ class MapOperator : public FlowableOperator<U, D, MapOperator<U, D, F>> {
       : Super(std::move(upstream)), function_(std::move(function)) {}
 
   void subscribe(Reference<Subscriber<D>> subscriber) override {
-    Super::upstream_->subscribe(
-        make_ref<Subscription>(this->ref_from_this(this), std::move(subscriber)));
+    Super::upstream_->subscribe(make_ref<Subscription>(
```
phoad: In some places, we only pass `this->ref_from_this(this)` without calling make_ref<Subscription>. 
i.e.: RequestResponseRequester::subscribe(..) method
Why do we need to call make_ref<Subscription>(.. ?

yarpl/include/yarpl/observable/ObservableOperator.h
lehecka: should we put this try..catch into observerOnNext?

yarpl/include/yarpl/single/SingleOperator.h
```diff
@@ -46,30 +47,84 @@ class SingleOperator : public Single<D> {
...
 
-   private:
+    void terminateErr(folly::exception_wrapper ew) {
+      terminateImpl(TerminateState::Both(), std::move(ew));
+    }
+
+    // SingleSubscription.
+
+    void cancel() override {
+      folly::exception_wrapper ew{std::runtime_error{"Cancelled"}};
```
lehecka: why to create the exception_wrapper when you call temrinate(Up()) which only cancels the subscription but never onErrors?

alexmalyshev: Hmm yeah, good point.  I think I did that because I needed to pass *something* to the `folly::Try<D>` argument.

I can pass an empty `Try` instead.

yarpl/include/yarpl/flowable/FlowableOperator.h
alexmalyshev: In this instance, `this` refers to a `Flowable`.  We're making a new `Subscription+Subscriber` object here with `make_ref` to pass to the upstream `Flowable` because we're creating a subscription.

`RequestResponderRequester` is different because it's not really a `Flowable`.  It has subscription-level lifetime for an RSocket stream.

yarpl/include/yarpl/observable/ObservableOperator.h
alexmalyshev: All `observerOnNext()` does is call `downstreamObserver_->onNext()`, which has nothing to do with the `F` function passed in `map(F)`.

alexmalyshev: Hmm yeah.  That's an issue with all the `FooOperator` classes in general.  They don't defend against getting follow-up `onNext()` callbacks after sending a `cancel()` to the upstream.

I think it's fine to wait for `SafeSubscriber/SubscriberBase` to deal with that, instead of duplicating the effort here.

yarpl/include/yarpl/flowable/FlowableOperator.h
phoad: I did not need any change in the parent class, Subscriber. The changes in here solved an issue in my tests, that is: "throw at onNext call".
D5952697 - the unit tests with name InputThrows and ResultThrows were not passing before adding this try/catch.

PR #770, opened 2017-09-27
add counters

ragansa:
Got it!

### DIFF
--- a/rsocket/RSocketServer.cpp
+++ b/rsocket/RSocketServer.cpp
@@ -160,7 +160,6 @@ void RSocketServer::onRSocketResume(
   auto result = serviceHandler->onResume(resumeParams.token);
   if (result.hasError()) {
     VLOG(3) << "Terminating RESUME attempt from client.  No ServerState found";
-    result.value()->rSocketStateMachine_->stats().resumeFailure();
     throw result.error();
   }
   auto serverState = std::move(result.value());
--- a/rsocket/RSocketServer.cpp
+++ b/rsocket/RSocketServer.cpp
@@ -160,6 +160,7 @@ void RSocketServer::onRSocketResume(
   auto result = serviceHandler->onResume(resumeParams.token);
   if (result.hasError()) {
     VLOG(3) << "Terminating RESUME attempt from client.  No ServerState found";
+    result.value()->rSocketStateMachine_->stats().resumeFailure();
     throw result.error();
   }
   auto serverState = std::move(result.value());
--- a/rsocket/statemachine/RSocketStateMachine.cpp
+++ b/rsocket/statemachine/RSocketStateMachine.cpp
@@ -75,10 +75,11 @@ void RSocketStateMachine::setResumable(bool resumable) {
 }
 
 void RSocketStateMachine::connectServer(
-    yarpl::Reference<FrameTransport> transport,
-    const SetupParameters& params) {
-  setResumable(params.resumable);
-  connect(std::move(transport), params.protocolVersion);
+    yarpl::Reference<FrameTransport> frameTransport,
+    const SetupParameters& setupParams) {
+  stats_->socketConnected();
+  setResumable(setupParams.resumable);
+  connect(std::move(frameTransport), setupParams.protocolVersion);
   sendPendingFrames();
 }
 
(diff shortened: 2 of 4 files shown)


rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -91,9 +92,11 @@ bool RSocketStateMachine::resumeServer(
             << "Terminating new connection (" << frameTransport.get() << ")";
     frameTransport->closeWithError(folly::exception_wrapper(
         std::runtime_error("Old connection still exists")));
+    stats_->resumeFailure();
     return false;
   }
   CHECK(connect(std::move(frameTransport), resumeParams.protocolVersion));
+  stats_->resumeSuccess();
```
phoad: I believe the bool value that is returned from the next line decides it the resumption is successful or not. So, we need to send resumeSuccess if the result of the next line is `true`.

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -91,9 +92,11 @@ bool RSocketStateMachine::resumeServer(
             << "Terminating new connection (" << frameTransport.get() << ")";
     frameTransport->closeWithError(folly::exception_wrapper(
         std::runtime_error("Old connection still exists")));
+    stats_->resumeFailure();
```
phoad: It looks like the resumption can fail with a few reasons. I am not sure we need to for v1, but it might be beneficial to be able to differentiate the reason of the failure. @somasun ?

rsocket/statemachine/RSocketStateMachine.cpp
phoad: We might need one more `resumeFailure()` call at the inside of resumeFromPositionOrClose function.

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -77,6 +77,7 @@ void RSocketStateMachine::setResumable(bool resumable) {
 void RSocketStateMachine::connectServer(
     yarpl::Reference<FrameTransport> frameTransport,
     const SetupParameters& setupParams) {
+  stats_->socketConnected();
```
somasun: It might be better to put this inside connect() function call.  So it will cover all cases -  client/server and reconnects/resumes

rsocket/statemachine/RSocketStateMachine.cpp
somasun: resumeFailure() or resumeSuccess() can be easily interpreted in RSocketServer.cpp.  Inside onRSocketResume() 
- if serverState->rSocketStateMachine_->resumeServer returns false or if result.hasError() then log resumeFailure()
- if serverState->rSocketStateMachine_->resumeServer returns true, then log resumeSuccess()

Similarly for the client side (RSocketClient.cpp), inside RSocketClient::resume()::ResumeCallback 
- onResumeOk() should log resumeSuccess()
- onResumeError() should log resumeFailure()

rsocket/statemachine/RSocketStateMachine.cpp
somasun: Right, we can just add resumeFailure() to get things running.  For v2 we can go for more details.

rsocket/statemachine/RSocketStateMachine.cpp
lexs: Consider making the callbacks contain more information, I actually already did this in D5147228 (see stacked diffs for usage).

lexs:
For some reason my comment is super hard to find, here it is again:
> Consider making the callbacks contain more information, I actually already did this in D5147228 (see stacked diffs for usage).

ragansa:
@lexs (y)
That is weird, I didn't get an email the first time either.

ragansa:
@somasun We don't have a resumeCache_ for:
```
resumeCache_->impliedPosition() ;
resumeCache_->position() ;
resumeCache_->lastResetPosition();
```
Can we get to these values from somewhere?

rsocket/RSocketServer.cpp
```diff
@@ -160,6 +160,7 @@ void RSocketServer::onRSocketResume(
   auto result = serviceHandler->onResume(resumeParams.token);
   if (result.hasError()) {
     VLOG(3) << "Terminating RESUME attempt from client.  No ServerState found";
+    result.value()->rSocketStateMachine_->stats().resumeFailure();
```
ragansa: @somasun Also, do we want to identify this kind of failure in a different way? (another function? Or a variant of the other resume?)

phoad: Should we provide the Stats object to RSocketServerHandler too?
Or we can just provide it to RSocketServer already.

somasun:
Rerunning the tests to make sure it passes on latest code.  The changes look okay otherwise.

PR #771, opened 2017-09-28
Clean up a bit of RSocketStateMachine's API

alexmalyshev:
* Rename connectClientSendSetup() to connectClient() and tryClientResume() to
  resumeClient(), to match the existing connectServer() and resumeServer().

* Rename isDisconnectedOrClosed() to isDisconnected().  A closed connection
  is disconnected after all.

* Remove the remoteResumeable_ field.  It's exactly the same as the isResumable_
  field.

* Remove uses of CHECK around the private connect() helper.  Throw exceptions
  instead.

* Rename requestFireAndForget() to fireAndForget() to match
  metadataPush().

* Shrink the RSocketMode field to 1 byte, make it const.

* Move a ton of public helper methods to private, as they're not used outside of
  RSocketStateMachine.

* Header comments all around.

### DIFF
--- a/rsocket/RSocketClient.cpp
+++ b/rsocket/RSocketClient.cpp
@@ -107,7 +107,7 @@ folly::Future<folly::Unit> RSocketClient::resume() {
         createState(connection.eventBase);
       }
 
-      stateMachine_->tryClientResume(
+      stateMachine_->resumeClient(
           token_,
           std::move(frameTransport),
           std::move(resumeCallback),
@@ -153,7 +153,7 @@ void RSocketClient::fromConnection(
     framedConnection = std::make_unique<FramedDuplexConnection>(
         std::move(connection), setupParameters.protocolVersion);
   }
-  stateMachine_->connectClientSendSetup(
+  stateMachine_->connectClient(
       std::move(framedConnection), std::move(setupParameters));
 }
 
--- a/rsocket/RSocketRequester.cpp
+++ b/rsocket/RSocketRequester.cpp
@@ -142,7 +142,7 @@ yarpl::Reference<yarpl::single::Single<void>> RSocketRequester::fireAndForget(
     ]() mutable {
       // TODO pass in SingleSubscriber for underlying layers to
       // call onSuccess/onError once put on network
-      srs->requestFireAndForget(std::move(request));
+      srs->fireAndForget(std::move(request));
       // right now just immediately call onSuccess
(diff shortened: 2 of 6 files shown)


somasun:
This is a nice set of changes.  I have a comment on
> Remove the remoteResumeable_ field. It's exactly the same as the isResumable_
field.

You are right - the current usage in RSocketStateMachine is same for isResumable_ and remoteResumable_.  I am not sure about the intention of having the two, but I definitely see some applications.   
(1) Server is not resumable and client is resumable.  Server would want to reject resumption attempts from client by checking resumable_ field.  RSocket v0.1 KEEPALIVE frame's serialization depends on whether resumability is enabled or not.   So the keepAlives originating from server will have different serialization depending on whether the client has resumability (remoteResumable_) enabled (I think this is clumsy and I like v1.0 way more).
(2) Client is not resumable, but server is resumable.  Client would need to remember that server is resumable while sending KEEPALIVE frames in v0.1 protocol.

I like your changes on resumability field (except for this weird corner case).  
If there is a guarantee that we will not be using v0.1 with this code, then I am happy with the change.

alexmalyshev:
> (1) Server is not resumable and client is resumable. Server would want to reject resumption attempts from client by checking resumable_ field. RSocket v0.1 KEEPALIVE frame's serialization depends on whether resumability is enabled or not. So the keepAlives originating from server will have different serialization depending on whether the client has resumability (remoteResumable_) enabled (I think this is clumsy and I like v1.0 way more).
(2) Client is not resumable, but server is resumable. Client would need to remember that server is resumable while sending KEEPALIVE frames in v0.1 protocol.

That sounds like the code in master is bugged for the v0.1 protocol then, and my PR doesn't make it any worse.  Would you prefer me fix it in this PR, or a followup?

somasun:
> That sounds like the code in master is bugged for the v0.1 protocol then, and my PR doesn't make it any worse. Would you prefer me fix it in this PR, or a followup?

Yes, I was stating the mere intention of those two fields.  I am okay with either approaches (whatever is convenient for you).  Ultimately, I would like us to stop using v0.1 sooner :).  Then the code could be simpler.

alexmalyshev:
> I am okay with either approaches (whatever is convenient for you). Ultimately, I would like us to stop using v0.1 sooner :). Then the code could be simpler.

Well then my preference would be to land this as is, and open an issue on the v0.1 bug.  We can fix it later if we need to, otherwise we'll wait for v0.1 to die out and we'll just close out the issue.

PR #775, opened 2017-09-29
Don't crash if user defined function throws at onNext

phoad:
We have two common ways to subscribe to a flowable. One is implementing a `Subscriber` and providing this to the `subscribe` method. This is kind of a pro way of consuming the `Flowable`.

Another easy to use, common usage, case is just providing a `lambda` function that consumes the data. When this way is selected, if the given `lambda` function throws, it causes the whole application to crash.

In this update, I have encapsulated the user provided function with try/catch, so we will not need to do any more changes in the base Subscriber class at all and still be able to provide safety for the simple usage.

Note: instead of calling `onError` method of the parent class, I have taken the `onError` method of the Subscriber to Base class via `using`. So when this function is overriden, it will not call the parent class, but the method which has overrided it. So it will call the `error` lambda that the user has provided.

### DIFF
--- a/yarpl/include/yarpl/flowable/Subscribers.h
+++ b/yarpl/include/yarpl/flowable/Subscribers.h
@@ -76,13 +76,27 @@ class Subscribers {
     }
 
     void onNext(T value) override {
-      next_(std::move(value));
+      if (userError_) {
+        return;
+      }
+      try {
+        next_(std::move(value));
+      } catch (const std::exception& exn) {
+        userError_ = true;
+        onError(folly::exception_wrapper{std::current_exception(), exn});
+        return;
+      }
+
       if (--pending_ < batch_ / 2) {
         const auto delta = batch_ - pending_;
         pending_ += delta;
         Subscriber<T>::subscription()->request(delta);
       }
     }
+
+   protected:
+    using Subscriber<T>::onError;
+    bool userError_{false};
 
    private:
     Next next_;
@@ -120,11 +134,14 @@ class Subscribers {
           complete_(std::move(complete)) {}
 
     void onComplete() {
-      Subscriber<T>::onComplete();
-      complete_();
+      if (!userError_) { // already errored?
+        Subscriber<T>::onComplete();
+        complete_();
+      }
     }
 
    private:
+    using Base<T, Next>::userError_;
     Complete complete_;
   };
 
--- a/yarpl/include/yarpl/flowable/Subscribers.h
+++ b/yarpl/include/yarpl/flowable/Subscribers.h
@@ -83,6 +83,7 @@ class Subscribers {
         next_(std::move(value));
       } catch (const std::exception& exn) {
         userError_ = true;
+        Subscriber<T>::subscription()->cancel();
         onError(folly::exception_wrapper{std::current_exception(), exn});
         return;
       }
--- a/yarpl/include/yarpl/flowable/Subscribers.h
+++ b/yarpl/include/yarpl/flowable/Subscribers.h
@@ -113,7 +113,13 @@ class Subscribers {
 
     void onError(folly::exception_wrapper error) override {
       Subscriber<T>::onError(error);
-      error_(std::move(error));
+      try {
+        error_(std::move(error));
+      } catch (const std::exception& exn) {
+        auto ew = folly::exception_wrapper{std::current_exception(), exn};
+        DLOG(FATAL) << "'error' method should not throw: " << ew.what();
+        LOG(ERROR) << "'error' method should not throw: " << ew.what();
+      }
     }
 
    private:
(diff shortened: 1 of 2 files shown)


yarpl/include/yarpl/flowable/Subscribers.h
```diff
@@ -120,11 +134,14 @@ class Subscribers {
           complete_(std::move(complete)) {}
 
     void onComplete() {
-      Subscriber<T>::onComplete();
-      complete_();
+      if (!userError_) { // already errored?
+        Subscriber<T>::onComplete();
```
alexmalyshev: You can't call `onComplete()` if you've already called `onError()`.

yarpl/include/yarpl/flowable/Subscribers.h
```diff
@@ -76,14 +76,28 @@ class Subscribers {
     }
 
     void onNext(T value) override {
-      next_(std::move(value));
+      if (userError_) {
```
alexmalyshev: You don't need `userError_`.  You can check if `Subscriber<T>::subscription()` is null or not.

yarpl/include/yarpl/flowable/Subscribers.h
phoad: I think the `!userError` check does that. The unit test verifies that.

yarpl/include/yarpl/flowable/Subscribers.h
phoad: If the user provides a Subscriber, then I believe that they need to write a Subscriber that handles this issue, at least for v1.
If the user only provides a lambda, then we should handle this issue.

yarpl/include/yarpl/flowable/Subscribers.h
```diff
@@ -76,14 +76,28 @@ class Subscribers {
...
+        return;
+      }
+      try {
+        next_(std::move(value));
+      } catch (const std::exception& exn) {
+        userError_ = true;
+        onError(folly::exception_wrapper{std::current_exception(), exn});
```
alexmalyshev: It seems like you'd want to `cancel()` the subscription here, otherwise the publisher is going to keep sending you `onNext()` even though the subscriber has already been terminated.

yarpl/include/yarpl/flowable/Subscribers.h
phoad: I want to be explicit about the `userError_` in here. 
The subscription() might be null for various reasons. I want to be able to differentiate this specific case. Because of this I decided to keep that variable.

phoad:
I have extended the update a bit to protect our server/client from crashes if user provided code throws for 'onerror' or 'oncomplete', I placed DLOG(FATAL) and LOG(ERROR).
We should not crash in prod but log error for these cases and we should crash for debug build.
Please advise if we need to do something but not this.

yarpl/test/FlowableTest.cpp
```diff
@@ -593,5 +593,20 @@ TEST(FlowableTest, SubscribeMultipleTimes) {
   EXPECT_EQ(results[4], std::vector<int64_t>({1, 2, 3, 4, 5}));
 }
 
+TEST(FlowableTest, ConsumerThrows_OnNext) {
```
lehecka: can we add similar test for onNext and onComplete with throwing exception?

yarpl/include/yarpl/flowable/Subscribers.h
alexmalyshev: In theory, it should only be nullptr if the Subscriber hasn't yet called `onSubscribe()` (in which you wouldn't be calling these other onFoo() methods), or if the `Subscriber` is terminated, which gives you exactly what you want.

yarpl/test/FlowableTest.cpp
phoad: let me do that.

yarpl/include/yarpl/flowable/Subscribers.h
phoad: ok.

yarpl/include/yarpl/flowable/Subscribers.h
```diff
@@ -98,7 +113,13 @@ class Subscribers {
 
     void onError(folly::exception_wrapper error) override {
       Subscriber<T>::onError(error);
-      error_(std::move(error));
+      try {
+        error_(std::move(error));
+      } catch (const std::exception& exn) {
+        auto ew = folly::exception_wrapper{std::current_exception(), exn};
+        DLOG(FATAL) << "'error' method should not throw: " << ew.what();
+        LOG(ERROR) << "'error' method should not throw: " << ew.what();
```
phoad: I think I don't need the LOG(ERROR)..
""DFATAL - logs FATAL error in debug mode (no NDEBUG macro defined)
  - reduces ERROR in production""

PR #778, opened 2017-09-29
Simplify StreamsWriter a little

alexmalyshev:
Most of the methods on it are equivalent to
RSocketStateMachine::outputFrameOrEnqueue().  So do that.

The main gnarly one that remains is StreamsWriter::writeNewStream(), which does
more than just send a frame.

### DIFF
--- a/rsocket/statemachine/RSocketStateMachine.cpp
+++ b/rsocket/statemachine/RSocketStateMachine.cpp
@@ -953,53 +953,20 @@ void RSocketStateMachine::writeNewStream(
   }
 }
 
-void RSocketStateMachine::writeRequestN(StreamId streamId, uint32_t n) {
-  outputFrameOrEnqueue(Frame_REQUEST_N(streamId, n));
-}
-
-void RSocketStateMachine::writePayload(
-    StreamId streamId,
-    Payload payload,
-    bool complete) {
-  Frame_PAYLOAD frame(
-      streamId,
-      FrameFlags::NEXT | (complete ? FrameFlags::COMPLETE : FrameFlags::EMPTY),
-      std::move(payload));
+void RSocketStateMachine::writeRequestN(Frame_REQUEST_N&& frame) {
   outputFrameOrEnqueue(std::move(frame));
 }
 
-void RSocketStateMachine::writeCloseStream(
-    StreamId streamId,
-    StreamCompletionSignal signal,
-    std::string message) {
-  switch (signal) {
-    case StreamCompletionSignal::COMPLETE:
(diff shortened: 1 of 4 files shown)


rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -924,53 +924,20 @@ void RSocketStateMachine::writeNewStream(
   }
 }
 
-void RSocketStateMachine::writeRequestN(StreamId streamId, uint32_t n) {
-  outputFrameOrEnqueue(Frame_REQUEST_N(streamId, n));
+void RSocketStateMachine::writeRequestN(Frame_REQUEST_N&& frame) {
```
phoad: Why do we need these specific function? We already call `outputFrameOrEnqueue(std::move(frame))` for each and every one of them. So why don't we just remove them all?

rsocket/statemachine/StreamsWriter.h
```diff
@@ -24,18 +19,12 @@ class StreamsWriter {
       Payload payload,
       bool TEMP_completed) = 0;
 
-  virtual void writeRequestN(StreamId streamId, uint32_t n) = 0;
-
-  virtual void
-  writePayload(StreamId streamId, Payload payload, bool complete) = 0;
+  virtual void writeRequestN(Frame_REQUEST_N&&) = 0;
```
phoad: Let's simplify it further as renaming this method as `write`?
```
virtual void write(std::unique_ptr<folly::IOBuf> frame)
```
Why do we care about what do we write?

rsocket/statemachine/RSocketStateMachine.cpp
alexmalyshev: I didn't want to expose `outputFrameOrEnqueue()` just yet, as I'm not sure if that's desirable.  Writing a `REQUEST_N` frame just sends the frame, but writing a `REQUEST_STREAM` frame has to do more bookkeeping within the `RSocketStateMachine`.  It'd be confusing to see why some callers did `outputFrameOrEnqueue(Frame_REQUEST_N{})` and some did `writeNewStream(...)`.

For now I'd rather keep all these methods as separate, to reduce confusion.  I agree that having one function simply "wrapping" another isn't great though.

rsocket/statemachine/StreamsWriter.h
alexmalyshev: See my other comment, but there's different behavior for writing different frames (e.g. `REQUEST_N` versus `REQUEST_STREAM`).

PR #779, opened 2017-09-30
add stats to RSocketServer

ragansa:
- any other things we ought to count?

### DIFF
--- a/rsocket/RSocketServer.cpp
+++ b/rsocket/RSocketServer.cpp
@@ -162,6 +162,7 @@ void RSocketServer::onRSocketResume(
     ResumeParameters resumeParams) {
   auto result = serviceHandler->onResume(resumeParams.token);
   if (result.hasError()) {
+    stats_->resumeFailedNoState();
     VLOG(3) << "Terminating RESUME attempt from client.  No ServerState found";
     throw result.error();
   }
--- a/rsocket/RSocketServer.cpp
+++ b/rsocket/RSocketServer.cpp
@@ -13,7 +13,8 @@
 namespace rsocket {
 
 RSocketServer::RSocketServer(
-    std::unique_ptr<ConnectionAcceptor> connectionAcceptor)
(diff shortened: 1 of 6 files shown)


rsocket/RSocketServer.cpp
```diff
@@ -82,6 +84,7 @@ void RSocketServer::acceptConnection(
     std::unique_ptr<DuplexConnection> connection,
     folly::EventBase&,
     std::shared_ptr<RSocketServiceHandler> serviceHandler) {
+  stats_->serverConnectionAccepted();
```
phoad: Should we place this line at the inside of onRSocketSetup function, or onRSocketResume?

rsocket/RSocketStats.h
```diff
@@ -38,7 +40,7 @@ class RSocketStats {
   virtual void streamBufferChanged(
       int64_t framesCountDelta,
       int64_t dataSizeDelta) = 0;
-
+  virtual void resumeFailedNoState() = 0;
```
phoad: nit: we might rename this function as resumeFailed() and create an enum of possible errors and pass it in this function. Or we might pass in a folly::exception_wrapper into this function. So we can let the party which implements this function to be able to know which kind of error has happened.

rsocket/RSocketServer.cpp
lehecka: I think this particular event is really just about getting an instance of DuplexConnection, the next step is counters in onSetup, onResume, error, ... which we hopefully have. 
I would leave this where it is.

rsocket/RSocketStats.h
lehecka: once we have more than one case of resumeFailed, I would go with Fuat's suggestion. Right now I can see only once  of resumeFailed.

somasun:
PR looks good.  Just rerunning the tests on latests changes.

PR #781, opened 2017-10-02
Allow client stateMachine to run on separate EventBase

somasun:
This PR allows clients (new/warm-resumed/cold-resumed) to run the stateMachine and transport in different EventBases.

Added normal connect/disconnect, warm-resumed-client and cold-resumed-client tests.

### DIFF
--- a/rsocket/RSocket.cpp
+++ b/rsocket/RSocket.cpp
@@ -12,17 +12,19 @@ folly::Future<std::unique_ptr<RSocketClient>> RSocket::createConnectedClient(
     std::shared_ptr<RSocketStats> stats,
     std::shared_ptr<RSocketConnectionEvents> connectionEvents,
     std::shared_ptr<ResumeManager> resumeManager,
-    std::shared_ptr<ColdResumeHandler> coldResumeHandler) {
+    std::shared_ptr<ColdResumeHandler> coldResumeHandler,
+    folly::EventBase* stateMachineEvb) {
   auto createRSC = [
-      connectionFactory,
-      setupParameters = std::move(setupParameters),
-      responder = std::move(responder),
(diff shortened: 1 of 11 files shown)


rsocket/RSocketClient.cpp
```diff
@@ -44,25 +45,30 @@ const std::shared_ptr<RSocketRequester>& RSocketClient::getRequester() const {
   return requester_;
 }
 
-folly::Future<folly::Unit> RSocketClient::resume() {
+folly::Future<folly::Unit> RSocketClient::resume(folly::EventBase* smEvb) {
```
lexs: Why does this take the EventBase, wouldn't it just be passed in the constructor?

rsocket/RSocket.cpp
```diff
@@ -12,17 +12,19 @@ folly::Future<std::unique_ptr<RSocketClient>> RSocket::createConnectedClient(
     std::shared_ptr<RSocketStats> stats,
     std::shared_ptr<RSocketConnectionEvents> connectionEvents,
     std::shared_ptr<ResumeManager> resumeManager,
-    std::shared_ptr<ColdResumeHandler> coldResumeHandler) {
+    std::shared_ptr<ColdResumeHandler> coldResumeHandler,
+    folly::EventBase* stateMachineEvb) {
```
lexs: I'd vote for having this be the first arg, not the last.

somasun: I believe most common usage of RSocketClient will be simple, where the user lets the RSocketClient::StateMachine use the Transport's EventBase.   I suspect only in sophisticated scenarios the user would want to explicitly control the StateMachine's EventBase.

PR #782, opened 2017-10-02
remove InternalSubscriber

dymk:
What's on the tin

Has instances which rely on a `subscrption_` use BaseSubscriber

### DIFF
--- a/rsocket/DuplexConnection.h
+++ b/rsocket/DuplexConnection.h
@@ -12,6 +12,31 @@ class IOBuf;
 
 namespace rsocket {
 
+using yarpl::Reference;
+
+class DuplexSubscriber :
+  public yarpl::flowable::Subscriber<std::unique_ptr<folly::IOBuf>>
+{
+public:
+  void onSubscribe(Reference<yarpl::flowable::Subscription> sub) override {
+    subscription_ = sub;
+  }
+  void onComplete() override {
+    subscription_.reset();
+  }
+  void onError(folly::exception_wrapper) override {
+    subscription_.reset();
+  }
+
+protected:
+  Reference<yarpl::flowable::Subscription> subscription() {
+    return subscription_;
+  }
+
+private:
+  Reference<yarpl::flowable::Subscription> subscription_;
+};
+
 /// Represents a connection of the underlying protocol, on top of which the
 /// RSocket protocol is layered.  The underlying protocol MUST provide an
 /// ordered, guaranteed, bidirectional transport of frames.  Moreover, frame
@@ -29,9 +54,9 @@ namespace rsocket {
 /// before the connection is destroyed.
 class DuplexConnection {
  public:
-  using InternalSubscriber =
-      yarpl::flowable::InternalSubscriber<std::unique_ptr<folly::IOBuf>>;
-  using Subscriber = yarpl::flowable::Subscriber<std::unique_ptr<folly::IOBuf>>;
+  using Subscriber =
+      yarpl::flowable::Subscriber<std::unique_ptr<folly::IOBuf>>;
+  using DuplexSubscriber = rsocket::DuplexSubscriber;
 
   virtual ~DuplexConnection() = default;
 
--- a/yarpl/include/yarpl/flowable/CancelingSubscriber.h
+++ b/yarpl/include/yarpl/flowable/CancelingSubscriber.h
@@ -13,15 +13,20 @@ namespace flowable {
  * A Subscriber that always cancels the subscription passed to it.
  */
 template <typename T>
-class CancelingSubscriber final : public InternalSubscriber<T> {
+class CancelingSubscriber final : public BaseSubscriber<T> {
  public:
-  void onSubscribe(
-      yarpl::Reference<yarpl::flowable::Subscription> sub) override {
-    sub->cancel();
+  void onSubscribeImpl() override {
+    this->cancel();
   }
 
-  void onNext(T) override {
+  void onNextImpl(T) override {
     throw std::logic_error{"CancelingSubscriber::onNext() can never be called"};
+  }
+  void onCompleteImpl() override {
+    throw std::logic_error{"CancelingSubscriber::onComplete() can never be called"};
+  }
+  void onErrorImpl(folly::exception_wrapper) override {
+    throw std::logic_error{"CancelingSubscriber::onError() can never be called"};
   }
 };
 }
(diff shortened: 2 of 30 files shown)


rsocket/DuplexConnection.h
```diff
@@ -29,9 +54,9 @@ namespace rsocket {
 /// before the connection is destroyed.
 class DuplexConnection {
  public:
-  using InternalSubscriber =
-      yarpl::flowable::InternalSubscriber<std::unique_ptr<folly::IOBuf>>;
-  using Subscriber = yarpl::flowable::Subscriber<std::unique_ptr<folly::IOBuf>>;
+  using Subscriber =
+      yarpl::flowable::Subscriber<std::unique_ptr<folly::IOBuf>>;
+  using DuplexSubscriber = rsocket::DuplexSubscriber;
```
lehecka: can we still call it DuplexConnection::Subscriber? I think DuplexSubscriber is confusing..

yarpl/include/yarpl/flowable/CancelingSubscriber.h
```diff
@@ -13,16 +13,21 @@ namespace flowable {
  * A Subscriber that always cancels the subscription passed to it.
  */
 template <typename T>
-class CancelingSubscriber final : public InternalSubscriber<T> {
+class CancelingSubscriber final : public BaseSubscriber<T> {
```
lehecka: Why is CancelingSubscriber deriving from the BaseSubscriber? Do we need any of the functionality?

yarpl/include/yarpl/flowable/CancelingSubscriber.h
```diff
@@ -13,16 +13,21 @@ namespace flowable {
...
  public:
-  void onSubscribe(
-      yarpl::Reference<yarpl::flowable::Subscription> sub) override {
-    sub->cancel();
+  void onSubscribeImpl() override {
+    this->cancel();
   }
 
-  void onNext(T) override {
+  void onNextImpl(T) override {
```
phoad: Would it be possible to make these private or protected?

rsocket/DuplexConnection.h
dymk: I don't think there's a clean way to do that, the DuplexConnection deals with the generic Subscriber interface in setInput/getOutput, and DuplexSubscriber is the old `InternalSubscriber` analog. It just so happened that the Subscriber interface previously implemented InternalSubscriber.

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -21,40 +21,6 @@ class Subscriber : public virtual Refcounted, public yarpl::enable_get_ref {
   virtual void onNext(T) = 0;
 };
 
-// codemod all things that inherit from Subscriber to inherit from
```
phoad: Is it possible to make the *impl(..) methods protected?
No one should call them but the classes can override them.

phoad: Furthermore, why don't we just provide empty implementation for them?

yarpl/include/yarpl/flowable/Subscribers.h
```diff
@@ -64,38 +64,36 @@ class Subscribers {
...
         const auto delta = batch_ - pending_;
         pending_ += delta;
-        InternalSubscriber<T>::subscription()->request(delta);
+        this->request(delta);
       }
     }
 
-   protected:
-    using InternalSubscriber<T>::onError;
+    void onCompleteImpl() override {}
```
phoad: these should be protected, and we will not need them if we implemented them this way in the parent class.

yarpl/test/FlowableTest.cpp
```diff
@@ -593,6 +586,22 @@ TEST(FlowableTest, SubscribeMultipleTimes) {
   EXPECT_EQ(results[4], std::vector<int64_t>({1, 2, 3, 4, 5}));
 }
 
+/* following test should probably behave like:
```
phoad: ?

yarpl/include/yarpl/flowable/CancelingSubscriber.h
dymk: It forces CancelingSubscriber to obey the semantics of "onNext will never be called after cancel()" that we've adopted for BaseSubscriber. It doesn't *need* to do this, but there's not much reason not to.

yarpl/test/ReferenceTest.cpp
```diff
@@ -11,14 +11,17 @@
...
-using yarpl::flowable::InternalSubscriber;
 using yarpl::flowable::Subscriber;
+using yarpl::flowable::BaseSubscriber;
 
 namespace {
 
 template <class T>
-class MySubscriber : public InternalSubscriber<T> {
-  void onNext(T) override {}
+class MySubscriber : public BaseSubscriber<T> {
```
phoad: Again, I just suggest we to have the empty implementations in the base.

yarpl/include/yarpl/flowable/Subscriber.h
dymk: No default implementation is provided for common methods, because explicit > implicit. The compiler won't let you forget to implement a pure virtual method.

yarpl/include/yarpl/flowable/Subscribers.h
dymk: A future diff can improve problems that aren't introduced here, let's keep this diff focused just on removing InternalSubscriber.

yarpl/test/ReferenceTest.cpp
dymk: The BaseSubscriber design intentionally leaves out a default implementation

yarpl/test/FlowableTest.cpp
dymk: We discussed how exceptions should bubble up in the meeting, sounds like we need to discuss it further though. I don't think it makes sense to swallow exceptions at the Subscriber level; that's rsocket's job. I left this note here for a future design discussion about it.

yarpl/include/yarpl/flowable/CancelingSubscriber.h
dymk: There wouldn't be much point; just cast to a BaseSubscriber (which is always going to be valid) and they're public again.

yarpl/test/FlowableTest.cpp
phoad: why do we need commented out code. can you remove it?

yarpl/include/yarpl/flowable/CancelingSubscriber.h
phoad: I mean why they are public in the BaseSubscriber at all?

PR #784, opened 2017-10-02
Fix crash in SetupResumeAcceptor::closeAll()

alexmalyshev:
It expected the currently-open subscribers not to have had onSubscribe() called.
But the subscribers can definitely be in the state of having had onSubscribe()
called, and still waiting on an on{Next,Complete,Error} to kick them out from
the set.

Add test that fails before this diff.

### DIFF
--- a/rsocket/internal/SetupResumeAcceptor.cpp
+++ b/rsocket/internal/SetupResumeAcceptor.cpp
@@ -42,62 +42,58 @@ void closeWithError(
 }
 }
 
-/// Subscriber that owns a connection, sets itself as that connection's input,
-/// and reads out a single frame before cancelling.
-class OneFrameSubscriber : public yarpl::flowable::BaseSubscriber<std::unique_ptr<folly::IOBuf>> {
- public:
-  OneFrameSubscriber(
-      SetupResumeAcceptor& acceptor,
-      std::unique_ptr<DuplexConnection> connection,
-      SetupResumeAcceptor::OnSetup onSetup,
-      SetupResumeAcceptor::OnResume onResume)
-      : acceptor_{acceptor},
-        connection_{std::move(connection)},
-        onSetup_{std::move(onSetup)},
-        onResume_{std::move(onResume)} {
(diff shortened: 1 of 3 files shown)


alexmalyshev:
Hmm so rebasing on top of master fixes the issue, but only because the new `DuplexSubscriber` silently overwrites subscriptions when calling `onSubscribe()` more than once.  I think this change is still valuable.

rsocket/internal/SetupResumeAcceptor.cpp
```diff
@@ -271,12 +279,11 @@ void SetupResumeAcceptor::closeAll() {
...
 
-  std::runtime_error exn{"SetupResumeAcceptor is shutting down"};
-
   auto connections = std::move(connections_);
   for (auto& connection : connections) {
-    connection->onSubscribe(yarpl::flowable::Subscription::empty());
-    connection->onError(exn);
+    // Nasty, but we can't type the set as holding OneFrameSubscriber objects
+    // without exposing the OneFrameSubscriber definition in the header.
+    static_cast<OneFrameSubscriber*>(connection.get())->close();
```
dymk: Expose it as `detail::OneFrameSubscriber` so we don't need to do any nasty casting, but make it clear this is a private implementation detail? This will eventually break when the type of `connections` changes at some point.

alexmalyshev: Can't make it `detail::OneFrameSubscriber`.  There's a cyclical dependency between `SetupResumeAcceptor` and `OneFrameSubscriber` because `OFS` uses `SRA::OnSetup` and `SRA::OnResume`.

I'll look into making it a nested class instead.

somasun:
Tried the patch, it fixes all the crashes I was seeing in DifferentEvb test (y)

alexmalyshev:
I'll wait for https://github.com/rsocket/rsocket-cpp/pull/792 to land, as that'll definitely conflict with this.

PR #787, opened 2017-10-03
rename AllowanceSemaphore => Allowance

dymk:
and make method names reflect what methods actually do. Some like "release" don't make sense, others like "acquire" are just not the most clear.


rsocket/internal/Allowance.h
```diff
@@ -8,23 +8,23 @@
...
 
-  AllowanceSemaphore() = default;
+  Allowance() = default;
 
-  explicit AllowanceSemaphore(ValueType initialValue) : value_(initialValue) {}
+  explicit Allowance(ValueType initialValue) : value_(initialValue) {}
 
-  bool tryAcquire(ValueType n = 1) {
-    if (!canAcquire(n)) {
+  bool tryConsume(ValueType n) {
```
alexmalyshev: Do we ever really want to do `tryConsume(n)` where `n > 1`?  Seems like we don't other than tests...

I'd figure we *have* to always consume by 1, since we're calling `onNext()` one-by-one, and any of them can cancel the subscription at any given time.

test/internal/AllowanceTest.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include "rsocket/internal/Allowance.h"
+
+using namespace ::testing;
+using namespace ::rsocket;
+
+TEST(AllowanceTest, Finite) {
+  Allowance sem;
```
alexmalyshev: Maybe rename the `sem` variables in this file too.

dymk: (y)

rsocket/internal/Allowance.h
dymk: If that's the case, maybe this method should be `consumeOne()` and tests fixed to expect only one allowance consumed. I feel like at least this is more explicit about what's going on, rather than assuming a default value of 1.

PR #790, opened 2017-10-03
wait for server to shutdown before destructing eventbases

dymk:
was causing a use-after-free when a destructor triggered in transportWorker would access the destroyed evb in SMWorker.


dymk:
I think the issue might be more fundamental than just waiting for the server to shut down. Looks like ScheduledFrameTransport/Processor hold dangling pointers to the statemachine eventbase, which shouldn't happen.

PR #792, opened 2017-10-03
SetupResumeAcceptor simplifications & refactorings 

dymk:
- `owner_` looks like it wasn't being used, remove it 
 - add checks to verify that methods being called in the correct eventbase
 - don't accept new connections if the `SetupResumeAcceptor` has closed already
 - worker thread name was shortened, else it gets truncated

### DIFF
--- a/rsocket/internal/SetupResumeAcceptor.cpp
+++ b/rsocket/internal/SetupResumeAcceptor.cpp
@@ -44,7 +44,7 @@ void closeWithError(
 
 /// Subscriber that owns a connection, sets itself as that connection's input,
 /// and reads out a single frame before cancelling.
-class OneFrameSubscriber : public DuplexConnection::DuplexSubscriber {
+class OneFrameSubscriber : public yarpl::flowable::BaseSubscriber<std::unique_ptr<folly::IOBuf>> {
  public:
   OneFrameSubscriber(
       SetupResumeAcceptor& acceptor,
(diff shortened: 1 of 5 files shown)


alexmalyshev:
`owner_` is actually used.  It allows `SetupResumeAcceptor` to continue running if the EventBase backing it has died.  Of course, it's questionable if that's desirable in the first place, but this did fix bugs in the past.

test/internal/SetupResumeAcceptorTest.cpp
```diff
@@ -256,8 +256,7 @@ TEST(SetupResumeAcceptor, EventBaseDisappear) {
   auto evb = std::make_unique<folly::EventBase>();
 
   SetupResumeAcceptor acceptor{
-      ProtocolVersion::Unknown, evb.get(), std::this_thread::get_id()};
-  evb.reset();
+      ProtocolVersion::Unknown, evb.get()};
```
alexmalyshev: Example usage of the `owner_` field in action, without it the `evb.reset()` will cause `~SetupResumeAcceptor` to crash.

dymk: It looks like `owner_` is supposed to be an optional field, but there's no usages where it's optional. Is there ever a case where we want to be able to destroy the EB before destroying the SetupResumeAcceptor? We already break if the user destroys an RSocketServer's EB before the closing it.

alexmalyshev: Yeah, the field only really fixes things for `SetupResumeAcceptor`, but `RSocketServer` is still hit by it.

We can probably remove this now after my recent `HttpConnectionAcceptor` changes in RTGW.

rsocket/internal/SetupResumeAcceptor.cpp
```diff
@@ -231,6 +224,13 @@ void SetupResumeAcceptor::accept(
     std::unique_ptr<DuplexConnection> connection,
     OnSetup onSetup,
     OnResume onResume) {
+  DCHECK(inOwnerThread());
+
+  if(closed_) {
+    DCHECK(false) << "can't accept after having closed_!";
```
alexmalyshev: I'm not sure if I'd want this as a `DCHECK`...  We shouldn't blow up if someone does `close() -> accept()`.

I think it's better to silently return, or throw an exception.

dymk: Makes sense

PR #793, opened 2017-10-04
Simplify RSocketClient StateMachine EventBase

somasun:


rsocket/RSocket.cpp
```diff
@@ -74,9 +74,10 @@ folly::Future<std::unique_ptr<RSocketClient>> RSocket::createResumedClient(
       std::move(stats),
       std::move(connectionEvents),
       std::move(resumeManager),
-      std::move(coldResumeHandler));
+      std::move(coldResumeHandler),
+      stateMachineEvb);
 
-  return c->resume(stateMachineEvb)
+  return c->resume()
```
dymk: Does `c->resume()` or `.then()` constitute a sequence point?
If not, moving `c` -> `client` might happen before the call to `c->resume()`, resulting in a nullptr deref/segfault.

somasun: c is a rawptr

PR #794, opened 2017-10-04
Clean up the KeepAlive semantics

somasun:
Addresses https://github.com/rsocket/rsocket-cpp/issues/772


rsocket/RSocket.h
```diff
@@ -19,7 +21,7 @@ class RSocket {
       std::shared_ptr<RSocketResponder> responder =
           std::make_shared<RSocketResponder>(),
       std::unique_ptr<KeepaliveTimer> keepaliveTimer =
```
lexs: Should we just change this to take `folly::Optional<std::chrono::milliseconds>`? It's unlikely we will ever have a different `KeepAlive` impl and this lets us setting the EventBase below.

lehecka:
I would suggest to make FollyKeepaliveTimer an internal class as there is no other implementation of this we ever needed. It would be created by the RSocketClient with the correct event base and passed into the RSocketStateMachine. We won't need the keepalive interface either.
We can have a new parameter for the RSocketClient ctor std::milliseconds keepaliveInterval = 5

somasun:
Would totally love to that.  I remember at some point there was a conversation about allowing application to pass in a KeepAlive implementation, so left it there assuming it is still the case.

Will update the PR.

rsocket/RSocket.cpp
```diff
@@ -18,7 +18,7 @@ folly::Future<std::unique_ptr<RSocketClient>> RSocket::createConnectedClient(
     connectionFactory,
     setupParameters = std::move(setupParameters),
     responder = std::move(responder),
-    keepaliveTimer = std::move(keepaliveTimer),
+    keepaliveInterval = std::move(keepaliveInterval),
```
alexmalyshev: Nit: All these instances of `std::move(keepaliveInternal)` don't need the move.  It's just an integer.

rsocket/RSocketClient.cpp
```diff
@@ -20,15 +20,15 @@ RSocketClient::RSocketClient(
...
+    std::chrono::milliseconds keepaliveInterval,
     std::shared_ptr<RSocketStats> stats,
     std::shared_ptr<RSocketConnectionEvents> connectionEvents,
     std::shared_ptr<ResumeManager> resumeManager,
     std::shared_ptr<ColdResumeHandler> coldResumeHandler)
     : connectionFactory_(std::move(connectionFactory)),
       connectionSet_(std::make_shared<ConnectionSet>()),
       responder_(std::move(responder)),
-      keepaliveTimer_(std::move(keepaliveTimer)),
+      keepaliveInterval_(std::move(keepaliveInterval)),
```
lexs: Do we need this variable, can we just create the timer here instead?

yschimke:
Two relevant issues for cleanup

https://github.com/rsocket/rsocket-cpp/issues/603
https://github.com/rsocket/rsocket-cpp/issues/706

Close when the cleanup happens

PR #795, opened 2017-10-04
All writes should be guarded with try/catch

phoad:
Details:  T22408574
realtime_gateway_task_crashes:trace:rsocket::FramedWriter::onNext:SIGSEGV

Because of asynchrony, there can be cases that the StreamStateMachine wants to write to the `output`, but the output is already closed, or more. 

I have added try/catch to each write operation to catch such errors and close the corresponding stream.

I will add a unit test to verify the update/fix.


alexmalyshev:
The linked task refers to a crash because of a heap-use-after-free.  There are no exceptions involved there as far as I can see.

PR #798, opened 2017-10-05
remove pure virtual from rsocketstats

ragansa:
Shall we also remove NoopStats?  Since it not doesn't do anything unique now.
Although, it's name is nice and clear about it's purpose..
Thoughts?


dymk:
Why though?

ragansa:
Every time we update, someone has to go through and add the new, unused functions to every class that's inheriting.

PR #800, opened 2017-10-14
Let limiting the max number of active streams

phoad:
In my integration of RSocket to Service Router, I need RSocketClient to inform 
 - if it is closed
 - how many streams does it currently serve
 - set/get max number of streams that can be active at a given time.


somasun:
I am not really convinced that we should have functions like `getStreamCount` and `setMaxActiveStreamCount` in RSocket implementation.   These should be primitives which are implemented on top of RSocket by the application.  It is better to keep RSocket focussed on a limited set of essential functionalities.

phoad:
You are right Mani :)

PR #803, opened 2017-10-17
Fix crash when old transport hasn't disconnected yet

somasun:
Without the change, the included test would trigger the following check in EventBase.cpp
`EventBase.cpp:211] Check failed: evbTid == curTid

Test passes fine after the change.


rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -100,8 +100,8 @@ bool RSocketStateMachine::resumeServer(
   if (!isDisconnected()) {
```
phoad: Should this be the first thing to do, namely should we immediately check if it is not disconnected and then perform the other calculations?

somasun: The calculations are for logging.  We do them irrespective of the outcome of `isDisconnected()`

PR #804, opened 2017-10-18
remove RSocketRequster::closeSocket()

dymk:
Simplifies usage and reduces confusion about how a requester can be used

Prior, it's possible to put the RSocketClient in a state such that it can't be resumed; fully destructing the client should only be possible by calling its destructor. 

Disconnecting the underlying socket should be done through `RSocketClient::disconnect()`

### DIFF
--- a/rsocket/RSocketRequester.cpp
+++ b/rsocket/RSocketRequester.cpp
@@ -23,14 +23,6 @@ RSocketRequester::~RSocketRequester() {
   VLOG(1) << "Destroying RSocketRequester";
 }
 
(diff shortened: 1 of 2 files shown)


dymk:
This is almost certainly a race between signaling methods and calls to request()/cancel(). 

`subscription_` in TestSubscriber is only reset in onError and onComplete, and the nullptr deref happens when `subscription_->cancel()` is called. 

Likely fixed by inheriting `class TestSubscriber : public BaseSubscriber<T>`

PR #805, opened 2017-10-19
Adding support for temporarily accessing the underlying transport object for TcpDuplexConnection

phoad:
I need to get a pointer to the underlying Transport object to be able to fulfill requirements of an interface.
So I have added getConnection and getTransport functions to the TcpDuplexConnection class and relevant classes. I kept the update at minimum to be able to accomplish the goal.

### DIFF
--- a/rsocket/framing/ScheduledFrameTransport.h
+++ b/rsocket/framing/ScheduledFrameTransport.h
@@ -33,7 +33,11 @@ class ScheduledFrameTransport : public FrameTransport,
   void close() override;
   void closeWithError(folly::exception_wrapper ex) override;
 
- private:
+  DuplexConnection* getConnection() override {
+    return frameTransport_ ? frameTransport_->getConnection() : nullptr;
+  }
+
+private:
   folly::EventBase* transportEvb_;
   folly::EventBase* stateMachineEvb_;
   yarpl::Reference<FrameTransport> frameTransport_;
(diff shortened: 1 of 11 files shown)


rsocket/transports/tcp/TcpDuplexConnection.h
```diff
@@ -24,6 +24,9 @@ class TcpDuplexConnection : public DuplexConnection {
 
   void setInput(yarpl::Reference<DuplexConnection::Subscriber>) override;
 
+  // Only to be used for observation purposes.
+  folly::AsyncSocket* getTransport();
```
alexmalyshev: Nit: I'd name it as `getSocket()` instead.

rsocket/framing/ScheduledFrameTransport.h
```diff
@@ -34,6 +34,15 @@ class ScheduledFrameTransport : public FrameTransport,
   void closeWithError(folly::exception_wrapper ex) override;
 
  private:
+  DuplexConnection* getConnection() override {
+    DLOG(FATAL)
+        << "ScheduledFrameTransport doesn't support getConnection method, "
+            "because it can create safe usage issues when EventBase of the "
+            "transport and the StateMachine is not the same.";
```
alexmalyshev: "StateMachine" here sounds like `RSocketStateMachine`, but  it's actually the "state machine evb" in the `RSocketClient`.  The `RSocketStateMachine` will always use the same thread as the transport.

I'd replace "StateMachine" with "RSocketClient", I feel like that's a little clearer.

rsocket/transports/tcp/TcpDuplexConnection.h
phoad: Let me keep it as getTransport for now. It will make all the functions to have the same name through RSocket, Proxygen, Thrift etc.

PR #807, opened 2017-10-19
Strict debugging for BaseSubscriber implementers

dymk:
Catches errors such as those found in #806, should help with debugging flakey errors in Travis


dymk:
Already builds with -DDEBUG in travis (y)

PR #816, opened 2017-10-28
WarmResumeManager Unittests + Bugfix

somasun:
When a large frame (larger than the cache size) was received, the firstSentPosition wasn't reset appropriately.  Fix it.

Add a suite of unittests.

### DIFF
--- a/rsocket/internal/WarmResumeManager.cpp
+++ b/rsocket/internal/WarmResumeManager.cpp
@@ -33,10 +33,13 @@ void WarmResumeManager::trackSentFrame(
 
     VLOG(6) << "Track sent frame " << frameType
             << " Allowance: " << consumerAllowance;
-    // if the frame is too huge, we don't cache it
+    // If the frame is too huge, we don't cache it.
+    // We empty the entire cache instead.
     if (frameDataLength > capacity_) {
       resetUpToPosition(lastSentPosition_);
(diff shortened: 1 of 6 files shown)


rsocket/internal/WarmResumeManager.cpp
```diff
@@ -33,10 +33,13 @@ void WarmResumeManager::trackSentFrame(
...
     VLOG(6) << "Track sent frame " << frameType
             << " Allowance: " << consumerAllowance;
-    // if the frame is too huge, we don't cache it
+    // If the frame is too huge, we don't cache it.
+    // We empty the entire cache instead.
     if (frameDataLength > capacity_) {
       resetUpToPosition(lastSentPosition_);
       lastSentPosition_ += frameDataLength;
+      firstSentPosition_ += frameDataLength;
+      DCHECK(firstSentPosition_ == lastSentPosition_);
```
alexmalyshev: This and line below should be `DCHECK_EQ`.

rsocket/internal/WarmResumeManager.h
```diff
@@ -71,6 +71,10 @@ class WarmResumeManager : public ResumeManager {
     LOG(FATAL) << "Not Implemented for Warm Resumption";
   }
 
+  size_t size() {
```
alexmalyshev: `size_t size() const {`?

PR #817, opened 2017-10-31
ColdResumeHandler should be nullptr as its not to be used in default

phoad:
When the coldResumeHandler is not null, the handleUnknownStream function uses this object to save a token and some states. 

In default, we don't need this functionality, so it should be assigned to `nullptr` at the constructor's default parameters. Why to pay the cost if we don't want to use it.

For FireAndForget, we don't need to create/register any stream. Don't pay the cost of creating a stream even though we will not use it.

### DIFF
--- a/rsocket/RSocket.h
+++ b/rsocket/RSocket.h
@@ -24,8 +24,7 @@ class RSocket {
       std::shared_ptr<RSocketConnectionEvents> connectionEvents =
           std::shared_ptr<RSocketConnectionEvents>(),
       std::shared_ptr<ResumeManager> resumeManager = nullptr,
-      std::shared_ptr<ColdResumeHandler> coldResumeHandler =
-          std::shared_ptr<ColdResumeHandler>(),
+      std::shared_ptr<ColdResumeHandler> coldResumeHandler = nullptr,
       folly::EventBase* stateMachineEvb = nullptr);
 
   // Creates a RSocketClient which cold-resumes from the provided state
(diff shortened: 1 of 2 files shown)


rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -715,7 +715,9 @@ void RSocketStateMachine::handleUnknownStream(
   // TODO: comparing string versions is odd because from version
   // 10.0 the lexicographic comparison doesn't work
   // we should change the version to struct
-  if (frameSerializer_->protocolVersion() > ProtocolVersion{0, 0} &&
+  DCHECK_GT(frameSerializer_->protocolVersion(), ProtocolVersion(0, 0));
+
+  if (frameType != FrameType::REQUEST_FNF &&
```
somasun: Even FNF frames have StreamIds and I think we should validate them.

rsocket/RSocket.h
```diff
@@ -24,8 +24,7 @@ class RSocket {
       std::shared_ptr<RSocketConnectionEvents> connectionEvents =
           std::shared_ptr<RSocketConnectionEvents>(),
       std::shared_ptr<ResumeManager> resumeManager = nullptr,
-      std::shared_ptr<ColdResumeHandler> coldResumeHandler =
-          std::shared_ptr<ColdResumeHandler>(),
+      std::shared_ptr<ColdResumeHandler> coldResumeHandler = nullptr,
```
lehecka: this change would break the RSocketClient::resume() (RSocketStateMachine::resume) method.

phoad: Actually I only need to update the `createClientFromConnection`, for Thrift integration purposes. Let me remove the remaining updates in this file.

phoad:
I have already updated the project which requires this parameter to be nullptr, to pass nullptr for that parameter. So let me close this diff without landing.

PR #819, opened 2017-11-02
Drop ProtocolVersion argument to SetupResumeAcceptor

alexmalyshev:
Have it always autodetect the proper version.  Right now this is how we always
use SetupResumeAcceptor.  In the future we'll drop the 0.1 version and only use
1.0, but that still won't require an explicit version parameter.


alexmalyshev:
Test failures are spurious bs about failing to allocate `[::1]:0`.  I'll look at them when I have the time.

PR #820, opened 2017-11-02
Microbatching TcpReaderWriter

dymk:


phoad:
There are some `timeout` related features that the clients cancel the tasks if the server doesn't reply in timely manner. It will be beneficial if we can test `timeout` with the updated code before doing more cleanup etc., we can also have a feeling of how much improvement we do performance wise.

dymk:
Alright, perf tests show that this is very promising. Running the Thrift benchmarks shows a near 2x increase in QPS. Math below; numbers are the QPS of the benchmark with neither, both, or either/or client/server using the new batching TcpReaderWriter. 

```
both = 1131412.6363636365
neither = 588488.8333333334
only_client = 596177.1609195402
only_server = 876601.59375

both/neither => 1.9225728209
only_client/neither => 1.0130645259
only_server/neither => 1.4895806753
```

dymk:
Closing until the transport layer story for RSocket stabilizes.

PR #824, opened 2017-11-06
Add utility functions to store ResumeToken

somasun:
These functions will be used to persist resumeId for cold-resumption

### DIFF
--- a/rsocket/internal/Common.cpp
+++ b/rsocket/internal/Common.cpp
@@ -92,6 +92,30 @@ StreamInterruptedException::StreamInterruptedException(int _terminatingSignal)
 
 ResumeIdentificationToken::ResumeIdentificationToken() {}
 
+ResumeIdentificationToken::ResumeIdentificationToken(const std::string& token) {
+  auto getNibble = [&token](size_t i) {
+    uint8_t nibble;
+    if (token[i] >= '0' && token[i] <= '9') {
+      nibble = token[i] - '0';
+    } else if (token[i] >= 'a' && token[i] <= 'f') {
+      nibble = token[i] - 'a' + 10;
+    } else {
+      throw std::invalid_argument("ResumeToken not in right format: " + token);
+    }
+    return nibble;
+  };
+  if (token.size() < 2 || token[0] != '0' || token[1] != 'x' ||
+      (token.size() % 2) != 0) {
+    throw std::invalid_argument("ResumeToken not in right format: " + token);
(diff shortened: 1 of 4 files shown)


rsocket/internal/Common.cpp
```diff
@@ -92,6 +92,30 @@ StreamInterruptedException::StreamInterruptedException(int _terminatingSignal)
 
 ResumeIdentificationToken::ResumeIdentificationToken() {}
 
+ResumeIdentificationToken::ResumeIdentificationToken(const std::string& token) {
+  auto getNibble = [&token](size_t i) {
+    uint8_t nibble;
+    if (token[i] >= '0' && token[i] <= '9') {
```
alexmalyshev: Can you use `strtol()` here or `std::string::stoi()` here?

rsocket/internal/Common.cpp
```diff
@@ -107,15 +131,20 @@ void ResumeIdentificationToken::set(std::vector<uint8_t> newBits) {
...
     out << HEX_CHARS[b & 0x0F];
   }
-  return out;
+  return out.str();
+}
+
+std::ostream& operator<<(
+    std::ostream& out,
+    const ResumeIdentificationToken& token) {
+  return out << token.str();
```
alexmalyshev: I think it should be the other way.  We shouldn't need to create a string to write to an `ostream`, especially if the `.str()` method is working with an `ostream` internally.

rsocket/internal/Common.cpp
somasun: `std::strtol()`'s parameters make it harder to convert only one char

`std::string::stoi()` doesn't get compiled with our setup, likely because of
https://stackoverflow.com/questions/19285775/why-are-stdstoi-and-stdarray-not-compiling-with-g-c11
It also has similar parameters like `std::strtol()`

alexmalyshev: Isn't it just `strtol(token.data() + i, token.data() + i + 1, 16)`?

That `stoi()` thing is weird, never seen that before...

somasun: Unfortunately no,  
```
*endptr*
Reference to an object of type char*, whose value is set by the function to the next character in str after the numerical value.
This parameter can also be a null pointer, in which case it is not used.
```

alexmalyshev: *blows raspberry*

PR #825, opened 2017-11-06
fix state checking for RequestResponseResponder

dymk:


rsocket/statemachine/RequestResponseResponder.cpp
```diff
@@ -13,16 +13,24 @@ using namespace yarpl::flowable;
...
   producingSubscription_ = std::move(subscription);
 }
 
 void RequestResponseResponder::onSuccess(Payload response) noexcept {
-  DCHECK(producingSubscription_) << "didnt call onSubscribe";
+#ifdef DEBUG
+  DCHECK(gotOnSubscribe_.load()) << "didnt call onSubscribe";
+  DCHECK(!gotTerminating_.exchange(true)) << "Already called onSuccess/onError";
+#endif
+  if(!producingSubscription_) return;
```
phoad: nit: format doesn't match to standards.

rsocket/statemachine/RequestResponseResponder.cpp
```diff
@@ -13,16 +13,24 @@ using namespace yarpl::flowable;
...
     subscription->cancel();
     return;
   }
-  DCHECK(!producingSubscription_);
   producingSubscription_ = std::move(subscription);
 }
 
 void RequestResponseResponder::onSuccess(Payload response) noexcept {
-  DCHECK(producingSubscription_) << "didnt call onSubscribe";
+#ifdef DEBUG
```
phoad: As we do DCHECK, why do we need DEBUG flag checking?

dymk: So we're not compiling in the additional atomics in debug code, which would be a perf hit for opt builds as well

rsocket/statemachine/RequestResponseResponder.cpp
dymk: oops (y)

rsocket/statemachine/RequestResponseResponder.cpp
```diff
@@ -37,7 +45,11 @@ void RequestResponseResponder::onSuccess(Payload response) noexcept {
 }
 
 void RequestResponseResponder::onError(folly::exception_wrapper ex) noexcept {
-  DCHECK(producingSubscription_);
+#ifdef DEBUG
+  DCHECK(gotOnSubscribe_.load()) << "didnt call onSubscribe";
+  DCHECK(!gotTerminating_.exchange(true)) << "Already called onSuccess/onError";
+#endif
+
```
phoad: nit: do we need the same if check in here also?

rsocket/statemachine/RequestResponseResponder.cpp
phoad: DCHECK lines will get deleted, isn't it?

dymk: They will, but DCHECKs won't be tested in opt code either, so we're not doing less bug checking compared to before

phoad: What I mean is do we need the `#ifndef DEBUG #endif` ,, these two lines at all?
As the line, which has DCHECK will be immediately removed in a non-debug build, why do we need to protect that region with #ifndef DEBUG?

dymk: The `DCHECK(x)` effectively compiles to `while(false) { CHECK(x) }`, so the expression `x` needs to be valid. `#define`ing out the atomics but not the DCHECKS would lead to "undefined symbol gotOnSubscribe_" type compilation errors.

rsocket/statemachine/RequestResponseResponder.cpp
dymk: The next line unconditionally clears producingSubscriber

rsocket/statemachine/RequestResponseResponder.cpp
phoad: Cool. This is what I was asking for.

rsocket/statemachine/RequestResponseResponder.cpp
phoad: But, what about the applicationError and closeStream calls? Do they try to send anything to the client side?

PR #829, opened 2017-11-13
Fix the build of yarpl under C++17 mode

Orvid:
`std::is_callable` became `std::is_invocable` and got a slightly different API shortly before C++17 was finalized.
To stay compatible with previous versions, use the backports in Folly.


Orvid:
The build passed, a test failed. I would expect this diff to only have build-time visible failures, and to have no runtime visible difference given that this only changes `enable_if` guards in ways that the compilation would simply fail if they were wrong.

dymk:
Almost certainly a flakey test, I've reenqueued the failing build&run.

PR #830, opened 2017-11-16
[Yarpl] Implement flatMap for Flowable

dymk:
WIP implementation of flow aware flatMap

TODO: 
 - Take an arbitrary lambda with a compatible type signature instead of a `folly::Function`
 - Buffer more than 1 element at a time (1 element currently for simplicity and to verify higher-level threading synchronization is correct). This should be the largest perf win. 

What has been tested to work:
 - Requesting an entire stream
 - Requesting 'n' elements of a stream, and having number of requested elements be respected
 - Error coming from the mapping function
 - Error coming from a Flowable coming from the mapping function
 - Pumping the flowable from multiple threads

### DIFF
--- a/yarpl/include/yarpl/flowable/FlowableOperator.h
+++ b/yarpl/include/yarpl/flowable/FlowableOperator.h
@@ -3,6 +3,7 @@
 #pragma once
 
 #include <cassert>
+#include <mutex>
 #include <utility>
 
 #include "yarpl/flowable/Flowable.h"
@@ -10,7 +11,9 @@
 #include "yarpl/flowable/Subscription.h"
 #include "yarpl/utils/credits.h"
 
+#include <boost/intrusive/list.hpp>
 #include <folly/Executor.h>
+#include <folly/Synchronized.h>
 #include <folly/functional/Invoke.h>
 
 namespace yarpl {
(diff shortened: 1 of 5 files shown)


yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +466,146 @@ class FromPublisherOperator : public Flowable<T> {
...
+      LOG(INFO) << "FMSubscription::request(" << n << ")";
+      requested_ += n;
+
+      if (!currentStreamSubscriber_ && !pending_) {
+        LOG(INFO) << "FMSubscription::request(" << n
+                  << ") nothing pending, requesting!";
+        pending_ = true;
+        SuperSubscription::request(1);
+      } else {
+        CHECK(currentStreamSubscriber_);
```
lehecka: can you DCHECK(pending_) for the sake of the reader's mind? If I am reading the code right, it should be always true.

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +466,146 @@ class FromPublisherOperator : public Flowable<T> {
...
+        LOG(INFO) << "FMSubscription::request(" << n
+                  << ") passing request through to currentStreamSubscriber_";
+        currentStreamSubscriber_->request(n);
```
lehecka: this will send request(n) only to the last newStream subscription, the rest of them will get stuck eventually without any request(k).

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +466,146 @@ class FromPublisherOperator : public Flowable<T> {
...
+      }
+    }
+
+    void cancel() override {
+      LOG(INFO) << "FMSubscription::cancel()";
+      SuperSubscription::cancel();
```
lehecka: this should cancel all newStream subscriptions, not only the last one.

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +466,146 @@ class FromPublisherOperator : public Flowable<T> {
...
+     public:
+      MappedStreamSubscriber(Reference<FMSubscription> subscription)
+          : flatMapSubscription_(std::move(subscription)) {}
+
+      void onSubscribeImpl() final {
+        LOG(INFO) << "MappedStreamSubscriber::onSubscribeImpl()";
+        if (flatMapSubscription_->requested_) {
+          LOG(INFO) << "MappedStreamSubscriber::onSubscribeImpl(): requsting "
+                    << flatMapSubscription_->requested_;
+          BaseSubscriber<R>::request(flatMapSubscription_->requested_);
```
lehecka: as mentioned above, you can't give all credits just to one subscriptions because the other ones would never make any progress.

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +466,146 @@ class FromPublisherOperator : public Flowable<T> {
...
+      };
+
+      void onNextImpl(R value) final {
+        LOG(INFO) << "MappedStreamSubscriber::onNextImpl(" << value << ")";
+        flatMapSubscription_->mappedOnNextImpl(std::move(value));
+      }
+
+      void onCompleteImpl() final {
+        LOG(INFO) << "MappedStreamSubscriber::onCompleteImpl()";
+        flatMapSubscription_->pending_ = false;
```
lehecka: there are many newStream subscriptions in flight, if one completes, you should be processing the rest of them, and not terminate the down stream.

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +466,146 @@ class FromPublisherOperator : public Flowable<T> {
...
+        flatMapSubscription_->currentStreamSubscriber_ = nullptr;
+        flatMapSubscription_->mappedOnCompleteImpl();
+      }
+      void onErrorImpl(folly::exception_wrapper ex) final {
+        LOG(INFO) << "MappedStreamSubscriber::onErrorImpl()";
+        flatMapSubscription_->onError(std::move(ex));
+        flatMapSubscription_->pending_ = false;
```
lehecka: same here.

yarpl/test/FlowableFlatMapTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+  auto subscriber = make_ref<TestSubscriber<int64_t>>(1);
+  f->subscribe(subscriber);
+
+  EXPECT_EQ(1UL, subscriber->values().size());
+  EXPECT_TRUE(
+      validate_flatmapped_values(subscriber->values(), {{10}, {20}, {30}}));
+
+  subscriber->request(3);
+  EXPECT_TRUE(validate_flatmapped_values(
```
phoad: nit, check values().size() once more?

yarpl/test/FlowableFlatMapTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+  EXPECT_TRUE(subscriber->isError());
+  EXPECT_EQ(subscriber->getErrorMsg(), "throwing in mapper!");
+}
+
+TEST(FlowableFlatMapTest, MappedStreamThrows) {
+  folly::Function<Reference<Flowable<int64_t>>(int)> func = [](int n) {
+    CHECK_EQ(1, n);
+
+    // flowable which emits an onNext, then the next iteration, emits an error
+    int64_t i = 1;
```
phoad: this 'i' will get deleted before it gets used?, already you pass it by value, so how can it become 0 at all?

yarpl/test/FlowableFlatMapTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+  auto sub = yarpl::make_ref<TestSubscriber<int>>();
+  sub->dropValues(true);
+
+  f->subscribe(sub);
+
+  sub->awaitTerminalEvent(std::chrono::seconds{5});
+  EXPECT_EQ(60000, sub->getValueCount());
+  EXPECT_TRUE(sub->isComplete());
+
+  p1->evb.stop();
```
phoad: Let's don't stop them and see what happens?

yarpl/test/FlowableFlatMapTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  folly::Function<void(int64_t)> onRequest;
+  folly::Function<void(void)> onCancel;
+};
+
+struct FlowableEvbPair {
+  FlowableEvbPair() = default;
+  Reference<Flowable<int>> flow{nullptr};
+  folly::EventBaseThread evb{};
```
phoad: Make it a ScopedEventBaseThread?, so we will not need to stop them?

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +470,379 @@ class FromPublisherOperator : public Flowable<T> {
...
+        : public BaseSubscriber<R>,
+          public boost::intrusive::list_base_hook<
+              boost::intrusive::link_mode<boost::intrusive::auto_unlink>> {
+      MappedStreamSubscriber(Reference<FMSubscription> subscription)
+          : flatMapSubscription_(std::move(subscription)) {}
+
+      void onSubscribeImpl() final {
+        {
+          auto l = flatMapSubscription_->lists.wlock();
+          CHECK(is_in_list(*this, l->pendingValue, l)) << " this=" << this;
```
phoad: isn't it possible that none of the subscriptions wait for any value initially?

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +470,379 @@ class FromPublisherOperator : public Flowable<T> {
...
+      clearAllSubscribers_.store(true);
+    }
+
+    void onTerminateImpl() final {
+      liveSubscribers_--;
+      drainLoop();
+    }
+
+    void request(int64_t n) override {
+      requested_ += n;
```
phoad: nit: check for overflow?

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +470,379 @@ class FromPublisherOperator : public Flowable<T> {
...
+    void onSubscribeImpl() final {
+      liveSubscribers_++;
+      SuperSubscription::onSubscribeImpl();
+      // TODO: make max parallelism configurable a-la RxJava 2.x's
+      // FlowableFlatMapOperator
+      SuperSubscription::request(std::numeric_limits<int64_t>::max());
+    }
+
+    void onNextImpl(T value) final {
+      auto&& flatMapOp = this->getFlowableOperator();
```
phoad: should we check if the Subscription already got an exception or closed?

yarpl/test/FlowableFlatMapTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+    onCancel();
+  }
+
+  folly::Function<void(int64_t)> onRequest;
+  folly::Function<void(void)> onCancel;
+};
+
+struct FlowableEvbPair {
+  FlowableEvbPair() = default;
+  Reference<Flowable<int>> flow{nullptr};
```
phoad: let's make this field's name be 'flowable'.

yarpl/test/FlowableFlatMapTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+  auto f = Flowables::range(0, 2)->flatMap(std::move(mappingFunc));
+
+  auto sub = yarpl::make_ref<TestSubscriber<int>>(0);
+  f->subscribe(sub);
+
+  sub->request(2);
+  sub->awaitValueCount(2);
+  EXPECT_TRUE(validate_flatmapped_values(sub->values(), {{10, 11}, {20, 21}}));
+
+  sub->cancel();
```
phoad: nit: check the count in sub->values() ?

yarpl/test/FlowableFlatMapTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+  subscriber->request(3);
+  EXPECT_TRUE(validate_flatmapped_values(
+      subscriber->values(), {{10, 11}, {20, 21, 22}, {30, 31, 32, 33}}));
+
+  subscriber->cancel();
+  EXPECT_EQ(subscriber->getValueCount(), 4);
+}
+
+TEST(FlowableFlatMapTest, MappingLambdaThrowsErrorOnFirstCall) {
```
phoad: What if only one of the providers of the flatmap errors? Should we drop it and continue with the rest or should we error?

Can we check the case that if doesn't throw for the first call and we have multiple providers?

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: (y)

yarpl/test/FlowableFlatMapTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+  std::vector<int64_t> ret;
+  std::copy_if(in.begin(), in.end(), std::back_inserter(ret), pred);
+  return ret;
+}
+
+std::vector<int64_t>
+filter_range(std::vector<int64_t> in, int64_t startat, int64_t endat) {
+  CHECK_LE(startat, endat);
+  return filter_run(
+      in, [=](int64_t i) { return (i >= startat) && (i < endat); });
```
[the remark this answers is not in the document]
dymk: 'in' isn't captured in this lambda

yarpl/test/FlowableFlatMapTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+  return folly::Function<Reference<Flowable<int64_t>>(int)>([](int n) {
+    switch (n) {
+      case 10:
+        return Flowables::range(n, 2);
+      case 20:
+        return Flowables::range(n, 3);
+      case 30:
+        return Flowables::range(n, 4);
+    }
+    return Flowables::range(n, 3);
```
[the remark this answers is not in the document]
dymk: Some compilers won't notice that all paths in the switch are return paths, iirc, so it'll warn against a missing return

yarpl/test/FlowableFlatMapTest.cpp
dymk: Because it's captured by value in a mutable lambda, so it'll dec each time the lambda fires

yarpl/test/FlowableFlatMapTest.cpp
dymk: (y)

yarpl/test/FlowableFlatMapTest.cpp
dymk: It needs to stop to flush all the enqueued work, else we're leaking threads. This is effectively 'join'.

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: That'll happen on each iteration of the work loop; we want to do checks like that in there so we know all work is serialized

dymk:
Thanks for the review @phoad, I'll start on the fixes :)

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: The next iteration of drainImpl() will take care of the cancel() or onError cases; because of the async nature of flatMap, it's unknown if a mapping subscriber or the upstream will throw an error first, so we serialize in drainImpl()

yarpl/include/yarpl/flowable/TestSubscriber.h
```diff
@@ -218,7 +227,15 @@ class TestSubscriber :
     }
   }
 
+  void dropValues(bool drop) {
+    dropValues_ = drop;
```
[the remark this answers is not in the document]
dymk: nice catch ;)

yarpl/test/FlowableFlatMapTest.cpp
dymk: I thought about this; I'm not sure if we should ignore the stream or not. @benjchristensen could you weigh in here? 

It seems like a bad idea to drop errors silently; perhaps if there's another way to supply the downstream with the errors? Eg a flatMap variant with the signature `Flowable<T> -> (T -> Flowable<R>) -> Flowable<Either<R, Error>>` ?

yarpl/test/FlowableFlatMapTest.cpp
dymk: Another reason they're not scoped: We need a way to manually signal to the EventBaseThread that we want it to flush its work queue and join; else it'll never be destroyed, as the Flowable has a reference to it, and it has a reference to the flowable. Cycle needs to be broken somewhere.

yarpl/test/FlowableFlatMapTest.cpp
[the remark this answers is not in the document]
phoad: http://en.cppreference.com/w/cpp/language/lambda

I believe when you pass `=` to the lambda capture list, it also captures the parameters of the function. So not only startat and endat, but also `in` might be getting captured?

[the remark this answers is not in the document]
dymk: From the `Explanation: captures` section:
`captures all automatic variables used in the body of the lambda by {reference, copy} and current object by reference if exists`

Lambda doesn't reference 'in', so it won't be captured

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -3,14 +3,18 @@
...
 #include <utility>
 
 #include "yarpl/flowable/Flowable.h"
 #include "yarpl/flowable/Subscriber.h"
 #include "yarpl/flowable/Subscription.h"
 #include "yarpl/utils/credits.h"
 
+#include <boost/intrusive/list.hpp>
 #include <folly/Executor.h>
+#include <folly/SpinLock.h>
```
phoad: do we use this?

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +470,384 @@ class FromPublisherOperator : public Flowable<T> {
...
+      // this is both a Subscriber and a Subscription<T>
+      Reference<FMSubscription> flatMapSubscription_{nullptr};
+    };
+
+    // friend struct MappedStreamSubscriber;
+
+    // used to make sure only one thread at a time is calling subscriberOnNext
+    std::atomic<int64_t> drainLoopMutex_{0};
+
+    using subscriber_list_type = boost::intrusive::list<
```
phoad: use camelCase ?

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +470,384 @@ class FromPublisherOperator : public Flowable<T> {
...
+      // FMSubscription's 'reference' to this object. FMSubscription
+      // clears this reference when it drops the MappedStreamSubscriber
+      // from one of its atomic lists
+      Reference<MappedStreamSubscriber> fmReference_{nullptr};
+
+      // this is both a Subscriber and a Subscription<T>
+      Reference<FMSubscription> flatMapSubscription_{nullptr};
+    };
+
+    // friend struct MappedStreamSubscriber;
```
phoad: ?

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +470,384 @@ class FromPublisherOperator : public Flowable<T> {
...
+    }
+
+    template <typename L>
+    static bool in_list_impl(
+        MappedStreamSubscriber const& elem,
+        subscriber_list_type const& list,
+        L const& /*lists*/,
+        bool should) {
+      if (is_in_list(elem, list) != should) {
+        // LOG(INFO) << "in without: " << is_in_list(elem, lists->withoutValue);
```
phoad: let's don't keep commented out code around..

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +470,384 @@ class FromPublisherOperator : public Flowable<T> {
...
+        : public BaseSubscriber<R>,
+          public boost::intrusive::list_base_hook<
+              boost::intrusive::link_mode<boost::intrusive::auto_unlink>> {
+      MappedStreamSubscriber(Reference<FMSubscription> subscription)
+          : flatMapSubscription_(std::move(subscription)) {}
+
+      void onSubscribeImpl() final {
+        {
+          auto l = flatMapSubscription_->lists.wlock();
+          DCHECK(is_in_list(*this, l->pendingValue, l)) << " this=" << this;
```
phoad: Is it possible to just get rid of all these 'is_in_list' calls and the corresponding function definitions.
If we want to be sure, lets write more tests..

phoad:
Is it possible for you to also put replies to the Ondrej's comments so I can understand if they are still valid or fulfilled?

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: These are invariant checks, I think it's helpful to leave them in. Else we'll be adding them back in when we're trying to debug issues related to these classes.

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: Oops, that's dead. I was testing if folly::Synchronized access was any faster with a spinlock.

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: oops, vestigial

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: n/a with updates

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: n/a with updates

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: n/a with updates

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: n/a with updates

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: n/a with updates

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +466,146 @@ class FromPublisherOperator : public Flowable<T> {
...
+
+    void onNextImpl(T value) override {
+      LOG(INFO) << "FMSubscription::onNextImpl(" << value << ")";
+      CHECK(!currentStreamSubscriber_ && pending_);
+
+      auto&& flatMapOp = this->getFlowableOperator();
+      Reference<Flowable<R>> newStream = flatMapOp->function_(std::move(value));
+      currentStreamSubscriber_ =
+          yarpl::make_ref<MappedStreamSubscriber>(this->ref_from_this(this));
+      newStream->subscribe(currentStreamSubscriber_);
```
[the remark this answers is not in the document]
dymk: (y) new implementation is multi-flowable aware

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +466,146 @@ class FromPublisherOperator : public Flowable<T> {
...
+    }
+
+    void onCompleteImpl() override {
+      LOG(INFO) << "FMSubscription::onCompleteImpl()";
+      SuperSubscription::onCompleteImpl();
```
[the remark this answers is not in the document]
dymk: n/a with updates

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +466,146 @@ class FromPublisherOperator : public Flowable<T> {
...
+    }
+
+    void onErrorImpl(folly::exception_wrapper ex) override {
+      LOG(INFO) << "FMSubscription::onErrorImpl()";
+      SuperSubscription::onErrorImpl(std::move(ex));
```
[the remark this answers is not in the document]
dymk: n/a with updates

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -466,6 +466,146 @@ class FromPublisherOperator : public Flowable<T> {
...
+
+    void request(int64_t n) override {
+      LOG(INFO) << "FMSubscription::request(" << n << ")";
+      requested_ += n;
+
+      if (!currentStreamSubscriber_ && !pending_) {
+        LOG(INFO) << "FMSubscription::request(" << n
+                  << ") nothing pending, requesting!";
+        pending_ = true;
+        SuperSubscription::request(1);
```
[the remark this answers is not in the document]
dymk: n/a with updates

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: n/a with updates

dymk:
Went ahead and replied to Ondrej's comments. Most aren't applicable anymore, I did a full rewrite of flatMap once I understood that it was supposed to be merging flowables rather than just concatenating them.

.travis.yml
```diff
@@ -80,7 +80,7 @@ matrix:
 
 env:
   global:
-    - BUILD_TYPE=Debug ASAN=On CTEST_OUTPUT_ON_FAILURE=1
+    - BUILD_TYPE=Debug ASAN=On CTEST_OUTPUT_ON_FAILURE=1 LSAN_OPTIONS=verbosity=1:log_threads=1
```
phoad: nit: if you would like, let's separate this change to another diff. otherwise it is ok.

dymk: (y) this is split out into #836

PR #831, opened 2017-11-22
Make it possible to set EventBase of RSocketRequester from outside

phoad:
Add a method to RSocketRequester that makes it possible to change the EventBase* data member in the RSocketRequester to be overriden.

With this simple change, we make it possible to perform EventBase switching on the client side for Thrift.

### DIFF
--- a/rsocket/RSocketRequester.cpp
+++ b/rsocket/RSocketRequester.cpp
@@ -17,14 +17,18 @@ namespace rsocket {
 RSocketRequester::RSocketRequester(
     std::shared_ptr<RSocketStateMachine> srs,
     EventBase& eventBase)
-    : stateMachine_(std::move(srs)), eventBase_(eventBase) {}
+    : stateMachine_(std::move(srs)), eventBase_(&eventBase) {}
 
 RSocketRequester::~RSocketRequester() {
   VLOG(1) << "Destroying RSocketRequester";
 }
 
+void RSocketRequester::setEventBase(folly::EventBase* evb) {
+  eventBase_ = evb;
+}
+
 void RSocketRequester::closeSocket() {
-  eventBase_.add([stateMachine = std::move(stateMachine_)] {
+  eventBase_->add([stateMachine = std::move(stateMachine_)] {
     VLOG(2) << "Closing RSocketStateMachine on EventBase";
     stateMachine->close(
         folly::exception_wrapper(), StreamCompletionSignal::SOCKET_CLOSED);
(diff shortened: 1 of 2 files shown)


rsocket/RSocketRequester.cpp
```diff
@@ -17,14 +17,18 @@ namespace rsocket {
...
     EventBase& eventBase)
-    : stateMachine_(std::move(srs)), eventBase_(eventBase) {}
+    : stateMachine_(std::move(srs)), eventBase_(&eventBase) {}
 
 RSocketRequester::~RSocketRequester() {
   VLOG(1) << "Destroying RSocketRequester";
 }
 
+void RSocketRequester::setEventBase(folly::EventBase* evb) {
+  eventBase_ = evb;
```
lexs: There is no way this can be safe, if there are pending callbacks on the old evb you'll now have interleaved callbacks from different threads. This needs to use the EventBase switcher that @dymk (I think?) worked on. Basically enqueuing the change on the original evb and then collecting callbacks until that executes. See https://github.com/rsocket/rsocket-cpp/blob/master/rsocket/internal/SwappableEventBase.h

dymk: Yep, I think this would require rewriting RSocketRequester to use SwappableEventBase and then enqueuing the swap on that, to keep events in rsocket nice and serialized.

phoad: The only case that we need event base switching is for Thrift integration. Thrift integration doesn't require having SwappableEventBase implementation. At least for the current state. 

Neither Proxygen have anything like SwappableEventBase. It is not a requirement for this feature.

rsocket/RSocketRequester.cpp
```diff
@@ -23,8 +23,15 @@ RSocketRequester::~RSocketRequester() {
...
 }
 
-void RSocketRequester::setEventBase(folly::EventBase* evb) {
-  eventBase_ = evb;
+bool RSocketRequester::canReplaceEventBase() {
+  return eventBaseIsReplaceable_;
+}
+
+bool RSocketRequester::setEventBase(folly::EventBase* evb) {
+  if (canReplaceEventBase()) {
```
dymk: Calls to RSocketRequester aren't synchronized, `eventBaseIsReplaceable_` and checks which depend on it need to be protected by a mutex

phoad: Even making it atomic would not help?

lexs: Which thread would setEventBase() be called on? I feel like the multithreading complexity of RSocket is already pretty high.

dymk: @phoad  Both `eventBaseIsReplaceable_` and access to the underlying `eventBase_` need to be protected in the same critical section for them to be synchronized, and to prevent enqueuing on the `eventBase_` then having it be replaced

phoad:
I will have a better fix for this.

PR #832, opened 2017-11-23
Fix unsynchronized access on subscription cancellation

vitaut:
We shouldn't acquire lock in the range expression for reasons explained in https://fburl.com/yl0onxkm.

### DIFF
--- a/yarpl/src/yarpl/observable/Subscriptions.cpp
+++ b/yarpl/src/yarpl/observable/Subscriptions.cpp
@@ -13,7 +13,10 @@ namespace observable {
  */
 void Subscription::cancel() {
   cancelled_ = true;
-  for(auto& subscription : *tiedSubscriptions_.rlock()) {
+  // Lock must be obtained here and not in the range expression for it to
+  // apply to the loop body.
+  auto locked = tiedSubscriptions_.wlock();
+  for(auto& subscription : *locked) {
     subscription->cancel();
   }
 }
@@ -41,7 +44,10 @@ void CallbackSubscription::cancel() {
   // mark cancelled 'true' and only if successful invoke 'onCancel()'
   if (cancelled_.compare_exchange_strong(expected, true)) {
     onCancel_();
-    for(auto& subscription : *tiedSubscriptions_.rlock()) {
+    // Lock must be obtained here and not in the range expression for it to
+    // apply to the loop body.
+    auto locked = tiedSubscriptions_.wlock();
+    for(auto& subscription : *locked) {
       subscription->cancel();
     }
   }


yarpl/src/yarpl/observable/Subscriptions.cpp
```diff
@@ -13,7 +13,10 @@ namespace observable {
  */
 void Subscription::cancel() {
   cancelled_ = true;
-  for(auto& subscription : *tiedSubscriptions_.rlock()) {
+  // Lock must be obtained here and not in the range expression for it to
+  // apply to the loop body.
+  auto locked = tiedSubscriptions_.rlock();
```
dymk: needs exclusive access; `wlock()`

yarpl/src/yarpl/observable/Subscriptions.cpp
```diff
@@ -41,7 +44,10 @@ void CallbackSubscription::cancel() {
   // mark cancelled 'true' and only if successful invoke 'onCancel()'
   if (cancelled_.compare_exchange_strong(expected, true)) {
     onCancel_();
-    for(auto& subscription : *tiedSubscriptions_.rlock()) {
+    // Lock must be obtained here and not in the range expression for it to
+    // apply to the loop body.
+    auto locked = tiedSubscriptions_.rlock();
```
dymk: same here

yarpl/src/yarpl/observable/Subscriptions.cpp
phoad: Actually I am not sure if we want the lock to be applied to the loop body at all. 
Namely the only reason for locking might be to access to the list in a safe way, but it doesn't mean that we would like to make the access to the subscription be synchronized. 

'subscription' in here is already a Reference, so it will not get destroyed or else.

So I don't believe this change is necessary..

vitaut:
There is more to it that meets the eye. Even if you don't need synchronization for elements, you can still have a crash when accessing an iterator that was invalidated by racing `push_back` in `cancel`. Therefore you need to hold the lock for the whole duration of iteration.

> This won't fix all the synchronization issues

Sure, the purpose of this diff is to fix the most severe and obvious misuse of `Synchronized` that may result in a crash, not fix all the synchronization issues in the class. Please open an issue for latter.

dymk:
Still needs an exclusive lock (eg `wlock()`, or consider `folly::Synchronized<..., std::mutex>`)

vitaut:
@dymk, could you elaborate why you think an exclusive lock is needed when iterating over `tiedSubscriptions_`?

dymk:
@vitaut Because we're mutating the elements of `tiedSubscriptions_`, by canceling them. We also don't want to call `cancel` in a non-synchronized way, according to the reactive streams spec.

vitaut:
Changed to write locks.

PR #838, opened 2017-12-05
Implement merge operator, bugfixes for flatMap

dymk:
Implements the merge operator for Flowable, which converts a Flowable<Flowable<T>> into a Flowable<T>. It's implemented as a specialization of `flatMap`. 

Also included are bugfixes for flatMap, in which flatMap would terminate its subscriber too early, once all of this sub-streams had completed (but possibly still had outstanding elements to be consumed).

### DIFF
--- a/yarpl/include/yarpl/flowable/Flowable.h
+++ b/yarpl/include/yarpl/flowable/Flowable.h
@@ -26,10 +26,10 @@ class Flowable;
 namespace detail {
 
 template <typename T>
-struct is_t_flowable : std::false_type {};
+struct IsFlowable : std::false_type {};
 
 template <typename R>
-struct is_t_flowable<Reference<Flowable<R>>> : std::true_type {
+struct IsFlowable<Reference<Flowable<R>>> : std::true_type {
   using ElemType = R;
 };
 
@@ -96,7 +96,7 @@ class Flowable : public virtual Refcounted, public yarpl::enable_get_ref {
 
   template <
       typename Function,
-      typename R = typename detail::is_t_flowable<
+      typename R = typename detail::IsFlowable<
           typename std::result_of<Function(T)>::type>::ElemType>
   Reference<Flowable<R>> flatMap(Function func);
 
@@ -120,7 +120,7 @@ class Flowable : public virtual Refcounted, public yarpl::enable_get_ref {
 
   template <typename Q>
   using enableWrapRef =
-      typename std::enable_if<detail::is_t_flowable<Q>::value, Q>::type;
+      typename std::enable_if<detail::IsFlowable<Q>::value, Q>::type;
 
   template <typename Q = T>
   enableWrapRef<Q> merge() {
--- a/yarpl/include/yarpl/flowable/Flowable.h
+++ b/yarpl/include/yarpl/flowable/Flowable.h
@@ -21,6 +21,21 @@ namespace yarpl {
 namespace flowable {
 
 template <typename T>
+class Flowable;
+
+namespace detail {
+
+template <typename T>
+struct is_t_flowable : std::false_type {};
+
+template <typename R>
+struct is_t_flowable<Reference<Flowable<R>>> : std::true_type {
+  using ElemType = R;
+};
+
+} // namespace detail
+
+template <typename T>
 class Flowable : public virtual Refcounted, public yarpl::enable_get_ref {
  public:
   virtual void subscribe(Reference<Subscriber<T>>) = 0;
@@ -79,8 +94,11 @@ class Flowable : public virtual Refcounted, public yarpl::enable_get_ref {
       typename R = typename std::result_of<Function(T)>::type>
   Reference<Flowable<R>> map(Function function);
 
-  template <typename R>
-  Reference<Flowable<R>> flatMap(folly::Function<Reference<Flowable<R>>(T)>);
+  template <
+      typename Function,
+      typename R = typename detail::is_t_flowable<
+          typename std::result_of<Function(T)>::type>::ElemType>
+  Reference<Flowable<R>> flatMap(Function func);
 
   template <typename Function>
   Reference<Flowable<T>> filter(Function function);
@@ -99,6 +117,15 @@ class Flowable : public virtual Refcounted, public yarpl::enable_get_ref {
   Reference<Flowable<T>> subscribeOn(folly::Executor&);
 
   Reference<Flowable<T>> observeOn(folly::Executor&);
+
+  template <typename Q>
+  using enableWrapRef =
+      typename std::enable_if<detail::is_t_flowable<Q>::value, Q>::type;
+
+  template <typename Q = T>
+  enableWrapRef<Q> merge() {
+    return this->flatMap([](auto f) { return std::move(f); });
+  }
 
   template <
       typename Emitter,
@@ -172,11 +199,10 @@ Reference<Flowable<T>> Flowable<T>::observeOn(folly::Executor& executor) {
 }
 
 template <typename T>
-template <typename R>
-Reference<Flowable<R>> Flowable<T>::flatMap(
-    folly::Function<Reference<Flowable<R>>(T)> func) {
+template <typename Function, typename R>
+Reference<Flowable<R>> Flowable<T>::flatMap(Function function) {
   return make_ref<FlatMapOperator<T, R>>(
-      this->ref_from_this(this), std::move(func));
+      this->ref_from_this(this), std::move(function));
 }
 
 } // flowable
--- a/yarpl/include/yarpl/flowable/FlowableOperator.h
+++ b/yarpl/include/yarpl/flowable/FlowableOperator.h
@@ -47,8 +47,8 @@ class FlowableOperator : public Flowable<D> {
         Reference<Subscriber<D>> subscriber)
         : flowableOperator_(std::move(flowable)),
           subscriber_(std::move(subscriber)) {
-      assert(flowableOperator_);
-      assert(subscriber_);
+      CHECK(flowableOperator_);
+      CHECK(subscriber_);
     }
 
     const Reference<Operator>& getFlowableOperator() {
@@ -534,31 +534,33 @@ class FlatMapOperator : public FlowableOperator<T, R, FlatMapOperator<T, R>> {
 
       liveSubscribers_++;
       mappedStream->subscribe(mappedSubscriber);
+      drainLoop();
     }
 
     void drainImpl() {
       // phase 1: clear out terminated subscribers
       {
-        SubscriberList trash;
-        auto clearList = [&trash](auto& list) {
+        auto clearList = [](auto& list, SubscriberList& t) {
           while (!list.empty()) {
             auto& elem = list.front();
-            elem.explicitlyCanceled = true;
+            auto r = elem.sync.wlock();
+            r->freeze = true;
             elem.unlink();
-            trash.push_back(elem);
+            t.push_back(elem);
           }
         };
 
+        SubscriberList clearTrash;
         if (clearAllSubscribers_.load()) {
           auto l = lists.wlock();
-          clearList(l->withValue);
-          clearList(l->withoutValue);
-          clearList(l->pendingValue);
+          clearList(l->withValue, clearTrash);
+          clearList(l->withoutValue, clearTrash);
+          clearList(l->pendingValue, clearTrash);
         }
 
         // clear elements while no locks are held
-        while (!trash.empty()) {
-          auto& elem = trash.front();
+        while (!clearTrash.empty()) {
+          auto& elem = clearTrash.front();
           elem.unlink();
           elem.cancel();
           elem.fmReference_ = nullptr;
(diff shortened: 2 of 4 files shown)


yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -542,13 +539,13 @@ class FlatMapOperator : public FlowableOperator<T, R, FlatMapOperator<T, R>> {
     void drainImpl() {
       // phase 1: clear out terminated subscribers
       {
-        auto clearList = [](auto& list) {
+        SubscriberList trash;
+        auto clearList = [&trash](auto& list) {
```
phoad: nit: if you make the 'trash' another input to the function, then you might declare that variable when the lambda gets used.

yarpl/test/FlowableFlatMapTest.cpp
```diff
@@ -271,5 +271,44 @@ TEST(FlowableFlatMapTest, MultithreadedLargeAmount) {
   p2->evb.stop();
 }
 
+TEST(FlowableFlatMapTest, MergeTest) {
+  auto sub = yarpl::make_ref<TestSubscriber<std::string>>(0);
+
+  auto p1 = Flowables::justN<std::string>({"foo", "bar"});
```
phoad: It is possible to write a test like this
 - One flowable from Thread A
 - One flowable from Thread B
 - They are merged into another flowable which lives in Thread C
So it will be a total merge of the flowables from different threads into a single thread.

dymk: Tests in this compilation unit do exercise multiple independent threads pushing into flatMap (and therefore merge, as merge is just a very light wrapper around flatMap)

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: I'm not sure I follow, you mean don't capture trash, but pass it in as another parameter?

phoad: nit: would it be better not not?

yarpl/include/yarpl/flowable/Flowable.h
```diff
@@ -20,6 +20,19 @@
...
 
+template <typename T>
+class Flowable;
+
+namespace detail {
+
+template <typename T>
+struct is_t_flowable : std::false_type {};
+
+template <typename R>
```
phoad: Do we need all these or do we make the code more complex, do we have hard requirement for these?

template <typename T> , should we use T instead of R?

yarpl/include/yarpl/flowable/Flowable.h
```diff
@@ -100,6 +113,16 @@ class Flowable : public virtual Refcounted, public yarpl::enable_get_ref {
...
   Reference<Flowable<T>> observeOn(folly::Executor&);
 
+  template <typename Q>
+  using enableWrapRef =
+      typename std::enable_if<detail::is_t_flowable<Q>::value, Q>::type;
+
+  template <typename Q = T>
+  enableWrapRef<Q> merge() {
+    folly::Function<T(T)> func = [](auto f) { return std::move(f); };
+    return this->flatMap(std::move(func));
```
phoad: Would it be possible to just copy paste the 'lambda' part into this. compiler might generate an optimized version instead of folly::Function..

this->flatMap([](auto f) { return std::move(f); });

yarpl/include/yarpl/flowable/FlowableOperator.h
dymk: I think it won't matter; in any optimized build the lambda will be inlined

yarpl/include/yarpl/flowable/Flowable.h
```diff
@@ -100,6 +113,16 @@ class Flowable : public virtual Refcounted, public yarpl::enable_get_ref {
 
   Reference<Flowable<T>> observeOn(folly::Executor&);
 
+  template <typename Q>
+  using enableWrapRef =
```
phoad: Do you use this more than once? Do we need to define it at all?

yarpl/include/yarpl/flowable/Flowable.h
```diff
@@ -20,6 +20,19 @@
...
+template <typename T>
+class Flowable;
+
+namespace detail {
+
+template <typename T>
+struct is_t_flowable : std::false_type {};
+
+template <typename R>
+struct is_t_flowable<Reference<Flowable<R>>> : std::true_type {};
```
alexmalyshev: Nit: I believe the convention is `is_flowable` for the struct type and `is_flowable_v` for the value type.

Thinking of `std::is_signed` and `std::is_signed_v`.

yarpl/include/yarpl/flowable/Flowable.h
dymk: Well the R references a different variable than T; T is `is_t_flowable<T>`, R is the template parameter that `T` is parameterized over in the case that `T` is a `Reference<Flowable<R>>`

yarpl/include/yarpl/flowable/Flowable.h
alexmalyshev: Unrelated, but I prefer `static_assert` always to `enable_if`.

yarpl/include/yarpl/flowable/Flowable.h
dymk: A limitation of flatMap right now is that it can't infer return type from the type of the lambda (templates are... hard. I can work on this more in another diff). It's held in flatMap as a folly::Function, so there won't be a perf hit.

alexmalyshev: +1, I think it'd make things a bit cleaner.

Doesn't really matter, but is the `std::move` necessary here?  I can never remember what the rules for RVO are, and which C++ standard made it mandatory.

yarpl/include/yarpl/flowable/Flowable.h
dymk: It's to help with readability. IMO a complex inline `enable_if` is pretty illegible as the return type for `merge`.

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -620,10 +627,37 @@ class FlatMapOperator : public FlowableOperator<T, R, FlatMapOperator<T, R>> {
...
+
+          // Subscribers might call onNext and then terminate; delay
+          // removing its liveSubscriber reference until we've delivered
+          // its element to the downstream subscriber and dropped its
+          // synchronized reference to `r`, as dropping the flatMapSubscription_
+          // reference may invoke its destructor
+          if (r->isTerminated) {
+            trash.push_back(*elem);
+            continue; // skips the next elem->request(1)
+          } else {
```
alexmalyshev: Nit: Don't need the `else`.

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -749,6 +798,7 @@ class FlatMapOperator : public FlowableOperator<T, R, FlatMapOperator<T, R>> {
         R value;
         bool hasValue{false};
         folly::exception_wrapper onErrorEx{nullptr};
+        bool isTerminated{false};
```
alexmalyshev: Move this just below `hasValue`.  Padding is evil.

yarpl/include/yarpl/flowable/Flowable.h
dymk: That's for the using alias which is the `::value` of the type; not the type itself

dymk:
```
template <typename T>
using is_t_flowable_v = is_t_flowable<T>::value
```
would be the convention iirc

yarpl/include/yarpl/flowable/Flowable.h
dymk: @alexmalyshev the `enable_if` is to SFINAE out `merge` when T is not a Flowable

yarpl/include/yarpl/flowable/Flowable.h
dymk: (y) fixed flatMap to deduce types from the lambda its given directly

yarpl/include/yarpl/flowable/Flowable.h
alexmalyshev: I know, my opinion is that SFINAE produces worse errors than a static_assert failure.

*goes off to the hills to shout about C++ Concepts*

phoad:
P58748359 -- does your update fix this issue?

dymk:
@phoad it does :)

yarpl/include/yarpl/Refcounted.h
```diff
@@ -311,56 +312,49 @@ class AtomicReference final {
...
   Reference<T> load() {
-    return Reference<T>(*this);
+    std::lock_guard<std::mutex> lg{mtx_};
+    if (pointer_) {
+      pointer_->incRef();
+      Reference<T> r{pointer_, detail::skip_initial_refcount_check{}};
+      return r;
+    }
+    else {
+      Reference<T> r{nullptr};
```
alexmalyshev: This else block is equivalent to

  return Reference<T>{};

yarpl/include/yarpl/Refcounted.h
```diff
@@ -311,56 +312,49 @@ class AtomicReference final {
...
-  AtomicReference& operator=(Reference<U>&& other) {
-    return assign(std::move(other));
-  }
-
   Reference<T> load() {
-    return Reference<T>(*this);
+    std::lock_guard<std::mutex> lg{mtx_};
+    if (pointer_) {
+      pointer_->incRef();
+      Reference<T> r{pointer_, detail::skip_initial_refcount_check{}};
```
alexmalyshev: You can get rid of `r` and return the `Reference<T>{...}` directly.

yarpl/include/yarpl/Refcounted.h
```diff
@@ -369,12 +363,14 @@ class AtomicReference final {
...
-    if (auto p = pointer_.load()) {
-      p->decRef();
+    if (pointer_) {
+      pointer_->decRef();
     }
   }
 
-  std::atomic<T*> pointer_{nullptr};
+  std::mutex mtx_;
+  // std::atomic<T*> pointer_{nullptr};
```
alexmalyshev: Dead code?

yarpl/include/yarpl/Refcounted.h
```diff
@@ -369,12 +363,14 @@ class AtomicReference final {
...
 
-    if (auto p = pointer_.load()) {
-      p->decRef();
+    if (pointer_) {
+      pointer_->decRef();
     }
   }
 
-  std::atomic<T*> pointer_{nullptr};
+  std::mutex mtx_;
```
alexmalyshev: sadpanda

yarpl/include/yarpl/flowable/Flowable.h
```diff
@@ -20,6 +20,21 @@
...
+class Flowable;
+
+namespace detail {
+
+template <typename T>
+struct is_t_flowable : std::false_type {};
+
+template <typename R>
+struct is_t_flowable<Reference<Flowable<R>>> : std::true_type {
+  using ElemType = R;
```
alexmalyshev: Nit: Using underscore-lowercase for `is_t_flowable` and then upper-camelcase for `ElemType` is a little odd.

I'd suggest picking one, either `is_t_flowable::value_type` or `IsFlowable::ElemType`.

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -93,7 +93,12 @@ class FlowableOperator : public Flowable<D> {
     // Subscriber.
 
     void onSubscribeImpl() override {
-      subscriber_->onSubscribe(this->ref_from_this(this));
+      if (auto s = subscriber_.load()) {
+        s->onSubscribe(this->ref_from_this(this));
+      }
+      else {
+        CHECK(false) << "subscriber_ is null";
```
alexmalyshev: Why not just crash?  The `CHECK(false)` doesn't really get you much.
```
  auto s = subscriber_.load();
  s->onSubscribe(...);
```

Also `CHECK(false)` should pretty much be `LOG(FATAL)` always.

yarpl/include/yarpl/Refcounted.h
```diff
@@ -311,56 +312,49 @@ class AtomicReference final {
...
     inc();
   }
 
-  explicit operator bool() const {
-    return pointer_;
-  }
-
-  T* operator->() const {
-    return pointer_;
+  void store(std::nullptr_t) {
```
alexmalyshev: Why is this needed?  Is there not an implicit `Reference<T>` ctor from `std::nullptr_t`?

dymk: Just an optimization, same as the Reference<> case. It would also be just excellent if we could get rid of Reference and avoid all this mess.

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -718,12 +782,19 @@ class FlatMapOperator : public FlowableOperator<T, R, FlatMapOperator<T, R>> {
           : flatMapSubscription_(std::move(subscription)) {}
 
       void onSubscribeImpl() final {
+#ifdef NDEBUG
```
alexmalyshev: Did you mean `#ifndef NDEBUG`?  For this, and all follow up uses?  You want the debugging code in the cases where NDEBUG is _not defined_ right, or am I getting myself all flipped up.

yarpl/include/yarpl/flowable/FlowableOperator.h
```diff
@@ -700,6 +757,13 @@ class FlatMapOperator : public FlowableOperator<T, R, FlatMapOperator<T, R>> {
       else {
         requested_ += n;
       }
+
+      if (n > 0) {
+        // TODO: make max parallelism configurable a-la RxJava 2.x's
+        // FlowableFlatMapOperator
+        SuperSubscription::request(std::numeric_limits<int64_t>::max());
```
alexmalyshev: For the short term I think this should be int32_t::max(), cause we have a lot of bad assertions in rsocket that assume that.

PR #840, opened 2017-12-06
ConsumerBase sends excessive numbers of requestN messages

phoad:
Prevent sending requestN(MAX_INT) message again and again.
Request more only if more than half of the requested amount is consumed.

### DIFF
--- a/rsocket/statemachine/ConsumerBase.cpp
+++ b/rsocket/statemachine/ConsumerBase.cpp
@@ -35,6 +35,11 @@ void ConsumerBase::cancelConsumer() {
   consumingSubscriber_ = nullptr;
 }
 
+void ConsumerBase::addImplicitAllowance(size_t n) {
+  allowance_.add(n);
+  activeRequests_.add(n);
+}
+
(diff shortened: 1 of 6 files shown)


rsocket/statemachine/ConsumerBase.cpp
```diff
@@ -97,12 +99,17 @@ void ConsumerBase::errorConsumer(folly::exception_wrapper ex) {
...
 
 void ConsumerBase::sendRequests() {
-  // TODO(stupaq): batch if remote end has some spare allowance
-  // TODO(stupaq): limit how much is synced to the other end
   size_t toSync = Frame_REQUEST_N::kMaxRequestN;
-  toSync = pendingAllowance_.consumeUpTo(toSync);
-  if (toSync > 0) {
-    writeRequestN(static_cast<uint32_t>(toSync));
+  size_t pending = pendingAllowance_.get();
+  toSync = std::min(pending, toSync);
```
alexmalyshev:
```
auto toSync = std::min<size_t>(pendingAllowance_.get(), Frame_REQUEST_N::kMaxRequestN);
```

PR #841, opened 2017-12-08
Turn RSocketStateMachine::disconnect() into close() for non-resumable…

alexmalyshev:
… connections

Making things simpler to follow along.  We have close() and closeWithError() for
closing connections, and now we'll have disconnect() and disconnectWithError()
for disconnecting them.  If a connection can't be reconnected, then disconnect()
should behave like close().

Renames disconnectOrCloseWithError() to disconnectWithError().

Move FrameSink into KeepaliveTimer.h.  It makes sense to have it there and not
with RSocketStateMachine right now.

Inline uses of RSocketStateMachine private methods setResumable(), isClosed(),
and getKeepaliveTime().

### DIFF
--- a/rsocket/statemachine/RSocketStateMachine.cpp
+++ b/rsocket/statemachine/RSocketStateMachine.cpp
@@ -68,17 +68,11 @@ RSocketStateMachine::~RSocketStateMachine() {
   // close method
 }
 
-void RSocketStateMachine::setResumable(bool resumable) {
-  // We should set this flag before we are connected
-  DCHECK(isDisconnected());
-  isResumable_ = resumable;
-}
-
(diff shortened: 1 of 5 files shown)


rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -177,8 +176,6 @@ void RSocketStateMachine::resumeClient(
 
   // Disconnect a previous client if there is one.
   disconnect(std::runtime_error{"Resuming client on a different connection"});
-
-  setResumable(true);
```
alexmalyshev: This needed to be moved up, otherwise the above `disconnect()` would always call `close()` for cold resumption.

PR #842, opened 2017-12-08
increase timeout for travis

dymk:
Travis keeps failing with errors like: 

 - C++ exception with description "timeout in awaitTerminalEvent" thrown in the test body.
 - C++ exception with description "timeout in awaitValueCount" thrown in the test body.

increase the amount of time that we await

Also remove 'make coverage'; we're not using that information, and it increases build/test time by 4 minutes or so.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -111,7 +111,7 @@ script:
   - make -j4
   - lcov --directory . --zerocounters
   - make test
-  - make coverage
+  # - make coverage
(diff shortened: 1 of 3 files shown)


dymk:
I looked into this more, I think it's due to races on `terminated_` in TestSubscriber. The condition var is getting a notification but reading an out of date `terminated_`; this change should fix that

PR #844, opened 2017-12-08
fix AtomicReference correctness

dymk:
Update AtomicReference to be threadsafe. Uses an approach similar to how std::shared_ptr behaves.


dymk:
#845 makes this obsolete

PR #845, opened 2017-12-08
Replace Refcounted with std::shared_ptr

dymk:
Replaces the intrusive yarpl::Refcounted with std::shared_ptr. Inherit `enable_get_ref` from a few more classes. 

Benchmarks show ~5% or so perf improvement with `std::shared_ptr`. Finding the impact of/if there are any memory usage differences should be done via canary (this diff introduces 1 word of overhead per object). 

Original `Reference` implementation
```
Benchmarks.cpp:11] Running benchmarks... (takes minutes)
StreamThroughputTcp.cpp:45] Running:
StreamThroughputTcp.cpp:46]   Server with 8 threads.
StreamThroughputTcp.cpp:47]   10 clients across 10 threads.
StreamThroughputTcp.cpp:49]   Running 1 streams of 1000000 items each.
============================================================================
/Users/dymk/code/c-c++/rsocket-cpp/benchmarks/StreamThroughputTcp.cpprelative  time/iter  iters/s
============================================================================
StreamThroughput                                              8.10s  123.40m
============================================================================
        8.54 real        18.39 user        28.62 sys

Benchmarks.cpp:11] Running benchmarks... (takes minutes)
FireForgetThroughputTcp.cpp:55] Running:
FireForgetThroughputTcp.cpp:56]   Server with 8 threads.
FireForgetThroughputTcp.cpp:57]   10 clients across 10 threads.
FireForgetThroughputTcp.cpp:59]   Running 1000000 requests in total.
============================================================================
/Users/dymk/code/c-c++/rsocket-cpp/benchmarks/FireForgetThroughputTcp.cpprelative  time/iter  iters/s
============================================================================
FireForgetThroughput                                         57.80s   17.30m
============================================================================
       58.24 real        41.56 user       147.67 sys
```

`std::shared_ptr`:

```
Benchmarks.cpp:11] Running benchmarks... (takes minutes)
StreamThroughputTcp.cpp:45] Running:
StreamThroughputTcp.cpp:46]   Server with 8 threads.
StreamThroughputTcp.cpp:47]   10 clients across 10 threads.
StreamThroughputTcp.cpp:49]   Running 1 streams of 1000000 items each.
============================================================================
/Users/dymk/code/c-c++/rsocket-cpp/benchmarks/StreamThroughputTcp.cpprelative  time/iter  iters/s
============================================================================
StreamThroughput                                              7.82s  127.95m
============================================================================
        8.20 real        19.17 user        28.43 sys

Benchmarks.cpp:11] Running benchmarks... (takes minutes)
FireForgetThroughputTcp.cpp:55] Running:
FireForgetThroughputTcp.cpp:56]   Server with 8 threads.
FireForgetThroughputTcp.cpp:57]   10 clients across 10 threads.
FireForgetThroughputTcp.cpp:59]   Running 1000000 requests in total.
============================================================================
/Users/dymk/code/c-c++/rsocket-cpp/benchmarks/FireForgetThroughputTcp.cpprelative  time/iter  iters/s
============================================================================
FireForgetThroughput                                         54.77s   18.26m
============================================================================
       55.21 real        41.35 user       155.63 sys
```


test/ColdResumptionTest.cpp
```diff
@@ -258,7 +258,7 @@ void coldResumer(uint32_t port, uint32_t client_num) {
   }
 }
 
-TEST(ColdResumptionTest, SuccessfulResumption) {
+TEST(ColdResumptionTest, DISABLED_SuccessfulResumption) {
```
dymk: This test was constantly flaking on me. Disabling until we can rootcause why it's flaky.

yarpl/CMakeLists.txt
```diff
@@ -21,18 +21,28 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-weak-vtables -Wno-padded")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-omit-frame-pointer")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -momit-leaf-frame-pointer")
 
+if(YARPL_WRAP_SHARED_IN_LOCK)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DYARPL_WRAP_SHARED_IN_LOCK")
+    message("Compiler lacks support std::atomic<std::shared_ptr>; wrapping with a mutex")
+elseif(YARPL_WRAP_SHARED_IN_ATOMIC)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DYARPL_WRAP_SHARED_IN_ATOMIC")
+    message("Compiler lacks std::shared_ptr atomic overloads; wrapping in std::atomic")
```
lexs: What is the difference between this and the above? Both mention the same thing lacking but different workarounds?

yarpl/include/yarpl/flowable/Subscription.h
```diff
@@ -9,7 +9,7 @@ namespace flowable {
 
 class Subscription : public virtual Refcounted {
  public:
-  virtual ~Subscription() = default;
+  // virtual ~Subscription() = default;
```
lexs: remove?

yarpl/CMakeLists.txt
dymk: Newer compilers/libstdc++ have `std::atomic_{exchange,load,store}` specialized for `std::shared_ptr<T>` directly, so the `std::atomic` structure never shows up. If that's not supported, the code falls back to wrapping an `std::shared_ptr` in `std::atomic` structure. If *that* isn't supported either (my local clang doesn't like it), it falls back further to locking with a mutex and shimming `{exchange,load,store}`.

CMakeLists.txt
```diff
@@ -156,10 +156,35 @@ set(GMOCK_LIBS
 set(CMAKE_CXX_STANDARD 14)
 
 # Common configuration for all build modes.
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
```
alexmalyshev: Huh, I'm surprised we didn't already have this, but sure, makes sense.

CMakeLists.txt
```diff
@@ -156,10 +156,35 @@ set(GMOCK_LIBS
...
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Woverloaded-virtual")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
 
+# does the compiler support std::atomic_load(&std::shared_ptr<T>) ?
+try_compile(HAS_NATIVE_ATOMIC_SHARED_PTR ${CMAKE_BINARY_DIR}/has_shared_ptr_support_rs
+    ${CMAKE_CURRENT_SOURCE_DIR}/yarpl/test/test_has_shared_ptr_support.cpp)
```
alexmalyshev: You could use https://cmake.org/cmake/help/v3.0/module/CheckCXXSourceCompiles.html to write the C++ code inline here if you want.  I wouldn't block the change on it though.

yarpl/include/yarpl/Refcounted.h
```diff
@@ -9,497 +9,87 @@
...
-  detail::inc_live(demangled_name);
-  detail::set_reference_name{std::move(demangled_name), *r};
 #endif
 
-  return std::move(r);
-}
+class Refcounted {};
 
-class enable_get_ref {
+class enable_get_ref : public std::enable_shared_from_this<enable_get_ref> {
```
alexmalyshev: Why do we need this as opposed to good ol' fashioned `std::enable_shared_from_this<T>`?  Is it because we need _some_ base class to be the shared virtual class?

dymk: Yep, we just need something to be the inherited class (not even virtual).

CMakeLists.txt
lehecka: I think this parameter is coming down from travis. Do we need to explicitly set it in cmake?

dymk:
Pulled in via b3f19273e41199cee6aef88a0df531a681a5a86c

PR #848, opened 2017-12-19
fbcode-ify structure of rsocket bench/test/examples

dymk:
Removes some cmake include magic so fbcode and cmake can build rsocket. Needed for Travis to run.


lehecka:
bla

PR #853, opened 2018-01-12
Prototype of using shuriken/rs as streaming API

dymk:
Prototypes out what using RSocket would look like if we used @per-gron's rs as our streaming API
Implements `requestStream` on `RSocketRequester`.


per-gron:
@dymk Nice that you are actually trying out rs! Lmk if you have any questions

phoad:
Try to make it an optional dependency
Don't copy paste the rs code but make it being downloaded to temporary folder when we build the project

phoad:
No progress on this diff.

PR #856, opened 2018-01-19
Implement fragmentation for requests

dymk:
WIP implementation of fragmented requests for RSocket. Allows payloads > 16MB to be sent. Should be efficient and not copy large payloads, if IOBuf refcounting is working correctly under the hood. Should not incur much extra overhead if a payload is not fragmented (aside from a cheap check to see if any fragments for the current stream already exist; should be no more expensive than `if(pointer == null)`).

Test implemented:
 - Request/Response

Tests not implemented yet:
 - Request Stream
 - Request Channel

Existing testsuite should pass.

### DIFF
--- a/rsocket/statemachine/RSocketStateMachine.cpp
+++ b/rsocket/statemachine/RSocketStateMachine.cpp
@@ -392,9 +392,9 @@ bool RSocketStateMachine::endStreamInternal(
   }
 
   // Remove from the map before notifying the stateMachine.
-  auto stateMachine = std::move(it->second);
+  auto streamElem = std::move(it->second);
   streamState_.streams_.erase(it);
-  stateMachine->endStream(signal);
+  streamElem.stateMachine->endStream(signal);
   return true;
 }
 
@@ -611,7 +611,8 @@ void RSocketStateMachine::handleStreamFrame(
   // we are purposely making a copy of the reference here to avoid problems with
   // lifetime of the stateMachine when a terminating signal is delivered which
   // will cause the stateMachine to be destroyed while in one of its methods
-  auto stateMachine = it->second;
+  auto& stateElem = it->second;
+  auto stateMachine = stateElem.stateMachine;
 
   switch (frameType) {
     case FrameType::REQUEST_N: {
@@ -634,10 +635,34 @@ void RSocketStateMachine::handleStreamFrame(
         return;
       }
       VLOG(3) << mode_ << " In: " << framePayload;
-      stateMachine->handlePayload(
-          std::move(framePayload.payload_),
-          framePayload.header_.flagsComplete(),
-          framePayload.header_.flagsNext());
+
+      if (!!(framePayload.header_.flags & FrameFlags::FOLLOWS) &&
+          !stateElem.fragmentAccumulator.anyFragments()) {
+        // first fragment seen for the frame; copy headers
+        stateElem.fragmentAccumulator.header = framePayload.header_;
+      }
+
+      bool const hasFollowsFlag =
+          !!(framePayload.header_.flags & FrameFlags::FOLLOWS);
+
+      if (hasFollowsFlag || stateElem.fragmentAccumulator.anyFragments()) {
+        stateElem.fragmentAccumulator.addPayload(
+            std::move(framePayload.payload_));
+
+        // final fragment in the payload, consume the payload
+        if (!hasFollowsFlag) {
+          stateMachine->handlePayload(
+              stateElem.fragmentAccumulator.consumePayload(),
+              stateElem.fragmentAccumulator.header.flagsComplete(),
+              stateElem.fragmentAccumulator.header.flagsNext());
+        }
+      } else {
+        // not a fragmented payload
+        stateMachine->handlePayload(
+            std::move(framePayload.payload_),
+            framePayload.header_.flagsComplete(),
+            framePayload.header_.flagsNext());
+      }
       break;
     }
     case FrameType::ERROR: {
@@ -681,15 +706,18 @@ void RSocketStateMachine::handleUnknownStream(
   // TODO: comparing string versions is odd because from version
   // 10.0 the lexicographic comparison doesn't work
   // we should change the version to struct
-  if (frameSerializer_->protocolVersion() > ProtocolVersion{0, 0} &&
-      !streamsFactory_.registerNewPeerStreamId(streamId)) {
-    return;
+  if (frameType != FrameType::PAYLOAD) {
+    // don't check registerNewPeerStreamId if it's a payload - it may be an
+    // additional fragment
+    if (frameSerializer_->protocolVersion() > ProtocolVersion{0, 0} &&
+        !streamsFactory_.registerNewPeerStreamId(streamId)) {
+      return;
+    }
   }
 
-  if (!isNewStreamFrame(frameType)) {
+  if (!isNewStreamFrame(frameType) && (frameType != FrameType::PAYLOAD)) {
     auto msg = folly::sformat(
         "Unexpected frame {} for stream {}", toString(frameType), streamId);
-    VLOG(1) << msg;
     closeWithError(Frame_ERROR::connectionError(std::move(msg)));
     return;
   }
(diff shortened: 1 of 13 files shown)


rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -681,15 +703,18 @@ void RSocketStateMachine::handleUnknownStream(
...
+    // don't check registerNewPeerStreamId if it's a payload - it may be an
+    // additional fragment
+    if (frameSerializer_->protocolVersion() > ProtocolVersion{0, 0} &&
+        !streamsFactory_.registerNewPeerStreamId(streamId)) {
+      return;
+    }
   }
 
-  if (!isNewStreamFrame(frameType)) {
+  if (!isNewStreamFrame(frameType) && !(frameType == FrameType::PAYLOAD)) {
```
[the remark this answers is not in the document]
dymk: Processing PAYLOAD frames for unknown streams is allowed because it's possible that the payload is a fragment for a stream that hasn't been initialized yet. If it's not found in the stream cache, the connection is closed with an error.

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -870,6 +993,72 @@ bool RSocketStateMachine::isClosed() const {
...
+
+    if(isFirstFrame) {
+      CHECK(metaLeft || dataLeft); // MUST be true, else we wouldn't have taken the writeFragmented path
+      isFirstFrame = false;
+      writeInitialFrame(std::move(sendme));
+    }
+    else {
+      // is this the last frame in the fragment chain?
+      bool follows = (metaLeft + dataLeft) > 0;
+      rs.outputFrameOrEnqueue(Frame_PAYLOAD(
```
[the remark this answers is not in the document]
dymk: I don't think so, see above comment regarding the protocol (PAYLOAD frames are what make up the remaining fragments of all stream elements)

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -634,10 +634,32 @@ void RSocketStateMachine::handleStreamFrame(
         return;
       }
       VLOG(3) << mode_ << " In: " << framePayload;
-      stateMachine->handlePayload(
-          std::move(framePayload.payload_),
-          framePayload.header_.flagsComplete(),
-          framePayload.header_.flagsNext());
+
```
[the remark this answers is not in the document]
dymk: The `StreamStateMachine` has no concept of a fragmented payload though; the only thing that knows about fragments and how to assemble them is the `RSocketStateMachine`. Perhaps the value type of the `streams_` map in `StreamState` should be a pair of `std::pair<StreamFragmentAccumulator, std::shared_ptr<StreamStateMachineBase>>` rather than just a `StreamStateMachineBase`.

rsocket/statemachine/RSocketStateMachine.cpp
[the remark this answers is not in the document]
lehecka: you are correct. I misread the spec. "When REQUEST_RESPONSE, REQUEST_FNF, REQUEST_STREAM, or REQUEST_CHANNEL frames need to be fragmented, the first frame is the REQUEST_* frame with the (F)ollows flag set, followed by a sequence of PAYLOAD frames."

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -681,15 +703,18 @@ void RSocketStateMachine::handleUnknownStream(
   // TODO: comparing string versions is odd because from version
   // 10.0 the lexicographic comparison doesn't work
   // we should change the version to struct
-  if (frameSerializer_->protocolVersion() > ProtocolVersion{0, 0} &&
-      !streamsFactory_.registerNewPeerStreamId(streamId)) {
-    return;
+  if(frameType != FrameType::PAYLOAD) {
```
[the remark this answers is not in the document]
dymk: It will register the stream ID for the incoming REQUEST, and then drop the rest of the fragments coming in via PAYLOAD frames later. We only want to register a new stream ID for non-Payload types, in that case.

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -870,6 +993,72 @@ bool RSocketStateMachine::isClosed() const {
...
+// of the data and metadata combined, plus the size of the frame header.
+// This assumes that the frame header will never be more than 512 bytes in
+// size. A CHECK in FrameTransportImpl enforces this. The idea is that
+// 16M is so much larger than the ~500 bytes possibly wasted that it won't
+// be noticable (0.003% wasted at most)
+constexpr size_t GENEROUS_MAX_FRAME_SIZE = 0xFFFFFF - 512;
+
+namespace detail {
+template <typename WriteInitialFrame>
+void writeFragmented(
```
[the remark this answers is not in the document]
dymk: I can try to abstract this out yeah

rsocket/statemachine/RSocketStateMachine.cpp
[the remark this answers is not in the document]
lehecka: the PAYLOAD frame has streamId so we always know to which stream it belongs to. Maybe if we would add add fragmentAccumulator to StreamStateMachineBase we would be able to move this logic out of RSStateMachine? (i am just thinking if this class isn't becoming "class everything" with many responsibilities which are impossible to test individually)

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -681,15 +706,18 @@ void RSocketStateMachine::handleUnknownStream(
   // TODO: comparing string versions is odd because from version
   // 10.0 the lexicographic comparison doesn't work
   // we should change the version to struct
-  if (frameSerializer_->protocolVersion() > ProtocolVersion{0, 0} &&
-      !streamsFactory_.registerNewPeerStreamId(streamId)) {
-    return;
+  if (frameType != FrameType::PAYLOAD) {
```
lehecka: Am I reading the code right that it is allowing PAYLOAD frame to come first on a new stream?

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -681,15 +706,18 @@ void RSocketStateMachine::handleUnknownStream(
...
+    // don't check registerNewPeerStreamId if it's a payload - it may be an
+    // additional fragment
+    if (frameSerializer_->protocolVersion() > ProtocolVersion{0, 0} &&
+        !streamsFactory_.registerNewPeerStreamId(streamId)) {
+      return;
+    }
   }
 
-  if (!isNewStreamFrame(frameType)) {
+  if (!isNewStreamFrame(frameType) && !(frameType == FrameType::PAYLOAD)) {
```
lehecka: frameType != FrameType::PAYLOAD

rsocket/statemachine/StreamFragmentAccumulator.h
```diff
@@ -0,0 +1,69 @@
...
+
+ public:
+  void addPayload(Payload p) {
+    if (p.metadata) {
+      if (!fragments.metadata) {
+        fragments.metadata = std::move(p.metadata);
+      } else {
+        fragments.metadata->prev()->appendChain(std::move(p.metadata));
+      }
+    }
```
lehecka: would you dislike using folly::IOBufQueue for metadata and data to avoid this if statement and messing up with iobuf internal structure?

rsocket/statemachine/RSocketStateMachine.cpp
```diff
@@ -882,13 +1076,46 @@ void RSocketStateMachine::writeNewStream(
         streamId, RequestOriginator::LOCAL, streamToken, streamType);
   }
 
+  // for simplicity, require that sent buffers don't consist of chains
+  if ((bufChainedLength(payload.metadata) + bufChainedLength(payload.data)) >
+      GENEROUS_MAX_FRAME_SIZE) {
+    detail::writeFragmented(
+        [&](Payload p) {
```
lehecka: this is only for some individuals with strong stomach. ;) Why would you not pass streamType into the writeFragmented method and inline the lambda inside the function too?

Can we put all of this logic to a separate class so that RSStateMachine is no blowing up in size and responsibilities?

rsocket/test/RequestResponseTest.cpp
```diff
@@ -198,3 +198,119 @@ TEST(RequestResponseTest, FailureOnRequest) {
...
+        lastVal.first, niceLongData, "data (recieved on client)");
+    debugStringMismatch(
+        lastVal.second, niceLongMetadata, "metadata (recieved on client)");
+  };
+
+  // All in one big chunk
+  checkForSizePattern({}, {});
+
+  // Small chunk, big chunk, small chunk
+  checkForSizePattern(
```
lehecka: I like this unit test!

rsocket/statemachine/RSocketStateMachine.h
```diff
@@ -244,10 +269,20 @@ class RSocketStateMachine final
       StreamType streamType,
       uint32_t initialRequestN,
       Payload payload) override;
+
+  void writeFragmented(
```
lehecka: is this declaration used?

rsocket/statemachine/RSocketStateMachine.cpp
dymk: No, the payload would be some other frame. The first frame received will be a `REQUEST_` type frame, however subsequent frames will be `PAYLOAD` frames in case of a fragmented request. These subsequent frames would fail the `!registerNewPeerStreamId` check. So the check is skipped if the frame is a `PAYLOAD`.

rsocket/statemachine/StreamFragmentAccumulator.h
dymk: It'd mean making `StreamFragmentAccumulator` quite a bit heavier. Also we need to track if the final `fragments.metadata` should be a nullptr or not; an IOBufQueue would unconditionally return an empty non-null `IOBuf` pointer when moving its accumulator. More complex than it needs to be.

PR #859, opened 2018-02-06
[yarpl] Implementation of PublishProcessor

lehecka:
This is gonna be useful for replacing fromPublisher when we need to defer connecting consumer to flowable.
http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/processors/PublishProcessor.html

new unit tests added.


yarpl/include/yarpl/flowable/PublishProcessor.h
```diff
@@ -0,0 +1,234 @@
...
+   private:
+    std::atomic<int64_t> credits_{0};
+    std::shared_ptr<Subscriber<T>> subscriber_;
+    PublishProcessor* processor_;
+  };
+
+  folly::Synchronized<std::shared_ptr<const PublishersVector>, std::mutex>
+      publishers_;
+
+  static std::shared_ptr<const PublishersVector> kCompletedPublishers;
```
dymk: Do these only exist to "tag" `publishers_`? If so, I would just keep an extra bit of state on `publishers_` within the syncrhonized and test if we're in a `completed` or `errored` state. The magic with the out of line, statically initialized vectors is confusing. Any modules that try to perform static initialization of a `PublishProcessor` are also now subject to the [static initialization order fiasco](https://isocpp.org/wiki/faq/ctors#static-init-order).

yarpl/include/yarpl/flowable/Subscriber.h
```diff
@@ -135,6 +135,10 @@ class BaseSubscriber : public Subscriber<T>, public yarpl::enable_get_ref {
 
   virtual void onTerminateImpl() {}
 
+  std::shared_ptr<Subscription> getSubscription() {
```
dymk: `BaseSubscriber` exposes all methods of `Subscription` on it directly, does there need to be a getter to expose it further?

yarpl/src/yarpl/utils/credits.cpp
```diff
@@ -89,6 +89,28 @@ int64_t consume(std::atomic<std::int64_t>* current, int64_t n) {
   }
 }
 
+bool tryConsume(std::atomic<std::int64_t>* current, int64_t n) {
```
dymk: we should 100% make this class single-threaded, and have all threadedness be explicit in the codebase. It's unclear what objects and methods are alright to call from multiple threads, and what needs explicit synchronization.

yarpl/include/yarpl/flowable/PublishProcessor.h
```diff
@@ -0,0 +1,234 @@
...
+
+    // cancel may race with terminate(), but the
+    // PublishProcessor::removePublisher will take care of that the race with
+    // on{Next, Error, Complete} methods is allowed by the spec
+    void cancel() override {
+      credits_ = credits::kCanceled;
+      processor_->removePublisher(this);
+    }
+
+    // we don't care about races with this method
```
dymk: Hm, why not?

yarpl/include/yarpl/flowable/PublishProcessor.h
```diff
@@ -0,0 +1,234 @@
...
+      if (*locked == oldPublishers) {
+        *locked = newPublishers;
+        return newPublishers;
+      }
+      // else the vector changed so we will have to do it again
+    }
+  }
+
+  void removePublisher(PublisherSubscription* subscriber) {
+    while (true) {
```
dymk: this can be a lot simpler and likely just as performant: Exclusive lock on `publishers_` and [`std::remove_if`](http://en.cppreference.com/w/cpp/algorithm/remove)

yarpl/include/yarpl/flowable/PublishProcessor.h
dymk: IMO we shouldn't be allowing requesting and signaling methods to race at all. We end up with implementations that need to pessimistically lock (which introduces both runtime overhead, and implementation complexity considerably). I think it's possible to have an architecture which is explicit in scheduling work on different executors and only performs synchronization at those boundaries (eg, `subscribe/observeOn`).

yarpl/include/yarpl/flowable/PublishProcessor.h
dymk: SIFO can occur if another module attempts to use `PublisherProcessor` directly though. STL types aren't special in the order they're statically initialized.

I think the code will be cleaner if the state is explicit, instead of relying on pointer equality. It'll be easier to introspect the state of the class as well when debugging. There's already a lock (the `folly::Syncronized`) that the state can be put behind.

PR #860, opened 2018-02-12
Flowable<T>::makeFlowable

dymk:
Easier and less bugprone way of making a flowable out of an invokable which directly takes a Subscriber. It's much harder to forget how to do flow control when the object being called *is* the subscription, which captures the provided subscriber.


phoad:
No progress on this diff.

PR #866, opened 2018-05-31
Update README.md

phoad:
Installing HEAD is not needed for Folly.
Better to install the latest release `brew install folly`


phoad:
No need.

PR #893, opened 2019-08-01
Fix stream id overflow problem

linux-china:
reuse stream id and avoid id conflict.


rsocket/statemachine/RSocketStateMachine.h
```diff
@@ -324,7 +324,7 @@ class RSocketStateMachine final
   /// Map of all individual stream state machines.
   std::unordered_map<StreamId, std::shared_ptr<StreamStateMachineBase>>
       streams_;
-  StreamId nextStreamId_;
+  std::atomic<StreamId> nextStreamId_;
```
OlegDokuka, line 327: Guess that should not be atomic since all the events are serialized over event bus

